PR #30, opened 2019-11-18
Added expressions and tests

SND96:
Added more comparison expressions, arithmetic expressions and logical expressions. Added tests for each of the new expressions.


src/expression/arithmetic_expression.py
```diff
@@ -0,0 +1,31 @@
...
+        children = []
+        if left is not None:
+            children.append(left)
+        if right is not None:
+            children.append(right)
+        super().__init__(exp_type, rtype=ExpressionReturnType.BOOLEAN,
+                         children=children)
+
+    def evaluate(self, *args):
+        vl = self._children[0].evaluate(*args)
```
gaurav274: Avoid using _children. use get_child

src/expression/logical_expression.py
```diff
@@ -0,0 +1,31 @@
...
+        if left is not None:
+            children.append(left)
+        if right is not None:
+            children.append(right)
+        super().__init__(exp_type, rtype=ExpressionReturnType.BOOLEAN,
+                         children=children)
+
+    def evaluate(self, *args):
+        # print(*args)
+        if len(self._children) == 2 :
```
gaurav274: Use get child_count. use get_child

src/expression/logical_expression.py
```diff
@@ -0,0 +1,31 @@
+from .abstract_expression import AbstractExpression, ExpressionType, \
```
gaurav274: rewrite this import. use src/....

SND96:
@gaurav274 Made the changes to _children and replaced with get_child functions. Also modified the import statements to be global.

PR #31, opened 2019-11-21
Integration of Depth estimation model with Eva

pratrajv:
@gaurav274


src/loaders/abstract_loader.py
```diff
@@ -44,12 +44,14 @@ class AbstractVideoLoader(metaclass=ABCMeta):
     """
 
     def __init__(self, video_metadata: VideoMetaInfo, batch_size=1,
-                 skip_frames=0, offset=None, limit=None):
+                 skip_frames=0, offset=None, limit=None, threshold=0, distance_metric='absolute_difference'):
```
saiprashanth173: This is an abstract class and contains those attributes which are common to all loaders. 

threshold and distance_metric look very specific to your use-case. You can create your own loader which inherits the AbstractVideoLoader and add there attributes to that class.

src/loaders/video_loader.py
```diff
@@ -29,6 +32,14 @@ def load(self):
                 _, frame = video.read()
                 continue
 
+            if self.threshold > 0:
```
saiprashanth173: I am assuming this logic handles the frames which will be passed to your UDF. 

If this logic is an optimization specific to your UDF, then I feel this code doesn't belong here.

src/udfs/abstract_udfs.py
```diff
@@ -46,3 +47,14 @@ def classify(self, batch: FrameBatch) -> List[Prediction]:
             List[Prediction]: The predictions made by the classifier
         """
         pass
+
+    @abstractmethod
```
saiprashanth173: We will consider renaming the **classify** method and give it a more generic name that suites other UDFs. For now, just override the **classify** method instead of creating a new abstract method named process_frames.
Refer this commit: The **classify** method now returns a list of **BasePrediction**. (Which should be DepthEstimationResult in your case)
https://github.com/georgia-tech-db/Eva/commit/d7f64d52ed7bf20eb251601b70388d81b57aa058

src/udfs/depth_estimator.py
```diff
@@ -0,0 +1,135 @@
...
+from src.udfs.abstract_udfs import AbstractClassifierUDF
+from src.models import FrameBatch, Prediction, FrameInfo, Point, BoundingBox, ColorSpace
+from torch.autograd import Variable
+from torchvision import transforms
+from src.depth_estimation_result import DepthEstimationResult
+from src.utils.frame_filter_util import FrameFilter
+
+sys.path.append('../')
+sys.path.append('.')
+folder = os.path.dirname(os.path.abspath(__file__))
```
gaurav274: Can you please remove these hard-coded paths

src/udfs/depth_estimator.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.model = net(num_classes=NUM_CLASSES, num_tasks=NUM_TASKS)
+        if HAS_CUDA:
+            _ = self.model.cuda()
+            _ = self.model.eval()
+        ckpt = torch.load(os.path.join(folder, 'ExpKITTI_joint.ckpt'))
+        self.model.load_state_dict(ckpt['state_dict'])
+
```
gaurav274: remove the hard-coded folder and required data paths. Assume these will be passed as parameters

src/udfs/depth_estimator.py
```diff
@@ -0,0 +1,135 @@
...
+            depths.append(depth)
+            segms.append(segm)
+
+            if i % 5 == 0:
+                depth_mask = depth > 15
+                depth_mask[100: ] = True
+
+        return segms, depths
+
+    def process_frames(self, batch: FrameBatch) -> List[DepthEstimationResult]:
```
gaurav274: use the abstract class method

src/depth_estimation_result.py
```diff
@@ -0,0 +1,62 @@
+from src.models import FrameBatch, Frame
+from typing import List, Tuple
+
+import numpy as np
+
+class DepthEstimationResult:
```
gaurav274: make it a child class of BasePrediction - wait for Prashanth's PR

PR #32, opened 2019-11-22
Rule Based Query Optimizer

kcaras:
Here is a Rule Based Query Optimizer that implements predicate pushdown, projection pushdown, join elimination, and simplify predicates such as 1==0. 

There are a total of 10 tests.


gaurav274:
Can you please make changes we discussed?
1. Move tests to tests dir
2. Remove all the binary files (*.avi files)
3. Remove the spatial and temporal filtering files. We can take them as a separate PR.

src/query_planner/abstract_select_plan.py
```diff
@@ -0,0 +1,61 @@
+"""Abstract class for all the scan planners
```
gaurav274: Confused. we already have an abstract scan then why abstract select?

src/query_planner/abstract_table_plan.py
```diff
@@ -0,0 +1,36 @@
+"""Abstract class for all the scan planners
```
gaurav274: What is a table plan?

src/query_planner/video_table_plan.py
```diff
@@ -0,0 +1,20 @@
+from src.loaders.abstract_loader import AbstractVideoLoader
```
gaurav274: This shouldn't be a plan node right?

src/query_optimizer/statement2plantree.py
```diff
@@ -0,0 +1,48 @@
...
+from models.catalog.video_info import VideoMetaInfo
+from models.catalog.properties import VideoFormat
+
+
+# class that will handle converting a statement to a plan tree
+class Statement2Plantree:
+    def __init__(self):
+        pass
+
+    # Function that converts the statement list to a plan tree
```
gaurav274: Comments are not as per our standards. And why is there a class for just one static function? We need to keep some sort of framework where other modules just call a single function with evaStatementList and then it is the job of this module to figure out which type of statement it is and call the corresponding converter.

src/query_optimizer/rule_query_optimizer_playground.py
```diff
@@ -0,0 +1,196 @@
+# this is an implementation of a Rule Based Query Optimizer
```
gaurav274: Remove this?

src/query_optimizer/rule_query_optimizer.py
```diff
@@ -0,0 +1,225 @@
+from expression.constant_value_expression import ConstantValueExpression
```
gaurav274: Can you please modify the commenting style? Use docstring. I will look at the functionality once you are done with this.

src/query_optimizer/rule_query_optimizer_playground.py
bkim360: Done.

src/query_planner/abstract_select_plan.py
bkim360: We were not sure about the naming scheme. Should we just port over the methods from abstract select to abstract scan and delete abstract select after?

src/query_planner/abstract_table_plan.py
bkim360: We thought we would need a table plan for the videos (the table representing the video).

src/query_optimizer/rule_query_optimizer.py
bkim360: Changed them for rule_query_optimizer.py and the abstract plans.

src/query_planner/abstract_select_plan.py
gaurav274: Yeah go ahead and do that.

src/query_planner/abstract_table_plan.py
gaurav274: Have a look at the table_ref in src/query_parser/table_ref.py.

bkim360: Does this mean we should scrap the video_table_plan and abstract_table_plan and replace them with the table_ref?

src/query_planner/abstract_select_plan.py
bkim360: I have changed the references of select plan to seq scan plan. If this is correct, I will delete the abstract select plan and logical select plan.

src/query_planner/abstract_table_plan.py
bkim360: I have changed the references of video_table_plan and abstract_table_plan into table_ref.

src/query_planner/video_table_plan.py
bkim360: Mentioned above.

PR #33, opened 2019-11-23
Eva query processor

saiprashanth173:
This PR will add/update.
1. query_planner (Plan Nodes for plan tree)
2. query_executor (Executor Nodes for query execution). PlanExecutor for executing the plan. 
3. FunctionExpressions for evaluating the UDFs


src/expression/comparison_expression.py
```diff
@@ -15,11 +15,18 @@ def __init__(self, exp_type: ExpressionType, left: AbstractExpression,
                          children=children)
 
     def evaluate(self, *args):
-        vl = self._children[0].evaluate(*args)
-        vr = self._children[1].evaluate(*args)
+        vls = self._children[0].evaluate(*args)
```
gaurav274: use get_child(0), get_child(1)

src/query_planner/abstract_plan.py
```diff
@@ -1,19 +1,14 @@
...
-class PlanNodeType(IntEnum):
-    SEQSCAN = 1
-    # add other types
+from src.query_planner.types import PlanNodeType
 
 
 class AbstractPlan(ABC):
 
-    def __init__(self):
+    def __init__(self, node_type):
```
gaurav274: what will be the node_type for an abstract scan node? I mean if we don't want to assign any type to abstract nodes then it is waste to keep it as a member variable

src/query_planner/abstract_scan_plan.py
```diff
@@ -2,35 +2,23 @@
...
-        self._column_ids = column_ids
-        self._video = video
-
-    @property
-    def video(self) -> Storage:
-        return self._video
 
     @property
-    def predicate(self) -> Expression:
+    def predicate(self) -> 'Expression':
```
gaurav274: AbstractExpression

src/query_planner/abstract_plan.py
saiprashanth173: Having a node_type for every plan makes it easy to perform a condition check (switch cases). If we see the conversion of a plan tree to the execution tree we will have to use the type to determine which executor needs to be created. 

Having this in the superclass (AbstractPlan) will make sure any Plan Node will have this property. 

If you check the implementation of the _build_execution_tree method of PlanExecutor, we make use of this node_type attribute. If we don't have this we need to use **isinstance**, which I feel is not a great thing to do.

src/expression/comparison_expression.py
saiprashanth173: Fixed

src/query_planner/abstract_scan_plan.py
saiprashanth173: Fixed

src/query_planner/abstract_plan.py
gaurav274: Agreed we won't be using isinstance, rather I meant making node_type as a method of each plan node. We don't need a variable in the base class. And we won't need to pass the node_type each time to create a node.

PR #34, opened 2019-11-23
Integrate frame differencing into EVA video loader

jhanavi:
**Description:** 
Integrate frame differencing into EVA video loader to be able to skip similar frames. Current EVA implementation has support for sampling every nth frame from a given video stream. Support has now been added to compare frames and skip frames based on following metrics:
1. Mean Square Error
2. Absolute Difference

The  threshold value and distance metric are taken as parameters by the Video Loader class.

The above distance metrics can be used to compare all the pixels in the frame or based on the scenario, only the foreground pixels. For the latter, the flag **compare_foreground** can be set to True so that the comparison would happen only on the detected foreground pixels. A function to identify the foreground mask has been added to the utility functions so that it can be leveraged by other components as well. 

**Testing:**
Appropriate tests have been added to test all the distance metrics as well as the foreground mask comparison. Following is the list of tests with each function name describing the tested scenario and expected result:
1. test_should_skip_identical_frames_absdiff
2. test_should_skip_identical_frames_msediff
3. test_should_skip_identical_frames_only_foreground

**Usage:**
Usage corresponding to above tests are:
1. test_should_skip_identical_frames_absdiff:
 > SimpleVideoLoader(video_info, threshold=0.5, distance_metric='absolute_difference')

2. test_should_skip_identical_frames_msediff
 > SimpleVideoLoader(video_info, threshold=0.5, distance_metric='mse_difference')

3. test_should_skip_identical_frames_only_foreground
 > SimpleVideoLoader(video_info, threshold=0.5, distance_metric='absolute_difference', compare_foreground=True)

The distance metric should correspond to the function name as defined in framediff_utils. Thus, as of now the following distance metrics can be passed:
1. absolute_difference
2. mse_difference


jhanavi:
@gaurav274 , kindly review our changes and let us know if there's anything we need to change.

src/loaders/abstract_loader.py
```diff
@@ -44,12 +44,17 @@ class AbstractVideoLoader(metaclass=ABCMeta):
     """
 
     def __init__(self, video_metadata: VideoMetaInfo, batch_size=1,
-                 skip_frames=0, offset=None, limit=None):
+                 skip_frames=0, offset=None, limit=None, threshold=0,
```
gaurav274: We don't want to change the abstract Loader. It will be better if you guys can add a new subclass with your updated parameters.

src/loaders/video_loader.py
```diff
@@ -33,10 +35,29 @@ def load(self):
                 _, frame = video.read()
                 continue
 
+            # Skip similar frames if threshold is a positive value
```
gaurav274: Again don't change the SimpleVideoLoader, it is meant to remain the way it is. right now I'm now even sure if we want your logic to be tightly coupled with a video loader. I mean we may even want to use frame differencing logic somewhere else during the pipeline. Shouldn't it be a PP(probabilistic predicate)? Just an opinion, do let me know if you guys have some other views.

src/utils/framediff_utils.py
```diff
@@ -0,0 +1,93 @@
+import cv2
+import numpy as np
+
+"""
+
+Utility functions to calculate inter-frame difference
+
+"""
+
+class DistanceMetrics:
```
gaurav274: We normally don't implement classes in utils. I mean util functions are anyways meant to be static methods so why keep them in a class when we are not using any object state.

src/utils/framediff_utils.py
```diff
@@ -0,0 +1,93 @@
...
+	def mse_difference(self, curr_frame, prev_frame):
+		curr_frame_grayscale = self.convert_to_grayscale(curr_frame)
+		prev_frame_grayscale = self.convert_to_grayscale(prev_frame)
+		total_pixels = curr_frame_grayscale.shape[0]*curr_frame_grayscale.shape[1]
+		frame_diff = curr_frame_grayscale.astype('float') - prev_frame_grayscale
+		mse_diff = np.sum(frame_diff**2)/float(total_pixels)
+		return mse_diff
+
+# Function to blacken out background of frame
+def mask_background(frame):
```
gaurav274: Can we just plug out all the hardcoded variables from this function and pass them as parameters. **kwargs is normally the way to do this in python. I mean the users can pass whatever parameters they want to control and rest can be defaults.

PR #35, opened 2019-11-24
Antlr Visitor Pattern Based FrameQL Parser Integration

gaurav274:
Key updates:
We can parse select input strings and convert them to an Eva Select Statement. (multiple from tables, join not supported yet)
Expressions generated directly from the parser. Can parse all the expression types supported by Eva. 

Coverage is decreasing. The reason being we have an auto-generated parser visitor file. I don't feel comfortable to remove the unused functionalities as we will progressively building the parser.


src/query_parser/select_statement.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @property
+    def from_table(self):
+        return self._from_table
+
+    @from_table.setter
+    def from_table(self, table: TableRef):
+        self._from_table = table
+
+    def __str__(self):
```
saiprashanth173: Use python string formatting instead of concatenation. 
ex. "SELECT {} FROM {}".format(str(columns), str(...)).

src/query_parser/table_ref.py
```diff
@@ -0,0 +1,34 @@
...
+    @property
+    def schema_name(self):
+        return self._schema_name
+
+    @property
+    def database_name(self):
+        return self._database_name
+
+
+class TableRef():
```
saiprashanth173: Brackets are redundant. 
make it class TableRef:

test/query_parser/test_parser.py
```diff
@@ -0,0 +1,78 @@
...
+from src.query_parser.eva_statement import EvaStatement
+from src.query_parser.eva_statement import StatementType
+from src.query_parser.select_statement import SelectStatement
+from src.expression.abstract_expression import ExpressionType
+from src.query_parser.table_ref import TableRef
+
+
+class ParserTest(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
saiprashanth173: Can we also add some test cases for scenarios where the attributes and name of video are in lower case? 

In SQL, it's common to have Keywords(SELECT, FROM, JOINT etc.) as capitals and everything else in lowercase.

src/query_parser/table_ref.py
```diff
@@ -0,0 +1,34 @@
+class TableInfo:
+    """
```
saiprashanth173: I am not sure if the naming is appropriate for Eva. 

What are we referring to by a table_name here? 
Coming to the terminology, what are we referring to by schema_name and database_name. Is this applicable in the case of EVA.

src/query_parser/table_ref.py
```diff
@@ -0,0 +1,34 @@
+class TableInfo:
```
saiprashanth173: Do you feel it's a good idea to move all the data classes to one place (**models** module)?

src/query_parser/eva_parser.py
```diff
@@ -0,0 +1,21 @@
+import sys
+from antlr4 import InputStream, CommonTokenStream
+from third_party.evaQL.parser.frameQLParser import frameQLParser
+from third_party.evaQL.parser.frameQLLexer import frameQLLexer
+from src.query_parser.eva_ql_parser_visitor import EvaParserVisitor
+
+
+class EvaFrameQLParser():
```
saiprashanth173: Brackets redundant.

src/query_parser/select_statement.py
```diff
@@ -0,0 +1,63 @@
...
+        list of target attributes in the select query,
+        each attribute is represented as a Abstract Expression
+    _from_table : TableRef
+        table reference in the select query
+    _where_clause : AbstractExpression
+        predicate of the select query, represented as a expression tree.
+    **kwargs : to support other functionality, Orderby, Distinct, Groupby.
+    """
+
+    def __init__(self, target_list=None, from_table=None,
```
saiprashanth173: Let's try to use type annotations as much as possible for clarity. 
ex: target_list: List[AbstractExpression]

src/expression/tuple_value_expression.py
```diff
@@ -3,15 +3,15 @@
 
 
 class TupleValueExpression(AbstractExpression):
-    def __init__(self, col_idx: int):
+    def __init__(self, col_idx=None, col_name=None):
```
saiprashanth173: Why do we need both col_idx and col_name?

PR #40, opened 2019-12-05
Create new pp to return boolean list for which frames to skip

jhanavi:
**Description:** 
Created a new pp to return a boolean list with values set to true if frame should be skipped. Current EVA implementation has support for sampling every nth frame from a given video stream. Support has now been added to compare frames and skip frames based on following metrics:
1. Mean Square Error
2. Absolute Difference

The  threshold value and distance metric are taken as parameters by the Video Loader class.

The above distance metrics can be used to compare all the pixels in the frame or based on the scenario, only the foreground pixels. For the latter, the flag **compare_foreground** can be set to True so that the comparison would happen only on the detected foreground pixels. A function to identify the foreground mask has been added to the utility functions so that it can be leveraged by other components as well. 

**Testing:**
Appropriate tests have been added to test all the distance metrics as well as the foreground mask comparison. Following is the list of tests with each function name describing the tested scenario and expected result:
1. test_should_skip_identical_frames_absdiff
2. test_should_skip_identical_frames_msediff
3. test_should_skip_identical_frames_only_foreground

**Usage:**
Usage corresponding to above tests are:
1. test_should_skip_identical_frames_absdiff:
 > SimpleVideoLoader(video_info, threshold=0.5, distance_metric='absolute_difference')

2. test_should_skip_identical_frames_msediff
 > SimpleVideoLoader(video_info, threshold=0.5, distance_metric='mse_difference')

3. test_should_skip_identical_frames_only_foreground
 > SimpleVideoLoader(video_info, threshold=0.5, distance_metric='absolute_difference', compare_foreground=True)

The distance metric should correspond to the function name as defined in image_utils. Thus, as of now the following distance metrics can be passed:
1. absolute_difference
2. mse_difference


jhanavi:
@gaurav274 @saiprashanth173 Kindly review this PR. We've decoupled the changes from video loader and created a new pp for identifying identical frames.

src/filters/frame_skipping_pp.py
```diff
@@ -0,0 +1,52 @@
+from src.filters.abstract_pp import AbstractPP
+from src.utils import image_utils
+from typing import List
+
+
+class frameSkippingPP(AbstractPP):
+
+    def __init__(self, threshold=0.0, compare_foreground=False,
```
gaurav274: Missing comments for what each parameter does? Names are self-explanatory but still, it is a good practice.

src/filters/frame_skipping_pp.py
```diff
@@ -0,0 +1,52 @@
...
+                    skipFrames.append(True)
+                else:
+                    skipFrames.append(False)
+                    prev_frame = frame
+            else:
+                skipFrames.append(False)
+                prev_frame = frame
+        return skipFrames
+
+    # Function to calculate frame difference based on distance metric
```
gaurav274: use docstring to write comments

src/utils/image_utils.py
```diff
@@ -0,0 +1,70 @@
...
+    curr_frame_gray = convert_to_grayscale(curr_frame)
+    prev_frame_gray = convert_to_grayscale(prev_frame)
+    total_pixels = curr_frame_gray.shape[0] * curr_frame_gray.shape[1]
+    frame_diff = curr_frame_gray.astype('float') - prev_frame_gray
+    mse_diff = np.sum(frame_diff**2) / float(total_pixels)
+    return mse_diff
+
+
+# Function to blacken out background of frame
+def mask_background(frame, dim=21, threshold1=10, threshold2=200,
```
gaurav274: Let us keep parameter annotation a standard for our system.

src/filters/frame_skipping_pp.py
```diff
@@ -0,0 +1,52 @@
+from src.filters.abstract_pp import AbstractPP
+from src.utils import image_utils
+from typing import List
+
+
+class frameSkippingPP(AbstractPP):
```
saiprashanth173: Class names start with capital letter. Rename to FrameSkippingPP

src/filters/frame_skipping_pp.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self, threshold=0.0, compare_foreground=False, 
+                 distance_metric='absolute_difference'):
```
saiprashanth173: string parameter, eg. 'absolute_difference' might cause confusion. 

This is not a major issue, as many python packages (sklearn etc.) take string params as input. For Eva, it would be good if we follow strong typing. We can make distance metric an *Enum* containing different metrics instead of strings.

PS: You will still be able to use *getattr* even with enums

src/filters/frame_skipping_pp.py
```diff
@@ -0,0 +1,52 @@
...
+        self.threshold = threshold
+        self.compare_foreground = compare_foreground
+        self.distance_metric = distance_metric
+
+    def predict(self, batch) -> List[bool]:
```
saiprashanth173: From your logic, parameter 'batch' is of type List[FrameBatch]. 

All our APIs PPs and UDFs take a FrameBatch as input and not List[FrameBatch]. Please change your logic accordingly. 

Use type hints for parameters (code style suggestion)

src/filters/frame_skipping_pp.py
saiprashanth173: Also, please add type hints both for parameters and return type.

src/filters/frame_skipping_pp.py
```diff
@@ -0,0 +1,52 @@
...
+        prev_frame = None
+        skipFrames = []
```
saiprashanth173: Code style: variable names are snake-case in python.

test/loaders/test_video_loader.py
```diff
@@ -38,11 +39,60 @@ def create_sample_video(self):
                              dtype=np.uint8)
             out.write(frame)
 
-    def setUp(self):
-        self.create_sample_video()
+    def create_sample_video_with_similar_frames(self):
```
saiprashanth173: Ahh... A few issues here. 

This is a PP filter, don't mix its tests with the loader. 

1. Create a new test file in `test/filters` and write your tests there. 
2. You are expected to write unit tests for your logic, so you wouldn't be needing simple video loader in your tests.
3. [For your new tests] Instead of creating a dummy video file with similar frames create a FrameBatch object and this will be input to your `predict` method. 

Thanks for updating setUp to setUpClass!

jhanavi: Addressed all three comments. Tests for the added pp moved to test_frame_skipping_pp.py in tests/filters. Now passing batch of frames instead of creating a new video. @saiprashanth173

src/filters/frame_skipping_pp.py
jhanavi: Modified input to pp to be a batch of frames.

src/utils/image_utils.py
jhanavi: Added

src/filters/frame_skipping_pp.py
jhanavi: Added comments for all functions

src/filters/frame_skipping_pp.py
jhanavi: Added type hints and converted all comments to docstring

src/filters/frame_skipping_pp.py
jhanavi: Renamed

src/filters/frame_skipping_pp.py
jhanavi: Renamed

src/filters/frame_skipping_pp.py
jhanavi: Made distance metric an ENUM.

src/filters/frame_skipping_pp.py
```diff
@@ -0,0 +1,101 @@
+from src.filters.abstract_pp import AbstractPP
+from src.utils import image_utils
+from typing import List
+from enum import Enum, unique
+
+
+@unique
+class DistanceMetric(Enum):
```
saiprashanth173: I guess a better place for DistanceMetric Enum is image_utils.

src/filters/frame_skipping_pp.py
```diff
@@ -0,0 +1,101 @@
...
+        objects and then calculates the frame difference
+
+        :param curr_frame: current frame to be processed from a batch
+        :param prev_frame: previous frame that was not skipped
+        :param distance_metric: DistanceMetric ENUM attributes
+        :return difference value of the two frames
+        """
+        curr_foreground = mask_background(curr_frame)
+        prev_foreground = mask_background(prev_frame)
+        return frame_difference(curr_foreground, prev_foreground,
```
saiprashanth173: Isn't this supposed to be self.frame_difference?

This method was never called in any tests. Where is mask_backgroud located? Couldn't find this in imports.

src/filters/frame_skipping_pp.py
```diff
@@ -0,0 +1,101 @@
...
+        skip_frames = []
+        frames = batch.frames_as_numpy_array()
+        for frame in frames:
+            if prev_frame is not None:
+                """
+                If compare_foreground set to true, calculate distance
+                metric on only the foreground pixels
+                else on the entire image.
+                """
+                if self.compare_foreground is True:
```
saiprashanth173: This logic is not tested.

src/utils/image_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    curr_frame_gray = convert_to_grayscale(curr_frame)
+    prev_frame_gray = convert_to_grayscale(prev_frame)
+    total_pixels = curr_frame_gray.shape[0] * curr_frame_gray.shape[1]
+    frame_diff = curr_frame_gray.astype('float') - prev_frame_gray
+    mse_diff = np.sum(frame_diff**2) / float(total_pixels)
+    return mse_diff
+
+
+def mask_background(frame, dim=21, threshold1=10, threshold2=200,
```
saiprashanth173: This method doesn't have any tests. 

If you feel testing the actual output is difficult, conisder using mocking and stubbing.

src/utils/image_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    background_mask = cv2.GaussianBlur(background_mask, (dim, dim), 0)
+    stacked_mask = np.dstack(3 * [background_mask])
+
+    frame = frame / 255.0
+    stacked_mask = stacked_mask / 255.0 
+    
+    masked_image = (stacked_mask * frame) + ((1 - stacked_mask) * color) 
+    masked_image = (masked_image * 255).astype('uint8')
+
+    return masked_image
```
saiprashanth173: please insert a new line

src/filters/frame_skipping_pp.py
Vaishnavik22: Added it in image_utils as suggested

src/filters/frame_skipping_pp.py
Vaishnavik22: Fixed this issue

src/filters/frame_skipping_pp.py
Vaishnavik22: testing this logic using mse and absolute difference

src/utils/image_utils.py
Vaishnavik22: Done

gaurav274:
Obsolete code

PR #41, opened 2019-12-05
Adding compression and recovery functionality to storage module 

alilakda:


alilakda:
Updates to Frame_Info and Video_Info to handle compressed videos.

PR #47, opened 2019-12-29
Fix setup commands

jarulraj:


script/formatting/formatter.py
```diff
@@ -0,0 +1,233 @@
...
+                   args.add_header,
+                   args.strip_header,
+                   args.format_code)
+    # BY DEFAULT, WE SCAN THE DEFAULT DIRS AND FIX THEM
+    else:
+        LOG.info("Default scan")
+        for dir in DEFAULT_DIRS:
+            LOG.info("Scanning : " + dir + "\n\n")
+
+#             LOG.info("Stripping headers : " + dir)
```
saiprashanth173: Any reason for having this commented code? 
Can we remove this?

src/catalog/catalog_manager.py
```diff
@@ -0,0 +1,74 @@
...
+            cls._instance = super(CatalogManager, cls).__new__(cls)
+
+            cls._instance.bootstrap_catalog()
+
+        return cls._instance
+
+    def bootstrap_catalog(self):
+
+        eva_dir = ConfigurationManager().get_value("core", "location")
+        output_url = os.path.join(eva_dir, CATALOG_DIR)
```
saiprashanth173: Aren't catalog_dir_url and output_url the same? 

output_url is only being for logging.

src/catalog/catalog_manager.py
```diff
@@ -0,0 +1,74 @@
...
+            dataset_file_url = os.path.join(catalog_dir_url, DATASET_FILE)
+            dataset_df = load_dataset_df(dataset_file_url)
+            dataset_df.show(1)
+
+            append_row(dataset_file_url, 'foo')
+
+        # Create catalog if it does not exist
+        else:
+
+            print("Catalog not found")
```
saiprashanth173: Can we replace this with logger?

src/catalog/dataset_df.py
```diff
@@ -0,0 +1,117 @@
...
+
+    dataset_df = load_dataset_df(dataset_df_url)
+
+    row_count = dataset_df.count()
+
+    if row_count == 0:
+        dataset_id = 0
+    else:
+        max_id = dataset_df.agg({"dataset_id": "max"}).collect()[0][0]
+        print(max_id)
```
saiprashanth173: replace print with logger

src/catalog/dataset_df.py
```diff
@@ -0,0 +1,117 @@
...
+    rows_rdd = rows_df.rdd
+
+    # Wrap dataset materialization portion.
+    with materialize_dataset(spark,
+                             dataset_df_url,
+                             petastorm_schema):
+
+        spark.createDataFrame(rows_rdd,
+                              petastorm_schema.as_spark_schema()) \
+            .coalesce(1) \
```
saiprashanth173: Shouldn't we be storing data into multiple partitions?

If we assuming that the dataset dataframe is small in size (few 100 records) then this should bee fine. For larger dataset sizes this will not be efficient.

src/catalog/schema.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self, name: str,
+                 type: ColumnType,
+                 is_nullable: bool = False,
+                 array_dimensions: List[int] = []):
+        self._name = name
+        self._type = type
+        self._is_nullable = is_nullable
+        self._array_dimensions = array_dimensions
+
+    def get_name(self):
```
saiprashanth173: can we use @property for getters instead of get_*

src/loaders/frame_loader.py
```diff
@@ -0,0 +1,88 @@
...
+
+def row_generator(x, H, W, C):
+    """Returns a single entry in the generated dataset.
+    Return a bunch of random values as an example."""
+    return {'frame_id': x,
+            'frame_data': np.random.randint(0, 10,
+                                            dtype=np.uint8, size=(H, W, C))}
+
+
+class FrameLoader():
```
saiprashanth173: Is this used for testing?

src/catalog/catalog_manager.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # Get filesystem path
+        catalog_os_path = urlparse(catalog_dir_url).path
+
+        # Check if catalog exists
+        if os.path.exists(catalog_os_path):
+
+            dataset_file_url = os.path.join(catalog_dir_url, DATASET_FILE)
+            dataset_df = load_dataset_df(dataset_file_url)
+            dataset_df.show(1)
```
gaurav274: This can be removed.

src/catalog/dataset_df.py
```diff
@@ -0,0 +1,117 @@
...
+
+from pyspark.sql.functions import (
+    col, max as max_, struct, monotonically_increasing_id
+)
+
+
+def get_dataset_df_schema():
+
+    column_1 = Column("dataset_id", ColumnType.INTEGER, False)
+    column_2 = Column("dataset_name", ColumnType.INTEGER, False)
```
gaurav274: Shouldn't this be ColumnType.STRING

src/catalog/dataset_df.py
```diff
@@ -0,0 +1,117 @@
...
+    print(dataset_id)
+
+    datset_df_schema = get_dataset_df_schema()
+    petastorm_schema = datset_df_schema.get_petastorm_schema()
+    dataset_pyspark_schema = petastorm_schema.as_spark_schema()
+
+    spark = Session().get_session()
+    Session().get_context()
+
+    row_1 = [dataset_id, dataset_id * 500]
```
gaurav274: I can't understand the logic behind 500?

test/catalog/test_schema.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_schema(self):
+
+        schema_name = "foo"
+        column_1 = Column("frame_id", ColumnType.INTEGER, False)
+        column_2 = Column("frame_data", ColumnType.NDARRAY, False, [28, 28])
+        column_3 = Column("frame_label", ColumnType.INTEGER, False)
+
+        schema = Schema(schema_name,
+                        [column_1, column_2, column_3])
+        print(schema)
```
gaurav274: Incomplete test case. Didn't actually test anything.

script/formatting/formatter.py
jarulraj: Uncommented the code.

src/catalog/catalog_manager.py
jarulraj: Cleaned up this code. 

There is an issue with converting file paths to urls like:
`/tmp/eva_datsets` to `file:///tmp/eva_datasets`

I am not sure if this is the correct solution (with urlparse).

src/catalog/catalog_manager.py
jarulraj: Logger output is not visible by default in test cases. We will have to figure out why that is the case.

We should also add a `validator.py` to remove all the `print` functions.

src/catalog/dataset_df.py
jarulraj: Yes, that's correct. Fixed.

src/catalog/dataset_df.py
jarulraj: We need fix up the logger. We will also need to couple it with Spark's logger (log4j) eventually.

src/catalog/dataset_df.py
jarulraj: I created some dummy data.

src/catalog/dataset_df.py
jarulraj: Yes. we should. How do we do that?

src/loaders/frame_loader.py
jarulraj: Yes, they are just placeholders.

script/formatting/formatter.py
```diff
@@ -0,0 +1,233 @@
...
+import os
+import re
+import sys
+import functools
+
+# ==============================================
+# CONFIGURATION
+# ==============================================
+
+# NOTE: absolute path to peloton directory is calculated from current directory
```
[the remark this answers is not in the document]
jarulraj: Fixed.

test/catalog/test_schema.py
jarulraj: Yes, didn't do any programmatic testing. Added a simple test.

PR #62, opened 2020-02-15
Adding githooks functionality, adding linker.sh in .travis

akhileshsiddhanti:
Added githooks support. Scripts formatter.py, validator.py work now after every commit. Moved existing binary checking code to .githooks/pre-commit.d/02-check-binary. Need to run script/linker.sh in the source directory once for the effect. Added this as an instruction in travis as well.

### DIFF
--- a/.githooks/pre-commit.d/01-formatter
+++ b/.githooks/pre-commit.d/01-formatter
@@ -1,0 +1,3 @@
+echo "hi"
+python script/formatting/formatter.py
+python script/formatting/validator.py
--- a/.githooks/pre-commit.d/01-formatter
+++ b/.githooks/pre-commit.d/01-formatter
@@ -1,3 +1,2 @@
-echo "hi"
 python script/formatting/formatter.py
 python script/formatting/validator.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ install:
 
 script:
   - ./script/antlr4/setup.sh
-  - ./script/linker.sh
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -26,6 +26,7 @@ install:
 
 script:
   - ./script/antlr4/setup.sh
+  - ./script/linker.sh
```
gaurav274: I am not sure if we need to add this to the .travis file.

.githooks/pre-commit.d/01-formatter
```diff
@@ -0,0 +1,3 @@
+echo "hi"
```
gaurav274: Drop this echo.

.travis.yml
akhileshsiddhanti: So it is for a one-time run to link .githooks to .git/hooks/ folder. How should we go about it?

.githooks/pre-commit.d/01-formatter
akhileshsiddhanti: Right. Removed.

.travis.yml
gaurav274: My understanding is that since Githooks is more related to commits and Travis is used for incrementally building Eva, we might not need to add git related stuff here.

akhileshsiddhanti: Ok, I have re-edited it.

akhileshsiddhanti:
@gaurav274 the new commit is here

PR #69, opened 2020-03-11
Postmaster (Server + Client)

jarulraj:
A postmaster based on `asyncio` module.

- server (uses command_handler to process a request)
- command_handler
- client (uses interpreter)
- interpreter
- test_server
- test_client


src/server/server.py
```diff
@@ -0,0 +1,139 @@
...
+
+    # These counters are used for realtime server monitoring
+    __connections__ = 0
+    __errors__ = 0
+
+    def connection_made(self, transport):
+        self.transport = transport
+
+        # Set timeout for sockets
+        if not set_socket_io_timeouts(self.transport, 60, 0):
```
saiprashanth173: Shouldn't the timeouts be configurable? 

Maybe, we can use them as attributes in EvaServer? 
eg. EvaServer(timeout_secs=...). By doing this we can read the config only when EvaServer in instantiated.

test/server/test_client.py
```diff
@@ -0,0 +1,58 @@
...
+        def timeout_server():
+            # need a more robust mechanism for when to cancel the future
+            time.sleep(2)
+            self.stop_clients_future.cancel()
+
+        thread = threading.Thread(target=timeout_server)
+        thread.daemon = True
+        thread.start()
+
+        summary = start_clients(client_count=client_count,
```
saiprashanth173: Should we have some assert statements here?

test/server/test_interpreter.py
```diff
@@ -0,0 +1,36 @@
...
+
+from src.server.interpreter import EvaCommandInterpreter
+
+
+class InterpreterTests(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def test_run_event_loop(self):
+
```
saiprashanth173: I assume these tests are to be completed when we finish the interpreter implementation.

src/utils/logging_manager.py
```diff
@@ -60,6 +61,23 @@ def log(self, string, level: LoggingLevel = LoggingLevel.DEBUG):
         elif level == LoggingLevel.CRITICAL:
             self._LOG.critical(string)
 
+    def setEffectiveLevel(self, level: LoggingLevel):
+
+        # Note: pytest logging level cannot be higher than WARNING
+        # https://github.com/segevfiner/pytest/blob/master/
+        # _pytest/logging.py#L246
+
+        if level == LoggingLevel.DEBUG:
```
saiprashanth173: Why do we need these if else conditions? 

Can we use : self._LOG.setLevel(level.value) instead?

test/server/test_server.py
```diff
@@ -0,0 +1,106 @@
...
+                         stop_server_future=self.stop_server_future)
+
+    def test_server_protocol(self):
+
+        eva_server = EvaServer()
+        eva_server.transport = mock.Mock()
+        eva_server.transport.close = MagicMock(return_value="closed")
+        eva_server.transport.abort = MagicMock(return_value="aborted")
+
+        # connection made
```
saiprashanth173: I feel this test should be split into multiple tests.

src/server/server.py
jarulraj: Yes, added it to the eva.yml file.

test/server/test_client.py
jarulraj: Done.

test/server/test_interpreter.py
jarulraj: Yes, we need to expand it.

src/utils/logging_manager.py
jarulraj: Good idea, done.

test/server/test_server.py
jarulraj: Done.

PR #72, opened 2020-03-21
Update command_handler: Adding support for executing queries

akhileshsiddhanti:


src/server/interpreter.py
```diff
@@ -20,19 +20,46 @@
...
 class EvaCommandInterpreter(Cmd):
 
     # Store results from server
     _server_result = None
+    url = None
+
+    def __init__(self):
+        super().__init__()
+
+        # Create table on connecting to server
```
jarulraj: Can we add unit tests in `test/server/interpreter`?

src/server/interpreter.py
```diff
@@ -20,19 +20,46 @@
...
+
+        dummy_info = TableInfo('MyVideo')
+        dummy_table = TableRef(dummy_info)
+
+        columns = [DataFrameColumn('Frame_ID', ColumnType.INTEGER),
```
jarulraj: Can we connect it to the rest of the system (instead of hard-coding the query)?

src/server/interpreter.py
akhileshsiddhanti: Moved things to command_handler, so checking there instead.

src/server/interpreter.py
akhileshsiddhanti: Changed now.

gaurav274:
Duplicate

gaurav274:
Updated with #109

PR #73, opened 2020-03-23
Changes to support delete dataset from catalog

swati21:


src/catalog/catalog_manager.py
```diff
@@ -200,3 +200,50 @@ def get_dataset_metadata(self, database_name: str, dataset_name: str,
...
+        and df_columns respectively
+
+        Arguments:
+           table_name = table name of  to be deleted.
+
+        Returns:
+           Returns the metadata id that will be deleted
+        """
+        metadata_id = self._dataset_service.dataset_by_name(table_name)
+        columns_to_be_deleted = self._column_service.columns_by_id_and_dataset_id(metadata_id, None)
```
gaurav274: It will be better if we define cascade delete in our relationship between columns and dataset and let sqlalchemy handle it, rather than explicitly doing this.

src/catalog/catalog_manager.py
```diff
@@ -200,3 +200,50 @@ def get_dataset_metadata(self, database_name: str, dataset_name: str,
...
+
+        self._column_service.delete_column(columns_to_be_deleted)
+        self._dataset_service.delete_dataset(metadata_id)
+
+        return metadata_id
```
gaurav274: Do we need a return value on delete? I couldn't think of a case where we might need this. Is there a specific reason?

src/catalog/catalog_manager.py
```diff
@@ -200,3 +200,50 @@ def get_dataset_metadata(self, database_name: str, dataset_name: str,
...
+        """
+        This method deletes the columns associated with the given
+        metadata
+
+        Arguments:
+           table_name-[str] table for which we will delete the columns
+           column_names - [List of columns that needs to deleted]
+
+        Retuns:
+           The updated DataFrameMetadata object for which the respective columns are deleted
```
gaurav274: I think we should remove this return part. Let us stick to just deleting the required columns.

src/catalog/services/df_service.py
```diff
@@ -87,3 +87,10 @@ def dataset_object_by_name(self, database_name, dataset_name,
         """
         return self.model.query.filter(
             self.model._name == dataset_name).one()
+        
+    def delete_dataset(self, metadata_id):
+        result = self.model.query \
+            .filter(self.model._id == metadata_id) \
+            .one()
+
```
gaurav274: Use try-catch here. We will be refactoring other places as-well where we are missing it.

PR #74, opened 2020-03-26
Unique column constraint

albert-hen:
Added support for just the unique column constraint, no tests written


src/parser/parser_visitor.py
```diff
@@ -35,6 +35,13 @@
 
 from src.parser.types import ParserColumnDataType
 
+import enum
```
gaurav274: move this enum to src/parser/types.py

src/parser/create_statement.py
```diff
@@ -23,10 +23,12 @@
 
 class ColumnDefinition:
     def __init__(self, col_name: str,
-                 col_type: ParserColumnDataType, col_dim: List[int]):
+                 col_type: ParserColumnDataType, col_dim: List[int],
+                 col_unique_column_constarint=False):
```
gaurav274: Rather than adding a new parameter just for unique, we should add a new generic constraint enum parameter. This way we will not have to add a new parameter for each different type of constraint.

src/parser/parser_visitor.py
```diff
@@ -177,16 +184,26 @@ def visitCreateDefinitions(
...
-            return ColumnDefinition(column_name, data_type, dimensions)
+            return ColumnDefinition(column_name, data_type, dimensions, unique)
 
     def visitColumnDefinition(self, ctx: evaql_parser.ColumnDefinitionContext):
 
         data_type, dimensions = self.visit(ctx.dataType())
-        return data_type, dimensions
+
+        constraint_count = len(ctx.columnConstraint())
+        unique = False
```
gaurav274: Rather than checking whether it is unique or not, we should return the constraint type.

src/parser/types.py
```diff
@@ -15,6 +15,21 @@
 from enum import IntEnum, unique
 
 
+class ColumnConstraintInformation:
+    def __init__(self):
+        self.nullable = False
+        self.default_value = None
+        self.primary = False
+        self.unique = False
+
```
gaurav274: Move from types.py to create_statement.py

src/parser/parser_visitor.py
```diff
@@ -180,16 +182,33 @@ def visitCreateDefinitions(
...
+
+        constraint_count = len(ctx.columnConstraint())
+
+        #unique = False
+        column_constraint_information = ColumnConstraintInformation()
+
+        for i in range(constraint_count):
+            return_type = self.visit(ctx.columnConstraint(i))
+            if return_type == ColumnConstraintEnum.UNIQUE:
+                #unique = True
```
gaurav274: Remove these commented codes

PR #76, opened 2020-04-03
UDF Support

gaurav274:
1. Parser support for both creating udf and using in the predicate.
2. Catalog models and services implemented for UDF.
3. Create UDF Plan and executor implemented.
ToDo:
Functional Expression and seq_scan executor logic has to be changed for an end to end udf support.


src/optimizer/optimizer_utils.py
```diff
@@ -128,8 +130,57 @@ def create_column_metadata(col_list: List[ColumnDefinition]):
...
+def xform_parser_column_type_to_catalog_type(
+        col_type: ParserColumnDataType) -> ColumnType:
+    """translate parser defined column type to the catalog type
+
+    Arguments:
+        col_type {ParserColumnDataType} -- input parser column type
+
+    Returns:
+        ColumnType -- catalog column type
+    """
```
saiprashanth173, line 176: Can we have a mappings file instead?

gaurav274: Done

PR #77, opened 2020-04-04
Changes for Catalog module

swati21:
- Addition of cascade delete
- Testcase for relevant functionalities
- Changes incorporated as suggested in previous PR

### DIFF
--- a/src/catalog/models/base_model.py
+++ b/src/catalog/models/base_model.py
@@ -74,9 +74,12 @@ class CustomModel:
 
     def delete(self):
         """Delete and commit"""
-
-        db_session.delete(self)
-        self._commit()
+        try:
+            db_session.delete(self)
+            self._commit()
+        except Exception:
+            LoggingManager().log("Object couldn't be deleted", LoggingLevel.ERROR)
+            raise Exception
 
     def _commit(self):
         """Try to commit. If an error is raised, the session is rollbacked."""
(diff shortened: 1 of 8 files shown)


src/catalog/services/df_service.py
```diff
@@ -87,3 +87,19 @@ def dataset_object_by_name(self, database_name, dataset_name,
         """
         return self.model.query.filter(
             self.model._name == dataset_name).one()
+
```
gaurav274, line 90: Add test case for this implementation

src/catalog/models/base_model.py
```diff
@@ -74,9 +74,12 @@ def update(self, **kwargs):
 
     def delete(self):
         """Delete and commit"""
-
-        db_session.delete(self)
-        self._commit()
+        try:
```
gaurav274, line 77: Add test case for this function. I think we don't have any unit testing for this file. Can you please create on and add test cases?

src/catalog/services/df_service.py
swati21: def (test_dataset_object_by_name_queries_with_name_returns_model_object) this testcasse is already present in the test_dataset_service.py

src/catalog/models/base_model.py
swati21: Yes we will need to create testcases for Base Model. Will add that

PR #84, opened 2020-04-20
Update the data structures to use Pandas DataFrames

saiprashanth173:
System Changes:
1. Make FrameBatch into a more generic data structure. The current implementation will store data as Pandas DataFrame. 
2. Define a new data structure for returning output from functions. This object again stores function output as pandas DataFrame. With this, we will be able to support functions with different outputs.
3. Update Plan Executor to return a Batch as output instead of a list of Batches.
4. Update UDFs, now every UDF will take as input a column vector instead of whole FrameBath. 
5. Add UDF binding utility, for integrating UDFs in SELECT queries. Also, added the catalog implementations to support this. 
6. Fix TupleValueExpression to deal with Batches instead of an 'unknown' list input type which was never supported by the system. 

------------------------------------------------
With these changes, we can now execute queries like:
SELECT FastRCNNObjectDetector(frame_data) FROM test WHERE FastRCNNObjectDetector(frame_data) = 'car' AND frame_id > 2; 

-------------------------------------------------

Refactoring:
FrameBatch is now called Batch as it supports any type of column.

### DIFF
--- a/src/models/inference/classifier_prediction.py
+++ b/src/models/inference/classifier_prediction.py
@@ -30,7 +30,9 @@ class Prediction(BasePrediction):
 
     """
 
-    def __init__(self, frame: Frame, labels: List[str], scores: List[float],
+    def __init__(self, frame: Frame,
+                 labels: List[str],
+                 scores: List[float],
                  boxes: List[BoundingBox] = None):
         self._boxes = boxes
         self._labels = labels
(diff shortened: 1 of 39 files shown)


src/models/inference/classifier_prediction.py
```diff
@@ -30,7 +30,9 @@ class Prediction(BasePrediction):
 
     """
```
gaurav274, line 32: Are we planning to implement prediction for each kind of UDF?

saiprashanth173: This is an old mechanism, outputs for UDFs are replaced with Outcome. I need to check where it's being used and delete if not needed.

PR #85, opened 2020-04-30
Splitting parser_visitor.py

albert-hen:
Follows this method:
https://stackoverflow.com/questions/47561840/python-how-can-i-separate-functions-of-class-into-multiple-files

One issue is that this requires adding an import statement in \_\_init\_\_.py every time a new function is added. This is because of a limitation in python, `import * only allowed at module level`.


src/parser/evaql/antlr4.bat
```diff
@@ -0,0 +1 @@
+java org.antlr.v4.Tool %*
```
gaurav274, line 1: Can you please remove this file?

src/parser/parser_visitor/__init__.py
```diff
@@ -0,0 +1,51 @@
+
```
gaurav274, line 1: The idea of spreading class functionalities across multiple files: [stack overflow](https://stackoverflow.com/questions/47561840/python-how-can-i-separate-functions-of-class-into-multiple-files). I'm not sure if there is a better way of approaching this problem. One downside is that since we can't use * to import, we are forced to import each function which looks clumsy. @saiprashanth173 @jarulraj Any suggestions?

saiprashanth173: @albert-hen good job with the refactoring. 
Another way to achieve this is using decorators. You can check this link.
https://medium.com/@mgarod/dynamically-add-a-method-to-a-class-in-python-c49204b85bd6


We can have our class set up with the basic methods required. Then we can use these hooks to add methods dynamically. 

Things might look like this:
src/parser/parser_visitor/visitors -> will have all the visitor files (containing methods with @add_method decorator)
src/parser/parser_visitor/visitors/__init__.py -> we can do an - import * of all the files in this folder, every time we add a new .py file we might have to add a new **import *** this file. 
src/parser/parser_visitor/__init__.py -> this file will expose the ParserVisitor class now.
from src.parser.parser_visitor import ParserVisitor
from src.parser.parser_visitor.visitors import *

----------------------------------------------
With the above setup, we should be able to import ParserVisitor anywhere as. 
from src.parser import ParserVisitor

All the visitor methods should be accessible from this class now.

**Drawbacks:**
This approach is not IDE friendly as methods are dynamically added to the class.

albert-hen: @saiprashanth173 @gaurav274 Thanks for the suggestion. What about using multiple inheritance described by the second reply in the stack overflow link. We define the `ParserVisitor(partA, partB, partC)` class with super-classes in which we keep the function definitions. Those super classes would then be sub-classes of `evaql_parserVisitor`. I'm not sure if this diamond inheritance pattern will work though. But this seems to be the cleanest way if it does works, right?

gaurav274:
Code was rewritten using multiple Inheritance.

PR #87, opened 2020-05-26
Command Handler PR Fix

jarulraj:


src/parser/parser_visitor/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+        visitInsertStatementValue
+
+    from ._create_statements import visitColumnCreateTable,\
+        visitCreateDefinitions,\
+        visitColumnDeclaration,\
+        visitColumnDefinition,\
+        visitUniqueKeyColumnConstraint,\
+        visitSimpleDataType
+
+    from ._create_statements import visitIntegerDataType,\
```
saiprashanth173, line 33: Would this approach work if we import it as:
from ._create_statements import *

src/parser/parser_visitor/_expressions.py
```diff
@@ -0,0 +1,107 @@
...
+
+from src.parser.evaql.evaql_parser import evaql_parser
+
+
+##################################################################
+# EXPRESSIONS
+##################################################################
+
+def visitStringLiteral(self, ctx: evaql_parser.StringLiteralContext):
+    # Fix a bug here; 'VAN' Literal gets converted to "'VAN'";
```
saiprashanth173, line 31: I guess we fixed this bug? @gaurav274

src/server/client.py
```diff
@@ -99,6 +99,7 @@ def data_received(self, data):
                              )
 
         self._response_chunk = response_chunk
+        print(response_chunk)
```
saiprashanth173, line 102: Can we remove print statements?

src/server/command_handler.py
```diff
@@ -17,17 +17,44 @@
...
+
+    convertor = StatementToPlanConvertor()
+    convertor.visit(insert_statement)
+
+    convertor.plan
+
+    eva_statement = parser.parse(query)
+    response_message = ""
+
+    for i in range(len(eva_statement)):
```
saiprashanth173, line 50: Any reason for not iterating over the statements directly?

for stmt in eva_statement: ....

Also, can we rename, eva_statement to eva_statements -> as it's a list of statements?

src/server/command_handler.py
```diff
@@ -17,17 +17,44 @@
...
 
-    response_message = "foo"
+        LoggingManager().log("Parse Tree: " + str(eva_statement[i]))
+        logical_plan = StatementToPlanConvertor().visit(eva_statement[i])
+        physical_plan = PlanGenerator().build(logical_plan)
+        df = PlanExecutor(physical_plan).execute_plan()
+        statement_response_message = ','.join(map(str, df.collect()))
+        response_message += statement_response_message
```
saiprashanth173, line 57: It would be good to represent response_message as list.
Where each entry in the list corresponds to a statement in eva_statements.

test/server/test_command_handler.py
```diff
@@ -35,9 +35,16 @@ def __init__(self, *args, **kwargs):
     def test_command_handler(self):
         transport = mock.Mock()
```
saiprashanth173, line 36: Not sure what we are testing here. 

Command handler logic for executing the whole pipeline and this test should be testing if everything happens as expected.

test/server/test_client.py
```diff
@@ -53,6 +54,8 @@ def timeout_server():
                                 loop=self.loop,
                                 stop_clients_future=self.stop_clients_future)
 
+        print(summary)
```
saiprashanth173, line 57: Please remove print statements.

src/server/command_handler.py
```diff
@@ -17,17 +17,44 @@
...
+    convertor.plan
+
+    eva_statement = parser.parse(query)
+    response_message = ""
+
+    for i in range(len(eva_statement)):
 
-    response_message = "foo"
+        LoggingManager().log("Parse Tree: " + str(eva_statement[i]))
+        logical_plan = StatementToPlanConvertor().visit(eva_statement[i])
```
saiprashanth173, line 53: I feel this logic goes into a different file. We can call it a driver. 

According to me, the command handler should only be responsible for:
1.  instantiating the Driver code 
2. Take user queries as input and pass it to the driver
3. Take the output of the driver, and convert it into any format for displaying/ data transfer.

Command handler should only be an interface between User Interface layer and the Core Eva code. It should be lightweight.

src/parser/parser_visitor/__init__.py
gaurav274: It doesn't. There is an alternative using multiple inheritances. For instance, createStatement(evaql_parserVisitor)... and then ParserVisitor(createStatement, ..).
Should we move to this approach?

src/parser/parser_visitor/_expressions.py
gaurav274: Yup, we did. We can remove the comment.

src/server/client.py
gaurav274: Sure.

src/server/command_handler.py
gaurav274: I agree. We should rewrite this.

src/server/interpreter.py
```diff
@@ -13,91 +13,61 @@
...
 
     def set_protocol(self, protocol):
         self.protocol = protocol
 
     def do_greet(self, line):
         print("greeting")
 
-    def onecmd(self, s):
+    def emptyline(self):
+        print("Enter a valid query.")
```
gaurav274, line 38: Do the print statements in this file display on the client's prompt? @jarulraj Shouldn't we do something like self.stdout.write()?

src/server/command_handler.py
gaurav274: It makes sense to keep response_messages as a list but I'm worried if asyncio message passing allows list. In case it doesn't we have to convert the list back to string.

gaurav274:
Duplicate PR89

PR #88, opened 2020-06-17
Enable GPU execution

saiprashanth173:
- Add a mechanism to enable GPU execution
- Implement a Pytorch UDF abstract class that can be used by all Pytorch based UDFs. This UDF abstract class will take care of moving models and Data to GPU.


src/udfs/pytorch_abstract_udf.py
```diff
@@ -0,0 +1,83 @@
...
+        Arguments:
+             val (Tensor): tensor to be converted
+        Returns:
+            np.ndarray: numpy array representation
+        """
+        return val.detach().cpu().numpy()
+
+    def to_device(self, device: str):
+        """
+
```
gaurav274, line 76: Please add the doc string.

PR #92, opened 2020-07-14
Load data

xzdandy:
- Replace old dataframe API with new storage engine. 
- Integration test for load. 
- Remove Loaders and uses readers.


src/executor/disk_based_storage_executor.py
```diff
@@ -32,17 +32,22 @@ class DiskStorageExecutor(AbstractStorageExecutor):
...
         super().__init__(node)
+        """
         self.storage = Loader(node.video,
                               batch_size=node.batch_size,
                               skip_frames=node.skip_frames,
                               limit=node.limit,
                               offset=node.offset,
                               curr_shard=node.curr_shard,
                               total_shards=node.total_shards)
+        """
```
gaurav274, line 43: We should get away with these comments. And do you think we still need a disk-based storage executor? Our storage engine should ideally replace this.

src/executor/load_executor.py
```diff
@@ -50,4 +50,6 @@ def exec(self):
                 data.append({'id': id, 'data': frame})
                 id += 1
         # Hook for the storage engine
-        append_rows(self.node.table_metainfo, data)
+        # We currently use create to empty exsiting table.
+        StorageEngine.create(self.node.table_metainfo)
+        StorageEngine.write_row(self.node.table_metainfo, data)
```
gaurav274, line 55: Move this code chunk inside the video_reader loop. So basically we read a batch we store that batch. And can we change the API name to write_batch? That will remove confusion between row and batch.

src/storage/petastorm_storage_engine.py
```diff
@@ -0,0 +1,137 @@
...
+                                 table.schema.petastorm_schema):
+
+            self.spark_session.createDataFrame(empty_rdd,
+                                               table.schema.pyspark_schema) \
+                .coalesce(1) \
+                .write \
+                .mode('overwrite') \
+                .parquet(self._spark_url(table))
+
+    def write_row(self, table: DataFrameMetadata, rows: []):
```
gaurav274, line 66: We need to change this APIs name. write_batch makes move sense I think

src/storage/petastorm_storage_engine.py
```diff
@@ -0,0 +1,137 @@
...
+                .parquet(self._spark_url(table))
+
+    def read(self, table: DataFrameMetadata) -> Iterator[Dict]:
+        """
+        Read the dataframe.
+
+        Return:
+            Iterator of Dict. Each Dict represents a Row.
+        """
+        with make_reader(self._spark_url(table)) as reader:
```
gaurav274, line 96: We are accidentally creating a new file for every read. Actually the table object should have the file path, we need to convert that to uri before feeding to the petastorm reader.

src/storage/petastorm_storage_engine.py
```diff
@@ -0,0 +1,137 @@
...
+                    str], predicate_func) -> Iterator[Dict]:
+        """
+        Read the rows that the predicate_func returns true on the given fields.
+
+        Argument:
+            fields List[str]: A list of fields (column names) to be considered.
+            predicate_func: customized function return bool
+        """
+        predicate = in_lambda(fields, predicate_func)
+        with make_reader(self._spark_url(table),
```
gaurav274, line 110: same as the read

src/storage/petastorm_storage_engine.py
gaurav274: We have the petastorm reader. I think we should use that class here instead of make_reader. Maybe add the predicate parameter support in the Petastorm reader since we need that for lambda.

src/storage/petastorm_storage_engine.py
```diff
@@ -0,0 +1,137 @@
...
+                 values: []) -> Iterator[Dict]:
+        """
+        Read the rows that matches the field's given values.
+
+        Argument:
+            field str: name of the field (column name). E.g., "id".
+            values List[]: A list of values to be included.
+        """
+        predicate = in_set(values, field)
+        with make_reader(self._spark_url(table),
```
gaurav274, line 125: same as read

test/executor/test_load_executor.py
```diff
@@ -42,10 +43,11 @@ def test_should_call_opencv_reader_and_storage_engine_and_config(
         append_mock.assert_called_once_with(table_metainfo, data)
```
gaurav274, line 43: We need to add a check that create_mock is called once with.
and even append_mock might change if we make changes in the load executor.

test/executor/test_load_executor.py
```diff
@@ -42,10 +43,11 @@ def test_should_call_opencv_reader_and_storage_engine_and_config(
...
 
     @patch('src.executor.load_executor.OpenCVReader')
-    @patch('src.executor.load_executor.append_rows')
+    @patch('src.executor.load_executor.StorageEngine.create')
+    @patch('src.executor.load_executor.StorageEngine.write_row')
     @patch('src.executor.load_executor.ConfigurationManager.get_value')
     def test_exec_config_returns_None(
-            self, get_val_mock, append_mock, cv_mock):
+            self, get_val_mock, append_mock, create_mock, cv_mock):
```
gaurav274, line 50: same as above.

test/integration_tests/test_insert_executor.py
```diff
@@ -60,7 +59,8 @@ def test_should_insert_row_in_table(self):
         insertExec.exec()
 
         # test if we have a added the in our storage
-        df = load_dataframe(url)
+        df = None  # will fix later
+        # df = load_dataframe(url)
```
gaurav274, line 63: I think this is still a work in progress. So we should drop this for now and maybe add with a later PR.

test/integration_tests/test_create_executor.py
```diff
@@ -35,12 +38,10 @@ def test_create_executor_should_create_table_in_storage(self):
...
         createExec = CreateExecutor(plan_node)
-        url = createExec.exec()
+        metadata = createExec.exec()
 
-        # test if we have a table created in our storage
-        df = load_dataframe(url)
-        self.assertEqual(2, len(df.columns))
-        self.assertEqual(df.columns, ['id', 'name'])
+        row_iter = StorageEngine.read(metadata)
+        self.assertFalse(any(True for _ in row_iter))
```
gaurav274, line 44: I think this is still a work in progress. So we should drop this for now and maybe add with a later PR.

test/integration_tests/test_load_executor.py
```diff
@@ -0,0 +1,91 @@
...
+        convertor = StatementToPlanConvertor()
+        convertor.visit(load_stmt)
+        logical_plan_node = convertor.plan
+        phy_plan_node = LoadDataPlan(
+            logical_plan_node.table_metainfo,
+            logical_plan_node.path)
+
+        loadExec = LoadDataExecutor(phy_plan_node)
+        loadExec.exec()
+
```
gaurav274, line 78: We don't need to make specific calls to LoadDataPlan and LoadDataExecutor.
following chunk of code should do the magic.    
`stmt = Parser().parse(query)[0]`
`l_plan = StatementToPlanConvertor().visit(stmt)`
`p_plan = PlanGenerator().build(l_plan)`
 `PlanExecutor(p_plan).execute_plan()`

test/integration_tests/test_load_executor.py
```diff
@@ -0,0 +1,91 @@
...
+        # Do we have select command now?
+
+        return_rows = list(
+            StorageEngine.read(
+                logical_plan_node.table_metainfo))
```
gaurav274, line 83: Get this info using the catalogmanager(). That way we can check that even the catalog has a valid entry.

src/executor/disk_based_storage_executor.py
xzdandy: Yeah, we can get rid of it.

src/storage/petastorm_storage_engine.py
xzdandy: Only change the API name or change the implementation?
I would prefer keeping row as dict for now.

src/executor/create_executor.py
```diff
@@ -47,5 +47,5 @@ def exec(self):
                                                     file_url,
                                                     self.node.column_list)
 
-        create_dataframe(metadata)
-        return file_url
+        StorageEngine.create(table=metadata)
```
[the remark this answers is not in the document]
xzdandy: Fixed by 79651a5

test/executor/test_load_executor.py
xzdandy: Fixed by 79651a5

test/executor/test_load_executor.py
xzdandy: Fixed by 79651a5

test/integration_tests/test_insert_executor.py
xzdandy: Fixed by 6de611c

test/integration_tests/test_create_executor.py
xzdandy: Fixed by 6de611c

src/storage/petastorm_storage_engine.py
gaurav274: We can return a batch object.

test/integration_tests/test_load_executor.py
xzdandy: Fixed by 7185fc2

test/integration_tests/test_load_executor.py
xzdandy: Fixed by 7185fc2

src/storage/petastorm_storage_engine.py
saiprashanth173: Just a suggestion, we can just name it `write`. `rows` is anyway a parameter.

src/storage/petastorm_storage_engine.py
saiprashanth173: Why have two methods in this case? why not just have `read` which takes optional params. `fields` and `predicate_func`. Technically if `fields` are not provided, shouldn't we take all the fields in the table?

src/storage/petastorm_storage_engine.py
saiprashanth173: Quick question (probably, I don't have the context): Where are we taking care of the `shard` we need to read?

gaurav274: We need to implement a storage context to take care of the sharding. Probably delaying that logic to late PR. Right now assuming we have a single server.

src/storage/petastorm_storage_engine.py
gaurav274: Makes sense.
Btw fields is something required by predicate_func

src/storage/petastorm_storage_engine.py
gaurav274: Yeah

PR #93, opened 2020-07-15
Readers support batch

gaurav274:
Readers will return Batch instead of raw data.


src/models/storage/frame.py
```diff
@@ -47,3 +47,6 @@ def __eq__(self, other):
         return self.index == other.index and \
             np.array_equal(self.data, other.data) and \
             self.info == other.info
+
+    def asdict(self):
+        return {'id': self._id, 'data': self._data, 'info': self._info}
```
xzdandy, line 52: Need to also update the "info" into column scheme. 
There should be a better way doing this. Otherwise, every time we change the row structure (likely when adding more functionality), we need to manually sync across the code base.

src/models/storage/batch.py
```diff
@@ -18,31 +18,37 @@
...
         super().__init__()
         if outcomes is None:
             outcomes = dict()
         if temp_outcomes is None:
             temp_outcomes = dict()
-
-        self._frames = frames
+        if isinstance(frames, DataFrame):
+            self._frames = frames
+        elif isinstance(frames, list):
```
saiprashanth173, line 46: I don't think this check is as simple as checking for type. 
For instance, ["string", 1, 2..] is a list. Also, [{'name': 1, 'data':[]}, {'col2': '..', 'col3': ...}] is a list.

I feel it's much easier to convey that the Batch API takes as input a pd.DataFrame. How it created is taken care by the API user outside.

src/models/storage/frame.py
saiprashanth173: I remember removing the usage of `Frame`. It was used in the old version of `Batch` implementation.

The purpose of `pd.DataFrame` based `Batch` implementation was extensibility and supporting different row structures.

saiprashanth173: I guess Frame is only being used in some tests and as a param typing hint. I was supposed to remove it from the code base when I changed the `Batch` implementation but left it as such in view of time. Feel free to delete this part.

src/readers/opencv_reader.py
```diff
@@ -14,31 +14,43 @@
...
         LoggingManager().log("Reading frames", LoggingLevel.INFO)
 
         _, frame = video.read()
-        frame_ind = video_start - 1
+        frame_id = self._start_frame_id
 
         while frame is not None:
-            frame_ind += 1
-            yield frame
+            yield Frame(frame_id, frame, None).asdict()
```
saiprashanth173, line 54: I feel using `Frame` for just conversion (asdict) is an overkill. We can instead create the dictionary directly.

test/models/test_prediction.py
```diff
@@ -27,7 +27,7 @@ class PredictionTest(unittest.TestCase):
 
     def test_should_check_if_batch_frames_equivalent_to_number_of_predictions(
             self):
-        batch = Batch(frames=[Frame(1, np.ones((1, 1)), None)])
```
saiprashanth173, line 30: This test was supposed to be updated to work with the `pd.DataFrame` change.

test/models/test_prediction.py
```diff
@@ -56,11 +56,12 @@ def test_should_check_if_batch_frames_equivalent_to_no_of_boxes_if_given(
...
                               boxes=boxes))
 
+    @unittest.skip('Depricated should remove')
     def test_should_return_list_of_predictions_for_each_frame_in_batch(self):
-        batch = Batch(frames=[Frame(1, np.ones((1, 1)), None)])
+        batch = Batch(frames=[Frame(1, np.ones((1, 1)), None).asdict()])
         predictions = [['A', 'B']]
         scores = [[1, 1]]
-        expected = [Prediction(batch.frames[0], predictions[0], scores[0])]
+        expected = [Prediction(batch.frames, predictions[0], scores[0])]
```
saiprashanth173, line 64: `Prediction` was replaced with a more general class. `Outcome`. This used `pd.DataFrames` again. 

We can remove `Prediction` from the code base.

gaurav274: Prediction Removed

test/models/test_prediction.py
gaurav274: Prediction Removed

src/models/storage/batch.py
gaurav274: Done

src/readers/opencv_reader.py
gaurav274: The intuition behind such an approach was to get rid of hardcoding the column names here.
But since we are removing Frames anyway so creating a dictionary directly.

src/models/storage/frame.py
gaurav274: Removed

src/models/storage/batch.py
```diff
@@ -18,31 +18,36 @@
...
+        frames (List[Dict], DataFrame): List of frames
         outcomes (Dict[str, List[BasePrediction]]): outcomes of running a udf
         with name 'x' as key
         identifier_column (str): A column used to uniquely a row
 
 
     """
 
-    def __init__(self, frames: DataFrame, outcomes=None, temp_outcomes=None,
+    def __init__(self, frames, outcomes=None, temp_outcomes=None,
```
xzdandy, line 37: keep the frames type as DataFrame

xzdandy: Also check whether line 44-50 is necessary.

gaurav274: Fixed.
Those lines are enforcing that frames are indeed dataframe.  So keeping it

PR #94, opened 2020-08-01
New Feature: Petastorm Storage Engine and Load statement

gaurav274:
- Petastorm storage engine added to replace basic dataframe APIs
- The load statement is supported. We can now load frames from a video path into our storage engine


src/models/storage/batch.py
```diff
@@ -41,8 +41,9 @@ def __init__(self, frames, outcomes=None, temp_outcomes=None,
             outcomes = dict()
         if temp_outcomes is None:
             temp_outcomes = dict()
+        # store the batch with columns sorted
         if isinstance(frames, DataFrame):
-            self._frames = frames
+            self._frames = frames[sorted(frames.columns)]
```
jarulraj, line 46: Can any operator leave it unsorted?

src/storage/petastorm_storage_engine.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        self._spark = Session()
+        self.spark_session = self._spark.get_session()
+        self.spark_context = self._spark.get_context()
+
+    def _spark_url(self, table: DataFrameMetadata) -> str:
+        """
+        Generate a spark/petastorm url given a table
+        """
+        return Path(table.file_url).resolve().as_uri()
```
jarulraj, line 43: This looks like a cleaner way to generate uris. Is the `datasets` location also set in this manner?

test/integration_tests/test_load_executor.py
```diff
@@ -0,0 +1,78 @@
...
+from src.storage import StorageEngine
+from src.models.storage.batch import Batch
+
+
+NUM_FRAMES = 10
+
+
+class LoadExecutorTest(unittest.TestCase):
+
+    def create_sample_video(self):
```
jarulraj, line 35: it looks like this function can be reused across multiple test cases. Should we move it a different file (like `src/utils/`)?

src/models/storage/batch.py
gaurav274: The Readers while creating a batch from video frames can leave it unsorted. Either we add this constraint while reading or I thought of adding it in the constructor itself.

src/storage/petastorm_storage_engine.py
gaurav274: Yeah, we are leveraging pathlib.Path for every system path in our system.

test/integration_tests/test_load_executor.py
gaurav274: Sure we can move it to the test utils

src/models/storage/batch.py
gaurav274: Batch(pd.DataFrame({'id': 1, 'data': 2})). The batch object is typically creating in this manner. So the order of columns is dependent on the order of keys in the dictionary. So to avoid inconsistency across the system I added the sorted requirement.

jarulraj: Sounds good. We could add a sort operator later.

PR #95, opened 2020-08-01
Fix integration test for load

xzdandy:


test/integration_tests/test_load_executor.py
```diff
@@ -47,5 +47,5 @@ def test_should_load_video_in_table(self):
         # Do we have select command now?
         metadata = CatalogManager().get_dataset_metadata("", "MyVideo")
         actual_batch = list(StorageEngine.read(metadata))[0]
-        expected_batch = list(create_dummy_batches())
+        expected_batch = list(create_dummy_batches())[0]
```
gaurav274, line 50: Why are picking index 0?

test/integration_tests/test_load_executor.py
```diff
@@ -35,7 +35,7 @@ def tearDown(self):
         os.remove('dummy.avi')
 
     # integration test
-    @unittest.skip("we need drop functionality before we can enable")
+    # @unittest.skip("we need drop functionality before we can enable")
```
gaurav274, line 38: We need to wait for drop. I have checked locally it works.

xzdandy:
I think travis does not work for me due to the mysql connection problem.

test/integration_tests/test_load_executor.py
xzdandy: Because it returns a list of batch. You can either remove or keep [0] index for both actual_batch and expected_batch.

test/integration_tests/test_load_executor.py
xzdandy: Drop is only needed for "Create". By the current way load is implemented, Drop is not required.

test/integration_tests/test_load_executor.py
gaurav274: Actual_batch is also a list and so is expected_batch
We can compare lists

test/integration_tests/test_load_executor.py
gaurav274: Yup sure got it.

test/integration_tests/test_load_executor.py
xzdandy: yeah, right. But i think in the current master, acutal_batch is "Batch" and expected_batch is list, so this testcase will fail.

test/catalog/test_catalog_manager.py
```diff
@@ -26,8 +26,7 @@ class CatalogManagerTests(unittest.TestCase):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-    @mock.patch('src.catalog.catalog_manager.init_db')
-    def test_catalog_manager_singleton_pattern(self, mocked_db):
+    def test_catalog_manager_singleton_pattern(self):
```
xzdandy, line 29: CatalogManager will be reused by other testcases. So init_db has to be called here.

xzdandy: Also considering removing the rest of @mock.path in this file. Since once the catalogmanaer is created, the stored instance will be directly returned and it will not be initialized again.

xzdandy:
Merged into #96

PR #96, opened 2020-08-01
Catalog Drop functionality Added 

gaurav274:
- The catalog now supports dropping table metadata(dropping a specific column in a table is not enabled)
- The catalog now supports dropping udfs
- We can reset the catalog manager state using reset function
- Missing test cases for catalog added 
- Integration test case added


src/catalog/models/udf_io.py
```diff
@@ -103,3 +103,12 @@ def __str__(self):
         column_str += ")\n"
 
         return column_str
+
+    def __eq__(self, other):
```
jarulraj, line 107: Does python generate a copy constructor by default?

src/catalog/catalog_manager.py
```diff
@@ -256,6 +256,40 @@ def create_udf(self, name: str, impl_file_path: str,
         self._udf_io_service.add_udf_io(udf_io_list)
         return metadata
```
jarulraj, line 257: We could add a README in this directory listing the schemata of all the catalog tables.

src/catalog/models/udf_io.py
gaurav274: I don't think so. In python assignment works as a binding of the target with the original object.

src/catalog/catalog_manager.py
gaurav274: Yeah, even I feel the need. I will add this to the ToDos. I will skip this now as we need this PR to enable integration test cases.

src/catalog/models/df_metadata.py
```diff
@@ -27,7 +27,8 @@ class DataFrameMetadata(BaseModel):
     _unique_identifier_column = Column('identifier_column', String(100))
 
     _columns = relationship('DataFrameColumn',
-                            back_populates="_dataset")
+                            back_populates="_dataset",
+                            cascade='all, delete, delete-orphan')
```
xzdandy, line 31: A quick question, what is cascade='all, delete, delete-orphan' for?

src/catalog/catalog_manager.py
```diff
@@ -256,6 +256,20 @@ def create_udf(self, name: str, impl_file_path: str,
...
 
+    def delete_metadata(self, table_name: str) -> bool:
```
xzdandy, line 259: Is this the only function required to drop a created udf? And table_name = udf_name?

xzdandy:
Is it worth having a function to drop/reinstate Catalogmanager on request. Current code base shares one Catalogmanager, and once it is created upon the first call, all the following calls reuse the created one. 
This creates problem. For example, in test_catalog_manager.py, init_db, DatasetService, DatasetColumnService are mocked functions, so the created Catalogmanager is not functional. However, the later integration test will reuse this dysfunctional Catalogmanager by default.

src/catalog/catalog_manager.py
gaurav274: No, this is to delete the table metadata. I will also add one for the UDF.

src/catalog/catalog_manager.py
```diff
@@ -48,7 +48,17 @@ def __init__(self):
         self._udf_service = UdfService()
         self._udf_io_service = UdfIOService()
 
-    def bootstrap_catalog(self):
+    def reset(self):
```
xzdandy, line 51: Could you please check locally? reset function can not be called twice during my test.

xzdandy: BTW, line 34-35 are redundant code? They are not used anywhere.

gaurav274: Removed

gaurav274: it now works on my system. Can you please verify?

xzdandy: Verified

PR #99, opened 2020-08-20
Fix problems when batch size is small

xzdandy:
- append ignores index 
- sort returned batch in integration test cases
- Not yield empty batch

### DIFF
--- a/src/executor/plan_executor.py
+++ b/src/executor/plan_executor.py
@@ -27,6 +27,7 @@ from src.executor.load_executor import LoadDataExecutor
 from src.executor.storage_executor import StorageExecutor
 from src.utils.logging_manager import LoggingManager, LoggingLevel
 
+
 class PlanExecutor:
     """
     This is an interface between plan tree and execution tree.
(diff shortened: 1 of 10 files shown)


src/executor/seq_scan_executor.py
```diff
@@ -51,4 +51,5 @@ def exec(self) -> Iterator[Batch]:
                 batches = [expr.evaluate(batch) for expr in self.project_expr]
                 batch = Batch.merge_column_wise(batches)
 
-            yield batch
+            if not batch.empty():
```
gaurav274, line 54: Do we need this check? Yielding an empty batch might make things easy for the caller

src/executor/plan_executor.py
```diff
@@ -106,7 +107,11 @@ def execute_plan(self):
             execution_tree.exec()
         else:
             for batch in execution_tree.exec():
-                output_batches += batch
+                if not batch.empty():
```
gaurav274, line 110: We can move this check to the _add_ function

src/executor/seq_scan_executor.py
xzdandy: I move this check to when we apply predicate, and it also works.

src/executor/plan_executor.py
xzdandy: Move to the add function.

PR #101, opened 2020-09-04
Add support for UNION operator

xzdandy:
1. Currently, only UNION ALL is supported, no deduplication. 
2. Fix a bug in DummyObjectDetector, so under any batch size, it can return the correct label.

### DIFF
--- a/src/planner/union_plan.py
+++ b/src/planner/union_plan.py
@@ -16,6 +16,7 @@
 from src.planner.types import PlanNodeType
 from src.planner.abstract_plan import AbstractPlan
 
+
 class UnionPlan(AbstractPlan):
     """
     This plan is used for storing information required for union operations.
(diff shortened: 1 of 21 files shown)


src/planner/union_plan.py
```diff
@@ -0,0 +1,37 @@
...
+from src.planner.types import PlanNodeType
+from src.planner.abstract_plan import AbstractPlan
+
+
+class UnionPlan(AbstractPlan):
+    """
+    This plan is used for storing information required for union operations.
+
+    Arguments:
+        column_ids: List[AbstractExpression]
```
gaurav274, line 25: Change this

gaurav274 (commented):
LGTM

src/planner/union_plan.py
xzdandy: Fix in a365f0d

gaurav274 (approved):

PR #105, opened 2020-09-30
pytorch and fastrcnn

xzdandy:
Add a new integration test cases for data/ua_detrac.mp4 and src/udfs/fastrcnn_object_detector.py.

Tested on Ada-01 with GPU. 

Add gpu_batch_size in eva.yml

### DIFF
--- a/src/udfs/fastrcnn_object_detector.py
+++ b/src/udfs/fastrcnn_object_detector.py
@@ -16,11 +16,13 @@ from typing import List
 
 import pandas as pd
 import torchvision
+import torch
 
 from torch import Tensor
 from src.models.catalog.frame_info import FrameInfo
 from src.models.catalog.properties import ColorSpace
 from src.udfs.pytorch_abstract_udf import PytorchAbstractUDF
+from src.configuration.configuration_manager import ConfigurationManager
 
 
 class FastRCNNObjectDetector(PytorchAbstractUDF):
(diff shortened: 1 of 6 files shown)


src/udfs/pytorch_abstract_udf.py
```diff
@@ -43,21 +44,25 @@ def get_device(self):
...
+    def transform(self, images: np.ndarray):
+        # reverse the channels from opencv
+        return self.transforms(Image.fromarray(images[:, :, ::-1]))\
+            .unsqueeze(0)
 
-    def forward(self, frames: np.ndarray):
-        return self.classify([self.transform(frame[0]) for frame in frames])
+    def forward(self, frames: List[np.ndarray]):
+        tens_batch = torch.cat([self.transform(x) for x in frames])\
+            .to(self.get_device())
```
gaurav274, line 54: Is this done to speed up the data transfer?

gaurav274 (commented):
Travis won't run a GPU test so we might have to skip it on the master build

src/udfs/pytorch_abstract_udf.py
xzdandy: This is batched (multiple frames) model for pytorch. But From the latency numbers I get.  I do not see performance improvement.

xzdandy:
Hopefully, gpu_batch_size 1 can work on travis with cpu version of pytorch.

src/udfs/fastrcnn_object_detector.py
```diff
@@ -105,12 +105,25 @@ def _get_predictions(self, frames: Tensor) -> List[Outcome]:
...
+                ignore_index=True)
+        return outcome
+
+    def classify(self, frames: Tensor) -> pd.DataFrame:
+        gpu_batch_size = ConfigurationManager()\
+            .get_value('executor', 'gpu_batch_size')
 
-    def classify(self, frames: Tensor) -> List[Outcome]:
-        return self._get_predictions(frames)
+        if gpu_batch_size:
```
gaurav274, line 121: Can we move this to the abstract class so that it happens automatically for every UDF?

xzdandy: Done in fe148be

gaurav274 (approved):

PR #107, opened 2020-10-06
Java memory config

xzdandy:
- Move all pyspark config to eva.yml
- Add java memory config.

Tested when spark.driver.memory is 4g, batch size 100 works. 
For local testing and travis, keep the default value of spark.driver.memory as 1g, where batch size 50 works.

### DIFF
--- a/test/integration_tests/test_select_executor.py
+++ b/test/integration_tests/test_select_executor.py
@@ -60,7 +60,7 @@ class SelectExecutorTest(unittest.TestCase):
         expected_batch = list(create_dummy_batches())
         self.assertEqual([actual_batch], expected_batch)
 
-    @unittest.skip('Too slow when batch size is small')
+    @unittest.skip('Too slow when batch size is small.')
     def test_should_load_and_select_real_video_in_table(self):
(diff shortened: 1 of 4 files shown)


test/integration_tests/test_select_executor.py
```diff
@@ -60,7 +60,7 @@ def test_should_load_and_select_in_table(self):
         expected_batch = list(create_dummy_batches())
         self.assertEqual([actual_batch], expected_batch)
 
-    @unittest.skip('Too slow when batch size is small')
+    @unittest.skip('Too slow when batch size is small.')
```
gaurav274, line 63: Is this slow for batch size 50? How about we set the configuration before running this test case and reset it back?

xzdandy: 30 seconds for batch size 50. We can keep batch size 50. There is no problem for that.

xzdandy: In test_pytorch, we will do this again. So I think it maybe better to disable this one for saving time.

gaurav274: Makes sense

gaurav274 (approved):

PR #102, opened 2020-09-09
Client Server support

xzdandy:
For easy performance testing and usage. 

Done: 
- Eva Server
- Command line Eva Client
- Server Response data model
- Streaming-based transmission for large-size response

TODO:  metrics 

Example Usage:
python eva.py

python eva_client.py # Below two queries will take a while.
LOAD DATA INFILE 'data/ua_detrac/ua_detrac.mp4' INTO MyVideo;
SELECT id, data FROM MyVideo WHERE id < 5;

PR #109, opened 2020-10-09
Client server

xzdandy:
Add client-server support, replace the original pr #102 

Example Usage:
Launch EVA database Server: `python eva.py`

Launch CLI: `python eva_client.py` 

Two queries below will take a while.
`LOAD DATA INFILE 'data/ua_detrac/ua_detrac.mp4' INTO MyVideo;`
`SELECT id, data FROM MyVideo WHERE id < 5;`

### DIFF
--- a/src/server/interpreter.py
+++ b/src/server/interpreter.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 from cmd import Cmd
+from src.models.server.response import Response
 
 
 class EvaCommandInterpreter(Cmd):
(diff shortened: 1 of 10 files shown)


src/server/interpreter.py
```diff
@@ -13,84 +13,62 @@
...
+            # next chunk is not avaiable yet
+            while len(self._protocol._response_chunks) <= next_chunk:
+                _ = 1
+            self._server_result += self._protocol._response_chunks[next_chunk]
+            next_chunk += 1
+
+        # print('Total length: %d, Received: %d' %
+        #      (result_length, len(self._server_result)))
+        response = Response.from_json(self._server_result)
+        print(response)
```
gaurav274, line 70: Remove this.

src/server/interpreter.py
```diff
@@ -13,84 +13,62 @@
...
+            _ = 1
+        segs = self._protocol._response_chunks[0].split('|', 1)
+        result_length = int(segs[0])
+        self._server_result = segs[1]
+        next_chunk = 1
+        while len(self._server_result) < result_length:
+            # print('Total length: %d, Received: %d' %
+            #      (result_length, len(self._server_result)), end='\r')
+            # next chunk is not avaiable yet
+            while len(self._protocol._response_chunks) <= next_chunk:
```
gaurav274, line 62: So the logic is that we spin on the pipe for the appropriate chunk and once have all the chunks, we exit.

gaurav274 (approved):

src/server/interpreter.py
xzdandy: This is for the command line client and print is the output (like those prints in do_greet and emptyline).

PR #118, opened 2020-10-25
SSD object detector

jiashenC:
* Add another UDF model SSD.
* Hacky way to rewrite pre-processing method in the model.

### DIFF
--- a/src/udfs/ssd_object_detector.py
+++ b/src/udfs/ssd_object_detector.py
@@ -74,8 +74,7 @@ class SSDObjectDetector(PytorchAbstractUDF):
             transforms.ToTensor(),
         ])
 
-    def _get_predictions(self, frames: Tensor) -> List[Outcome]:
-        print(frames.size())
+    def _get_predictions(self, frames: Tensor) -> pd.DataFrame:
         assert frames.size()[-1] == frames.size()[-2] == 300
 
         prediction = self.model(frames)
@@ -85,11 +84,21 @@ class SSDObjectDetector(PytorchAbstractUDF):
         ploc, plabel = [val.float() for val in prediction]
         encoded = encoder.decode_batch(ploc, plabel, criteria=0.5)
 
-        res = []
+        res = pd.DataFrame()
+
         for batch in encoded:
             bboxes, classes, confidences = [
                 x.detach().cpu().numpy() for x in batch]
             best = np.argwhere(confidences > self.threshold).squeeze()
+
+            # deal with empty detection
+            if len(best.shape) == 0:
+                res.append({
+                    "label": [],
+                    "pred_score": [],
+                    "pred_boxes": []
+                })
+                continue
 
             label, bbox, conf = [], [], []
             for idx in best:
@@ -100,15 +109,15 @@ class SSDObjectDetector(PytorchAbstractUDF):
                 bbox.append([x, y, w, h])
                 conf.append(confidences[idx])
 
-            res.append(
-                Outcome(pd.DataFrame(
-                    {"label": label, "pred_score": conf, "pred_boxes": bbox}
-                ), "label")
-            )
+            res.append({
+                "label": label,
+                "pred_score": conf,
+                "pred_boxes": bbox
+            })
 
         return res
 
-    def classify(self, frames: Tensor) -> List[Outcome]:
+    def classify(self, frames: Tensor) -> pd.DataFrame:
         return self._get_predictions(frames)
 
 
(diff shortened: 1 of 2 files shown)


test/integration_tests/test_pytorch.py
```diff
@@ -42,3 +42,20 @@ def test_should_run_pytorch_and_fastrcnn(self):
...
+                  OUTPUT (label TEXT(10))
+                  TYPE  Classification
+                  IMPL  'src/udfs/ssd_object_detector.py';
+        """
+        perform_query(create_udf_query)
+
+        select_query = """SELECT SSDObjectDetector(data) FROM MyVideo
+                        WHERE id < 5;"""
+        actual_batch = perform_query(select_query)
+        self.assertEqual(actual_batch.batch_size, 5)
```
gaurav274, line 61: I think we should also add a check to compare the returned tuples.

src/udfs/ssd_object_detector.py
```diff
@@ -0,0 +1,375 @@
...
+
+    @property
+    def transforms(self) -> Compose:
+        return Compose([
+            transforms.Resize([300, 300]),
+            transforms.ToTensor(),
+        ])
+
+    def _get_predictions(self, frames: Tensor) -> List[Outcome]:
+        print(frames.size())
```
gaurav274, line 78: Can get away with this print.

src/udfs/ssd_object_detector.py
```diff
@@ -0,0 +1,375 @@
...
+        return FrameInfo(-1, -1, 3, ColorSpace.RGB)
+
+    @property
+    def transforms(self) -> Compose:
+        return Compose([
+            transforms.Resize([300, 300]),
+            transforms.ToTensor(),
+        ])
+
+    def _get_predictions(self, frames: Tensor) -> List[Outcome]:
```
gaurav274, line 77: @xzdandy  I think pd.Dataframe is more generic. We should remove the Outcome all together from the repository.

src/udfs/ssd_object_detector.py
```diff
@@ -0,0 +1,375 @@
...
+
+            res.append(
+                Outcome(pd.DataFrame(
+                    {"label": label, "pred_score": conf, "pred_boxes": bbox}
+                ), "label")
+            )
+
+        return res
+
+    def classify(self, frames: Tensor) -> List[Outcome]:
```
gaurav274, line 111: pd.Dataframe should be the return type

src/udfs/ssd_object_detector.py
```diff
@@ -0,0 +1,375 @@
...
+
+        # load ssd from pytorch hub api and default to cpu
+        self.model = torch.hub.load(
+            "NVIDIA/DeepLearningExamples:torchhub",
+            "nvidia_ssd",
+            pretrained=False)
+        model_state_dict = torch.hub.load_state_dict_from_url(
+            "https://api.ngc.nvidia.com/v2/models/nvidia/ssdpyt_fp32/versions/1/files/nvidia_ssdpyt_fp32_20190225.pt",
+            map_location=torch.device("cpu"))["model"]
+        self.model.load_state_dict(model_state_dict)
```
gaurav274, line 36: Will torch cache the model and downloaded weights? We don't want these downloads to happen at each call.

src/udfs/ssd_object_detector.py
jiashenC: So it should be a list of panda frames?

src/udfs/ssd_object_detector.py
jiashenC: Yes, it will cache.

src/udfs/ssd_object_detector.py
xzdandy: def classify(self, frames: Tensor) -> pd.DataFrame:, 
panda dataframe is a table. Every row is oen result.

jiashenC: I see. Basically I was looking [here](https://github.com/georgia-tech-db/eva-reuse/blob/df6df0251361a85e6b4cd2123db8770affa76f6e/src/udfs/fastrcnn_object_detector.py#L108), but I got confused because every branch has it own format 🙂 .

xzdandy: Ah. I see. eva_reuse haven't merged the latest master branch.

jiashenC:
I think this is ready. 

The build fails due to PyTorch endpoint HTTP error. I see they have a fix PR ready but hasn't been merged.

gaurav274 (changes requested):
Updated your privileges. Can you please verify why it is still breaking the build? Apart from the HTTP error, you might also want to fix the code style error.

Try setting up the git hooks(see readme). It will auto-fix a lot of code style errors

gaurav274 (approved):

PR #124, opened 2020-12-10
Limit Implementation

Abdullahshah:
Limit Implementation Steps:

1.  Add parser functionality

- Updated `parser_visitor.py` to process the limit clause (visitor just returns the number in the statement i.e 3 as a `ConstantValueExpression`)
- Updated `select_statement.py` to include limit count
- Test case in `test/parser/test_parser.py`

2. Create a logical plan

- Created a `LogicalLimit` operator in `optimizer/operators.py`
- Call the `LogicalLimit` operator in `optimzer/statement_to_opr_convertor.py` when creating the logical plan
- Test case in `test/optimizer/test_statement_to_opr_convertor.py`

3. Create a physical plan

- Updated `optimizer/plan_generator.py` to consider Limit as a Seq Scan node type
- Created a limit physical plan in `planner/limit_plan.py`. This just receives the information of the limit count from the logical plan and constructs a physical version of it
- Updated `optimizer/generators/seq_scan_generator.py` to construct the physical `LimitPlan` after receiving a `LogicalLimit` plan
- Test cases in `test/optimizer/generators/test_seq_scan_generator.py` and `test/optimizer/test_plan_generator.py`

4. Create an executor

- Created `executor/limit_executor.py` which actually processes how the engine will limit the number of rows returned
- Update the plan executor in `executor/plan_executor.py` to call the `LimitExecutor` upon processing a `LimitPlan` created in the previous step
- Test cases in `test/executor/test_limit_executor.py`

5. Add integration tests
- Added integration tests to test end to end system for limit in `test/integration_tests/test_select_executor.py`

### DIFF
--- a/src/executor/limit_executor.py
+++ b/src/executor/limit_executor.py
@@ -1,0 +1,66 @@
+# coding=utf-8
+# Copyright 2018-2020 EVA
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Iterator
+
+from src.models.storage.batch import Batch
+from src.executor.abstract_executor import AbstractExecutor
+from src.planner.limit_plan import LimitPlan
+from src.utils.logging_manager import LoggingManager, LoggingLevel
+
+
+class LimitExecutor(AbstractExecutor):
+    """
+    Limits the number of rows returned
+
+    Arguments:
+        node (AbstractPlan): The Limit Plan
+
+    """
+
+    def __init__(self, node: LimitPlan):
+        super().__init__(node)
+        self._limit_count = node.limit_value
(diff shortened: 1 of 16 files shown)


src/executor/limit_executor.py
```diff
@@ -0,0 +1,66 @@
...
+    def validate(self):
+        pass
+
+    def exec(self) -> Iterator[Batch]:
+        child_executor = self.children[0]
+        aggregated_batch_list = []
+
+        # aggregates the batches into one large batch
+        for batch in child_executor.exec():
+            aggregated_batch_list.append(batch)
```
gaurav274, line 46: Can we break out as soon as we get the desired tuple count? That will be way faster.

gaurav274 (changes requested):

gaurav274:
@xzdandy Can you please verify the changes to the limit_executor.py? Rest LGTM. 
I changed the implementation. The limit operator shouldn't be treated as a pipeline breaker. So I return as soon as get the required tuples.

src/optimizer/operators.py
```diff
@@ -38,6 +39,7 @@ class OperatorType(IntEnum):
     LOGICALQUERYDERIVEDGET = 8,
     LOGICALUNION = 9,
     LOGICALORDERBY = 10
+    LOGICALLIMIT = 11
```
xzdandy, line 42: Could we change this enum assignment to `auto()` like in the optimizer?

xzdandy:
The rest looks good to me.

src/optimizer/operators.py
gaurav274: Sure, I will do that.

gaurav274 (approved):

PR #126, opened 2021-01-04
Change np.string_ to np.str_ in the schema_util

xzdandy:
For backward compatibility, `np.string_` continues to map to `np.bytes_` in Nmpy
https://numpy.org/doc/stable/reference/arrays.dtypes.html#string-dtype-note

Meanwhile, default decode in petastorm uses `numpy_dtype`: https://github.com/uber/petastorm/blob/6c64a37e80d41f2565866db9c2d0df4cb5a958af/petastorm/codecs.py#L256

The above two facts cause that a string object ('people') becomes a byte object (b'people') from petastorm reader.

Example: `CREATE MATERIALIZED VIEW uadtrac_fastRCNN (id, label) AS SELECT id, FastRCNNObjectDetector(data).label FROM MyVideo WHERE id < 5;` Then `SELECT id, label FROM uadtrac_fastRCNN`. The label column we get back consists of bytes objects instead of strings. 

To fix this, we can change `np.string_` to either `np.str_` or `np.unicode_`. This pr changes `np.string_` to `np.str_`.


script/install/conda_eva_environment.yml
```diff
@@ -5,7 +5,7 @@ channels:
   - defaults
   - pytorch
 dependencies:
-  - python=3.7
+  - python=3.7.6
```
xzdandy, line 8: Seems python 3.7.9 does not work with the current master code.

gaurav274 (approved):
LGTM

PR #127, opened 2021-01-08
Modularized Parser

gaurav274:
Parser modularized, moved from a single file to multiple modules


src/parser/parser_visitor/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+
+    ##################################################################
+    # STATEMENTS
+    ##################################################################
+
+    def visitDdlStatement(self, ctx: evaql_parser.DdlStatementContext):
+        ddl_statement = self.visitChildren(ctx)
+        return ddl_statement
+
+    def visitDmlStatement(self, ctx: evaql_parser.DdlStatementContext):
```
xzdandy, line 58: What's the difference between visitDdlStatement on line 54?

src/parser/parser_visitor/_expressions.py
```diff
@@ -0,0 +1,106 @@
...
+        if ctx.REAL_LITERAL() is not None:
+            return ConstantValueExpression(float(ctx.getText()))
+
+        if ctx.decimalLiteral() is not None:
+            return ConstantValueExpression(self.visit(ctx.decimalLiteral()))
+        return self.visitChildren(ctx)
+
+    def visitLogicalExpression(
+            self, ctx: evaql_parser.LogicalExpressionContext):
+        if len(ctx.children) < 3:
```
xzdandy, line 49: LOGICAL_NOT only has 2 children

src/parser/parser_visitor/_expressions.py
```diff
@@ -0,0 +1,106 @@
...
+            return ExpressionType.INVALID
+
+    def visitLogicalOperator(self, ctx: evaql_parser.LogicalOperatorContext):
+        op = ctx.getText()
+
+        if op == 'OR':
+            return ExpressionType.LOGICAL_OR
+        elif op == 'AND':
+            return ExpressionType.LOGICAL_AND
+        else:
```
xzdandy, line 95: I see we are not supporting NOT. I was thinking we have NOT supported, because https://github.com/georgia-tech-db/eva/blob/master/src/expression/logical_expression.py

src/parser/parser_visitor/__init__.py
gaurav274: duplicate function, I'll remove one.

gaurav274: Duplicate functions. I'll remove one

src/parser/parser_visitor/_expressions.py
gaurav274: We are not yet. We will add when we need be.

xzdandy:
@gaurav274 we need to merge this first.

xzdandy (approved):

PR #131, opened 2021-01-22
Add cost attribute for UDF

jiashenC:
* Add cost attribute for UDF. If the attribute is not specified, it is default to `-1` to indicate infinite cost.
* Change catalog manager to support creating UDF with cost.
* Change UDF service to support creating UDF with cost.
* Change parser, statement, operator, planner and executor to support creating UDF with cost. Modify according test so that UDF creation with default value also passes the test.
* Change parser syntax to allow optional cost attribute.
* Add test case to test UDF can be created with specified cost attribute.

### DIFF
--- a/src/parser/evaql/evaql_parser.g4
+++ b/src/parser/evaql/evaql_parser.g4
@@ -68,6 +68,7 @@ createUdf
       OUTPUT createDefinitions
       TYPE   udfType
       IMPL   udfImpl
+      COST   udfCost
     ;
 
 // details
@@ -81,6 +82,10 @@ udfType
 
 udfImpl
     : stringLiteral
+    ;
+
+udfCost
+    : decimalLiteral
     ;
 
 indexType
--- a/src/parser/evaql/evaql_parser.g4
+++ b/src/parser/evaql/evaql_parser.g4
@@ -68,6 +68,12 @@ createUdf
       OUTPUT createDefinitions
       TYPE   udfType
       IMPL   udfImpl
+    | CREATE UDF
+      ifNotExists?
+      udfName
+      INPUT  createDefinitions
+      OUTPUT createDefinitions
+      TYPE   udfType
       COST   udfCost
     ;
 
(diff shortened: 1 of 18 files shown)


src/parser/evaql/evaql_parser.g4
```diff
@@ -83,6 +84,10 @@ udfImpl
     : stringLiteral
     ;
 
+udfCost
+    : COST decimalLiteral
```
gaurav274, line 88: Should the user provide the COST? Shouldn't it be something computed by the system and maintained internally? Can you provide an example where we might require the user to provide the COST?

gaurav274 (changes requested):

src/parser/evaql/evaql_parser.g4
jiashenC: Ok, so my intuition is that if we use system profiled runtime as COST, it can depend on other factors such as input dimension. It is better to just ask user to input a relative COST? Otherwise, the monitored COST can change very frequently if input dimension changes.

gaurav274: Even if the user provides the COST, it is dependent on the input. One more concern, shouldn't the cost be dependent on the underlying system (available hardware, memory, etc.)? I can understand the cost might fluctuate based on these parameters but even in the traditional databases, the cost estimates are not always accurate. We might have to keep adjusting based on the dataset (inputs), hardware, etc. 
With the current approach, we need to provide the user with guidelines on how the cost needs to be defined. And defining such generic guidelines will get complicated. We even have to handle scenarios where the user provides erroneous costs. The cost across UDFs have to be consistent, which might be a pain if multiple users are registering the UDFs. 

This is just my take on this. We should discuss this in our weekly meeting. @jarulraj @xzdandy your suggestions?

jarulraj:
Same issue as with accuracy. We could even tag it ourselves and just fall back to a default.

PR #132, opened 2021-01-27
Remove batch add (panda append) operation as much as possible.

xzdandy:
Performance improvement.

Motivation: Panda append operation leads to data copy every time, which incurs non-negligible overhead.

Solution: 

1.  Replace with panda concat function when data merging is inevitable. 
2. Yield the results as batch generator. (This is also a support for later EVA sever streaming results (on the fly) to EVA client)


xzdandy:
The db_api branch has improved do_query implantation in [interpreter](https://github.com/georgia-tech-db/eva/blob/master/src/server/interpreter.py). Unit tests for do_query will be added when merging db_api branch.

PR #136, opened 2021-02-04
WIP: Optimizer merge

jiashenC:
As a starting point to merge the optimizer code.


src/optimizer/binder.py
```diff
@@ -0,0 +1,74 @@
...
+        yield from itertools.product(curr_iterator, *child_binders)
+
+    @staticmethod
+    def build_opr_tree_from_inorder_repr(inorder_repr: tuple) -> Operator:
+        if isinstance(inorder_repr, Operator):
+            return inorder_repr
+        else:
+            opr_tree = inorder_repr[0]
+            if len(inorder_repr) > 1:
+                # remove old children
```
jiashenC, line 65: @xzdandy I am going to work on adding unittest for binder. I am little confused here. 
inorder_repr is the inorder traversal of the tree? If it is inorder, the left child will actually appear before the root right? 
Why do we remove the old children here? To prevent infinite loop?

xzdandy: The group expression keeps the correct parent-child relationship between operators. So we need to clear the operator's original children list and use the group expression to generate the new children list.

xzdandy: I believe it should be post order. The name of the function is wrong. @gaurav274 Correct me if I am wrong.

gaurav274: Yeah, I always get confused by these order traversals. I think it should be pre-order. It is root followed by children from left to right.

test/optimizer/test_memo.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class MemoTest(unittest.TestCase):
+    def test_memo_add_with_no_forcing_id(self):
+        group_expr = MagicMock()
+        group_expr.group_id = INVALID_GROUP_ID
+        memo = Memo()
+        self.assertEqual(memo.add_group_expr(group_expr), group_expr)
+
+    def test_memo_add_with_forcing_id(self):
```
jiashenC, line 16: @xzdandy @gaurav274 
What is the expected behavior for memo if the program creates group expression with forced id when the memo is empty?
Currently, the program cannot handle this testcase due to index out of bound error.

xzdandy: good point, add an assert in b7df722

jiashenC: I see. So program should not force id if it is empty? 

I will skip some of my test cases then.

src/optimizer/optimizer_context.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def xform_opr_to_group_expr(
+        self,
+        opr: Operator,
+        root_group_id: int = INVALID_GROUP_ID,
+        is_root: bool = False,
+        copy_opr: bool = True
+    ) -> GroupExpression:
+        """
+        Generate a group expression from a logical operator.
```
jiashenC, line 52: Also related to the last discussion.

There are two cases that root id cannot be guaranteed. 
1. Group ID is just a list index, so I think `group_id <= len(group)`. Otherwise, there will be an empty group in the middle.
2. Again if memo is empty, the Group ID cannot be guaranteed.

test/optimizer/test_memo.py
xzdandy: Yes. The only time we want to force a id for a group expression is when rewrite in optimizer happens. Say after rewrite, operator **x** become operator **y**. Then the group expression containing operator **y** has to have the same id as the group expression containing operator **x** because operator  **x** can be a child of some other operators and we need to maintain the tree structure after the rewrite.

src/optimizer/optimizer_context.py
xzdandy: The time you want to guarantee a root id is when a rewrite happens. So there must be group expressions in the memo. Neither of the cases you mentioned should happen if optimizer works as intended.

src/optimizer/operators.py
```diff
@@ -76,13 +89,23 @@ def __eq__(self, other):
             is_subtree_equal = is_subtree_equal and (child1 == child2)
         return is_subtree_equal
 
+    def is_logical(self):
+        return self._opr_type < OperatorType.LOGICALDELIMITER
```
jiashenC, line 93: Why other two are not considered as logical?

src/optimizer/group.py
```diff
@@ -0,0 +1,105 @@
...
+        existing_winner = self._winner_exprs.get(property, None)
+        if not existing_winner or existing_winner.cost > cost:
+            self._winner_exprs[property] = Winner(expr, cost)
+
+    def clear_grp_exprs(self):
+        self._logical_exprs.clear()
+        self._physical_exprs.clear()
+
+    def _add_logical_expr(self, expr: 'GroupExpression'):
+        self._logical_exprs.append(expr)
```
jiashenC, line 102: Since only one logical expression is supported, I wonder do we want to put a sanity check here?

src/optimizer/binder.py
```diff
@@ -0,0 +1,74 @@
...
+        self._pattern = pattern
+        self._memo = memo
+
+    @staticmethod
+    def _grp_binder(id: int, pattern: Pattern, memo: Memo):
+        grp = memo.groups[id]
+        grp_exprs = grp.logical_exprs
+
+        for expr in grp.logical_exprs:
+            yield from Binder._binder(expr, pattern, memo)
```
jiashenC, line 37: I think I was confused here. 

To my understanding, binder returns all matches for a group expression. Am I correct?

Also what does _grp_binder do vs _binder?

src/optimizer/operators.py
xzdandy: It's a mistake when merging the master branch. Fixed.

src/optimizer/binder.py
xzdandy: Yes. You are right. Since logical_exprs only has one expression in the current implementation, so there is no difference.

src/optimizer/rules/rules.py
```diff
@@ -300,7 +300,7 @@ def _rotate(self, before):
             node_cost += cost
         children_cost = sorted(children_cost, key=lambda entry: entry[0])
         updated_children = [entry[1] for entry in children_cost]
-        before.chidren = updated_children
+        before.rewrite_children(updated_children)
```
jiashenC, line 303: @xzdandy @gaurav274 
Please check my fix here.
First, it is a typo. Second, `children` is a property so it does not get override in place.
My fix is to add a rewrite function for abstract expression, so reordering of children can happen in place.

jiashenC: https://github.com/georgia-tech-db/eva/pull/136/commits/f60fbe7e9be34834b5cd574e836b86a30f1f0cf0#diff-4579d676d8e267e19140e82e669831b8238ef971eedb783b0ef07fcd310b88d4R88

My fix for expression is here.

xzdandy: This is an experimental rule in EVA reuse work (and not enabled I believe in the mat branch). I would suggest removing it from the initial optimize merge.  In practice, you use the following way to make a property updatable.
```
@children.setter
def children(self, value):
    self._children = value
```

jiashenC: Ok, then I will just comment out the entire rule for now?

jiashenC: Actually this looks like a very useful rule to me. Can we just leave here? I will revert it back to original version but add a setter to AbstractExpression.

xzdandy: We have not tested this rule as the version in the mat branch since I do not expect it in the mat branch. That's why there exists a typo error and the property problem you found.
If we want to leave it in the pr, we probably should test it first to make sure it actually works.
And I would recommend adding the following comments to the Rule to avoid confusion:
1. The rule is not enabled.
2. The rule is not functional without pr #138.

gaurav274: Yup, comment out this rule. I'm replacing this with another more generic rule (UDF reordering).

src/optimizer/group_expression.py
```diff
@@ -0,0 +1,81 @@
...
+        self._children = new_children
+
+    def append_child(self, child_id: int):
+        self._children.append(child_id)
+
+    @property
+    def rules_explored(self):
+        return self._rules_explored
+
+    def mark_rule_explored(self, rule_id: int):
```
jiashenC, line 58: @xzdandy @gaurav274 
I am little confused why we use bitwise OR here.
Let's say rule id 1 and id 2 are explored. Then `self._rule_explored = 3`. 
Then when we check rule id 3, it will mark as if rule id 3 has been explored.

xzdandy: Actually, that's good point. I think rule id needs to be 1,2,4,8,16 to make this work.  So 
```
self._rules_explored |= ( 1<< rule_id)
self._rules_explored & (1 << rule_id)
```

jiashenC: Yeah, it has to be 2's power. Then we can only have max of 64 rules in this case. It is good for now, but will it become an issue in the future if number of rules increases?

xzdandy: Python int is not max at 64 bits. I tried `a = 1 << 1024` in python, and it works fine, though I do not know what is the max left bit shift operation Python can support.

jiashenC: Wow, in Python3, integer can actually go infinite [source](https://stackoverflow.com/questions/7604966/maximum-and-minimum-values-for-ints).

src/optimizer/binder.py
```diff
@@ -0,0 +1,74 @@
...
+    def build_opr_tree_from_inorder_repr(inorder_repr: tuple) -> Operator:
+        if isinstance(inorder_repr, Operator):
+            return inorder_repr
+        else:
+            opr_tree = inorder_repr[0]
+            if len(inorder_repr) > 1:
+                # remove old children
+                opr_tree.children.clear()
+                for child in inorder_repr[1:]:
+                    opr_tree.append_child(Binder.build_opr_tree_from_inorder_repr(child))
```
jiashenC, line 68: Additional question about binder. 
Is it true that binder will only handle opr_tree with max height of 2?
I think the code assume the rest of operators are all children of the first operator if a tuple is passed in.

src/optimizer/group.py
xzdandy: I missed this question. Group should support multiple logical expressions. It's current rewrite or memo limitation why we only support one.

src/optimizer/binder.py
xzdandy: I think the child is still a tuple. But I am also not 100% sure.

jiashenC: I think child is an operator. The pre_order_repr is a tuple of operators.

src/optimizer/optimizer_tasks.py
```diff
@@ -0,0 +1,254 @@
...
+                    self.root_expr, self.optimizer_context))
+
+        for child in self.root_expr.children:
+            child_expr = self.optimizer_context.memo.groups[child] \
+                .logical_exprs[0]
+            self.optimizer_context.task_stack.push(TopDownRewrite(
+                child_expr, self.optimizer_context))
+
+
+class BottomUpRewrite(OptimizerTask):
```
jiashenC, line 126: @xzdandy @gaurav274 
What is the necessity of BottomUpRewrite here? given we already have a top down rewrite optimization.

src/optimizer/plan_generator.py
```diff
@@ -95,18 +95,18 @@ def optimize(self, logical_plan: Operator):
     def build(self, logical_plan: Operator):
         # apply optimizations
 
-        return self.optimize(logical_plan)
+        self.optimize(logical_plan)
```
jiashenC, line 98: @xzdandy @gaurav274 
By fixing this, some of plan generator tests can pass [here](https://github.com/georgia-tech-db/eva/pull/136/commits/c99a3094d58b0dc84595e03e78038e510d97fa64). Other test cases still fail because of missing implementation rule. Hence, the winner of each group is empty. Should I introduce more implementation rules for those missing logical operators?

xzdandy: We miss limit and order by right? Do we miss any other implementation rules?

src/optimizer/optimizer_tasks.py
xzdandy: We need multiple iterations to apply all rules. But it is a good question what is the difference between applying merely top down rewrite multiple times and applying top down and bottom up alternatively.

src/optimizer/plan_generator.py
jiashenC: We also need Project, Filter, and Orderby.

xzdandy: Got. I think you can add them. They should be pretty straightforward. Or I can do that at a little late times.

jiashenC: Sure, I can work on Project and Filter first.

src/optimizer/optimizer_tasks.py
jiashenC: I see. Because those papers that I skimmed through only picked one way to apply rules, I am just curious why we need both directions here.

gaurav274: Consider the logical tree as A->B->C->D. In top-down, we start from A and keep applying rules as we go down. So, if let's say we merge B and C to C', we never go back to A and check if certain rules that weren't valid earlier(A->B) are valid now(A->C', since the subtree changed). So the idea is to quickly apply the majority of the rewrite rules. 
But in the bottom-up, we start from D and keep going up. So, if we rewrite let say B->C to C', we again try to apply rules to (C'->D). So, it is expensive but thorough.

gaurav274: @jiashenC The optimizer discussed in those papers is the part we run after doing the top-down and bottom-up rewrite. Right now, we aren't doing much there. All the cost-based interesting stuff happens there. That part at a high level is just top-down.

jiashenC: I see. That makes sense.

src/optimizer/rules/rules.py
```diff
@@ -0,0 +1,629 @@
...
+        return True
+
+    def apply(self, before: LogicalProject, context: OptimizerContext):
+        after = SeqScanPlan(None, before.target_list)
+        return after
+
+
+class LogicalFilterToSeqScan(Rule):
+    def __init__(self):
+        pattern = Pattern(OperatorType.LOGICALFILTER)
```
jiashenC, line 486: @xzdandy @gaurav274 
Can you check implementation rules here? I am not sure whether the pattern needs a Dummy, and how many? 
Can you also proofread other implementation rules that I added regarding the same issue?

xzdandy: I will take a look later today.

jiashenC: Where those mentioned integration tests?

xzdandy: https://github.com/georgia-tech-db/eva/tree/master/test/integration_tests

src/optimizer/binder.py
jiashenC: Multi-level tree can be handled with nested tuples here.

src/executor/plan_executor.py
```diff
@@ -98,27 +98,11 @@ def _clean_execution_tree(self, tree_root: AbstractExecutor):
...
-            PlanNodeType.CREATE_UDF,
-            PlanNodeType.LOAD_DATA)
-        if execution_tree.node.node_type in _INSERT_CREATE_LOAD:
-            execution_tree.exec()
-        else:
-            for batch in execution_tree.exec():
-                output_batches += batch
-
+        output = execution_tree.exec()
+        if output is not None:
```
xzdandy, line 106: We can use `yield from []` in the executors. And `yield from execution_tree.exec()` here to avoid this `output is not None` check.

src/models/storage/batch.py
```diff
@@ -16,7 +16,8 @@
 import numpy as np
 import pandas as pd
 
-from typing import List
+from typing import Iterable
+
```
xzdandy, line 20: Shall we consider removing the sorted from batch __init__ in this pr?

src/udfs/fastrcnn_object_detector.py
```diff
@@ -105,9 +105,9 @@ def _get_predictions(self, frames: Tensor) -> pd.DataFrame:
             pred_score = list(pred_score[:pred_t + 1])
             outcome = outcome.append(
                 {
-                    "label": pred_class,
-                    "pred_score": pred_score,
-                    "pred_boxes": pred_boxes
+                    "label": pred_class[0],
```
xzdandy, line 108: I think we no longer this?

src/optimizer/rules/rules.py
```diff
@@ -0,0 +1,655 @@
...
+class RuleType(IntFlag):
+    """
+    Manages enums for all the supported rules
+    """
+    # REWRITE RULES(LOGICAL -> LOGICAL)
+    EMBED_FILTER_INTO_GET = auto()
+    EMBED_PROJECT_INTO_GET = auto()
+    EMBED_FILTER_INTO_DERIVED_GET = auto()
+    EMBED_PROJECT_INTO_DERIVED_GET = auto()
+    PUSHDOWN_FILTER_THROUGH_SAMPLE = auto()
```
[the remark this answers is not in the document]
gaurav274: Yes, for the sample operator.

src/executor/plan_executor.py
gaurav274: The reason for this logic is that I thought we shouldn't force all the executors to always yield. There are cases where you want to return instead of yielding or maybe someone implementing a new executor and missed yielding in one of the conditional branches (as there can be multiple exit points in the code). How do you feel about this?

src/udfs/fastrcnn_object_detector.py
gaurav274: Yes, will correct this.

xzdandy (approved):

PR #138, opened 2021-02-11
Implemented short-circuiting for logical expressions

RodrigoDLPontes:
Implemented short-circuiting for logical expressions (AND and OR). Short circuiting is implemented as follows:

- If the left-hand side is entirely true or false (for OR and AND, respectively), the logical expression immediately returns
- If not, a "mask" is passed down to the right-hand side indicating which frames still need to be evaluated (i.e. those that evaluated to true for AND, and to false for OR)
- All expressions pass down the mask until it reaches a tuple value expression, at which point the mask is used to remove frames
- `batch.py` now implements indexing with the mask to allow the tuple value expression to do so
- Once the right-hand side of the logical expression is evaluated, the truncated results are combined with the left-hand side values to produce the final result (which is equivalent to just naively performing AND or OR with both sides)
- The logical expression is also able to receive a mask and pass it down, thus allowing for nested logical expressions (e.g. id < 10 OR id > 20 AND id < 30)
- Mask is passed around as a keyword argument (part of `**kwargs`)


src/models/storage/batch.py
```diff
@@ -191,6 +182,30 @@ def __getitem__(self, indices) -> 'Batch':
...
+        new_batch = Batch(new_frames)
+        for key in self._outcomes:
+            new_batch._outcomes[key] = [self._outcomes[key][i]
+                                        for i in required_frame_ids]
+        for key in self._temp_outcomes:
+            new_batch._temp_outcomes[key] = [self._temp_outcomes[key][i]
+                                             for i in required_frame_ids]
+        return new_batch
+
+    def _get_frames_from_mask(self, mask):
```
gaurav274, line 199: We can reuse the indices function

gaurav274 (commented):
Add a few test cases using the MagicMock library to test the logic.

src/expression/logical_expression.py
```diff
@@ -30,18 +30,25 @@ def __init__(self, exp_type: ExpressionType, left: AbstractExpression,
...
-                return Batch(pd.DataFrame(left_values | right_values))
-
+                if left_values.all().bool(): # check if all are true
+                    return Batch(left_values)
+                kwargs["mask"] = ~left_values[0]
+                right_values = self.get_child(1).evaluate(*args, **kwargs).frames
+                left_values[kwargs["mask"]] = right_values
+                return Batch(pd.DataFrame(left_values))
         else:
             values = self.get_child(0).evaluate(*args).frames
```
gaurav274, line 51: **kwargs missing

gaurav274 (changes requested):
LGTM

test/integration_tests/test_select_executor.py
```diff
@@ -162,3 +162,8 @@ def test_select_and_limit(self):
 
         self.assertEqual(actual_batch.batch_size, expected_batch[0].batch_size)
         self.assertEqual(actual_batch, expected_batch[0])
+
+    def test_logical_expression_shortcircuits(self):
+        query = "SELECT id FROM MyVideo WHERE id < 10 OR id > 20 AND id < 30;"
+        actual_batch = perform_query(query)
+        print(actual_batch)
```
gaurav274, line 169: remove the print and add some assertion

src/expression/logical_expression.py
```diff
@@ -30,18 +30,27 @@ def __init__(self, exp_type: ExpressionType, left: AbstractExpression,
...
+                    1).evaluate(*args, **kwargs).frames
+                left_values.iloc[kwargs["mask"]] = right_values
+                return Batch(pd.DataFrame(left_values))
             elif self.etype == ExpressionType.LOGICAL_OR:
-                return Batch(pd.DataFrame(left_values | right_values))
-
+                if left_values.all().bool():  # check if all are true
+                    return Batch(left_values)
+                kwargs["mask"] = left_values[~left_values[0]].index.tolist()
+                right_values = self.get_child(
```
gaurav274, line 48: Move this logic outside the if/else to reduce code duplicate

gaurav274 (changes requested):
Why is it failing on circleci?

src/models/storage/batch.py
RodrigoDLPontes: Fixed

src/expression/logical_expression.py
RodrigoDLPontes: Fixed

test/integration_tests/test_select_executor.py
RodrigoDLPontes: Removed that case entirely since I was just using it for debugging (added plenty of unit tests though)

gaurav274 (commented):
LGTM. Let's merge it after we merge the versioning freeze PR.

gaurav274 (approved):

src/expression/logical_expression.py
```diff
@@ -30,18 +30,23 @@ def __init__(self, exp_type: ExpressionType, left: AbstractExpression,
...
+                if (~left_values).all().bool():  # check if all are false
+                    return Batch(left_values)
+                kwargs["mask"] = left_values[left_values[0]].index.tolist()
             elif self.etype == ExpressionType.LOGICAL_OR:
-                return Batch(pd.DataFrame(left_values | right_values))
-
+                if left_values.all().bool():  # check if all are true
+                    return Batch(left_values)
+                kwargs["mask"] = left_values[~left_values[0]].index.tolist()
+            right_values = self.get_child(
```
xzdandy, line 44: This will still evaluate the right expression on all rows.

xzdandy: My bad. I see the tuple_value expression.

PR #139, opened 2021-02-12
Add sampling operator

karan-sarkar:


src/executor/sample_executor.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Arguments:
+        node (AbstractPlan): The Sample Plan
+
+    """
+
+    def __init__(self, node: SamplePlan):
+        super().__init__(node)
+        self._sample_freq = node.sample_value
+        self.BATCH_MAX_SIZE = ConfigurationManager().get_value(
```
xzdandy, line 35: Is this variable used?

src/executor/sample_executor.py
```diff
@@ -0,0 +1,49 @@
...
+        pass
+
+    def exec(self) -> Iterator[Batch]:
+        child_executor = self.children[0]
+        # aggregates the batches into one large batch
+        current = 0
+        for batch in child_executor.exec():
+            start = current
+            for i in range(start, len(batch), self._sample_freq):
+                yield batch[i:i + 1]
```
xzdandy, line 48: This will make the batch size become one after the sample operator.  Maybe using `df[df.index.isin([1,3])]` to select dataframe rows by a list of index is better. df is the panda dataframe object. [1,3] is the desired row indices.

test/executor/test_sample_executor.py
```diff
@@ -0,0 +1,38 @@
...
+
+        sample_executor = SampleExecutor(plan)
+        sample_executor.append_child(DummyExecutor(batches))
+        reduced_batches = list(sample_executor.exec())
+
+        total_size = 0
+        for batch in reduced_batches:
+            total_size += batch.batch_size
+
+        self.assertEqual(total_size, target_size)
```
xzdandy, line 38: Compare the content and use sample_value that does not divides the batch_size.

test/optimizer/test_statement_to_opr_convertor.py
```diff
@@ -346,6 +347,48 @@ def test_visit_select_limit(self, mock_p, mock_c, mock_d):
...
+            if expected_plan:
+                plan.append_child(expected_plan)
+            expected_plan = plan
+
+        self.assertEqual(expected_plan, actual_plan)
+
+        wrong_plan = plans[0]
+        for plan in plans[1:]:
+            wrong_plan.append_child(plan)
+        self.assertNotEqual(wrong_plan, actual_plan)
```
xzdandy, line 390: what is this wrong_plan?

src/executor/sample_executor.py
xzdandy: Actually better to use _get_frames_from_indices from the Batch class.

xzdandy: @gaurav274 we might want to change the _get_frames_from_indices's implementation check
https://stackoverflow.com/questions/19155718/select-pandas-rows-based-on-list-index

test/executor/test_sample_executor.py
karan-sarkar: Fixed.

src/executor/sample_executor.py
karan-sarkar: Fixed.

src/executor/sample_executor.py
karan-sarkar: Fixed.

test/optimizer/test_statement_to_opr_convertor.py
karan-sarkar: This is something that was in the other tests. It tests whether the plan has changed after adding commands I think.

src/executor/sample_executor.py
xzdandy: Sorry, I should point to the slicing support for Batch (which internally calls the _get_frames_from_indices). So 
```
filter = range(current, len(batch), self._sample_freq)
yield batch._get_frames_from_indices(filter)
```
can be simplified to `Batch[current::self._sample_freq]`.

test/optimizer/test_statement_to_opr_convertor.py
xzdandy: @gaurav274 What is the purpose for this wrong_plan? I am a little confused.

xzdandy:
Overall the pr LGTM

src/executor/sample_executor.py
karan-sarkar: I don't know if that makes sense because I need to keep changing the start index. I cant start at zero.

xzdandy: It's a slicing representation [start:stop:step], so Batch[current::self._sample_freq] start at current, instead of zero.

karan-sarkar: Fixed.

test/optimizer/test_statement_to_opr_convertor.py
gaurav274: Ahh, after thinking a lot, I recall adding it for one of the test cases, to double-check that the plan tree comparison is working correctly. Earlier it returned true though the plans were different. I don't think we need it for other.

src/executor/sample_executor.py
gaurav274: Here, we are yielding 1 frame at a time. I believe it makes more sense to first aggregate all the sampled frames in one batch and then yield that sampled batch.

src/executor/sample_executor.py
```diff
@@ -0,0 +1,49 @@
...
+        self._sample_freq = node.sample_value
+        self.BATCH_MAX_SIZE = ConfigurationManager().get_value(
+            "executor", "batch_size")
+
+    def validate(self):
+        pass
+
+    def exec(self) -> Iterator[Batch]:
+        child_executor = self.children[0]
+        # aggregates the batches into one large batch
```
gaurav274, line 40: Remove this comment. It's misleading.

src/parser/evaql/evaql_parser.g4
```diff
@@ -253,6 +253,7 @@ selectElement
 
 fromClause
     : FROM tableSources
+      sample=sampleClause?
```
gaurav274, line 256: Rethinking this design choice. Do you think the sampling should be done at the table level or query level? Right now we are doing it at the query level because we have a single sample value for the entire query. What if we want different sample values for different table sources? 
@xzdandy @karan-sarkar what are your thoughts on this design?

src/executor/sample_executor.py
karan-sarkar: This is outdated now. It has been updated to aggregate now.

src/parser/evaql/evaql_parser.g4
karan-sarkar: Do you mean put the sample on the load statement?

src/executor/sample_executor.py
karan-sarkar: Fixed.

src/parser/evaql/evaql_parser.g4
xzdandy: Could you give an example query in your mind?

gaurav274:
Needs to be redesigned

PR #141, opened 2021-02-13
Bring in Circle CI 

jiashenC:
* Migrate to circle ci.
* Use python-coveralls to publish coverage report.
* Installation instructions carries over user config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,5 @@
 version: 2.1
 
-orbs:
-  coveralls: coveralls/coveralls@1.0.6
-
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


xzdandy:
Do we need to force merge this? Since travis-ci does not work.

jiashenC:
> Do we need to force merge this? Since travis-ci does not work.

I don't think so. This can be actually tested with Circle CI already, but since yesterday they claimed our setting violates their agreement. I am trying to figure out why ...

jarulraj (approved):

xzdandy (approved):

PR #143, opened 2021-02-15
Added "Explode" operator

llxxll12345:
Moved everything from branch lliu374 to here.


xzdandy:
You could use `flake8 --select E,F src/ test/ --exclude src/filters,src/parser/evaql` to check the code formatting issues.

src/catalog/catalog_manager.py
```diff
@@ -284,6 +284,42 @@ def get_udf_by_name(self, name: str) -> UdfMetadata:
...
+
+        Args:
+            udf_id (int): id of the udf
+
+        Returns:
+            Inputs of the udf.
+        """
+        return self._udf_io_service.get_inputs_by_udf_id(udf_id)
+
+    def get_outputs_by_udf_id(self, udf_id: int):
```
xzdandy, line 311: Are these three functions used in the Explode operator?

src/catalog/services/udf_io_service.py
```diff
@@ -61,3 +61,16 @@ def udf_io_by_name(self, name: str):
             return self.model.query.filter(self.model._name == name).one()
         except NoResultFound:
             return None
+
+    def udf_io_by_id(self, udf_io_id: str):
```
xzdandy, line 65: Same for this function.

test/optimizer/test_statement_to_opr_convertor.py
```diff
@@ -306,6 +307,38 @@ def test_visit_select_orderby(self, mock_p, mock_c, mock_d):
...
+        plans.append(LogicalGet(TableRef(TableInfo("MyTable")), m))
+
+        expected_plan = None
+        for plan in reversed(plans):
+            if expected_plan:
+                plan.append_child(expected_plan)
+            expected_plan = plan
+
+        self.assertEqual(expected_plan, actual_plan)
+        wrong_plan = plans[0]
```
xzdandy, line 336: what is this wrong_plan?

src/executor/explode_executor.py
```diff
@@ -0,0 +1,65 @@
...
+                first_row = frames.iloc[0]
+                len_arr = []
+                for column in self._column_list:
+                    col_name = column.col_name
+                    if col_name not in frames.columns:
+                        raise KeyError("Column doesn't exist in the data frame: {}".format(column.col_name))
+                    element = first_row[col_name]
+                    if type(element) == list and len_arr == []:
+                        col_list = frames[col_name].values.tolist()
+                        len_arr = [len(r) for r in col_list]
```
xzdandy, line 54: This logic seems a little weird to me. Should we get len_arr from all exploded columns and assert they are equal?
BTW: if the exploded column has to be a list, then we should raise exception if it is not. For example, is a query like `EXPLODE SELECT id, YOLO(frame).label FROM video, id;` allowed?

xzdandy:
Let's also add an integration test.

src/catalog/catalog_manager.py
xzdandy: I would say yes. The unused functions will lead to decrease in test converge.

src/executor/explode_executor.py
```diff
@@ -0,0 +1,104 @@
...
+                        col_list = frames[col_name].values
+                        col_list = [c.flatten().tolist() for c in col_list]
+                        exploded_list.append(np.concatenate(
+                            col_list))
+                    else:
+                        exploded_list.append(
+                            np.repeat(frames[column], len_arr))
+
+                exploded_list = np.column_stack(exploded_list)
+                batch._frames = pd.DataFrame(exploded_list,
```
xzdandy, line 92: @gaurav274 Do we want to maintain the batch size after the explode operator? I am fine either way.

gaurav274: I recommend we stick with the maximum batch size.

src/executor/explode_executor.py
```diff
@@ -0,0 +1,104 @@
...
+    def exec(self) -> Iterator[Batch]:
+        for batch in self.children[0].exec():
+            if not batch.empty():
+                frames: pd.DataFrame = batch.frames
+                old_types = frames.dtypes.values
+
+                first_row = frames.iloc[0]
+                len_arr = []
+
+                for column in self._column_list:
```
gaurav274, line 57: I don't feel very confortable with inferring logic based on the first row. It is possible that the first row is NULL and then everything falls apart. All the inferring should be done either using the catalog or batch column types. Avoid inferring based on the batch element.

src/executor/explode_executor.py
```diff
@@ -0,0 +1,104 @@
...
+                    col_name = column.col_name
+
+                    if col_name not in frames.columns:
+                        raise KeyError("Column doesn't exist in "
+                                       "the data frame: {}"
+                                       .format(column.col_name))
+
+                    element = first_row[col_name]
+                    if (type(element) == list or
```
gaurav274, line 66: We should never have a list dtype in our batch. It should be either primitive datatypes or ndarray.

src/executor/explode_executor.py
llxxll12345: I will think of another kind of implementation but it may take some time.

Issue #144, opened 2021-02-22
High priority Queries involving Object detection to be supported


gaurav274:
UNNEST: #143 
JOIN: TBD
ARRAY_FUNCTIONs: TBD (https://www.postgresql.org/docs/8.4/functions-array.html)
```
 
-- GET frames with pedestrians
SELECT id, frame
FROM DETRAC
WHERE ['pedestrain'] <@ ObjDet(frame).labels;

-- GET frames with a pedestrian and a car
SELECT id, frame
FROM DETRAC
WHERE ['pedestrain', 'car'] <@ ObjDet(frame).labels;

-- GET frames with more than 5 cars
SELECT id, frame
FROM DETRAC
WHERE array_count(ObjDet(frame).labels, 'car') > 5;

-- GET frames with 2 pedestrians and 5 car
SELECT id, frame
FROM DETRAC
WHERE array_count(ObjDet(frame).labels, 'car') = 5 
       and array_count(ObjDet(frame).labels, 'pedestrian') = 2;

-- GET frames with red cars
SELECT id, frame
FROM DETRAC, UNNEST(ObjDet(frame)) as T(label, bbox) 
WHERE label = 'car' and COLOR(frame, bbox) = 'red';

-- GET frames with cars masking 50% frame area
SELECT id, frame
FROM DETRAC, UNNEST(ObjDet(frame)) as T(label, bbox) 
WHERE label = 'car' and AREA(frame, bbox) > 0.5;

-- GET bboxes of all red cars
SELECT id, frame, bboxes
FROM DETRAC, UNNEST(ObjDet(frame)) as T(label, bbox)  
WHERE label = 'car' and AREA(frame, bbox) = 'red' 
GROUPBY id;

-- GET first 100 frames with red car
SELECT id, frame
FROM DETRAC, UNNEST(ObjDet(frame)) as T(label, bbox) 
WHERE label = 'car' and COLOR(frame, bbox) = 'red'
LIMIT 100;










```

gaurav274:
#304 #316

PR #146, opened 2021-03-01
Contain operator and generic ndarray syntax support in EVA

xzdandy:
1. Add generic Ndarray syntax support, which is NDARRAY \[array_type\]\(dimensions\)
Examples:
```
CREATE UDF DummyObjectDetector
INPUT  (Frame_Array NDARRAY UINT8(3, 256, 256))
OUTPUT (label NDARRAY STR(10))
TYPE  Classification
IMPL  'test/util.py
```

array_type can be one of the following: `[INT8, UINT8, INT16, INT32, INT64, UNICODE, BOOL, FLOAT32, FLOAT64, DECIMAL, STR, DATETIME]`
**Notice**: array_type is enforced when writing with petastorm, so data needs to be at least convertible to the defined array_type.
2. Add contain comparison operator `@>`,`<@`, 
Examples:
```
SELECT id,DummyObjectDetector(data) FROM MyVideo
WHERE DummyObjectDetector(data).label <@ ['person', 'bicycle']
ORDER BY id;
```
Check issue #144 or [test/integration_tests/test_udf_executor.py](test/integration_tests/test_udf_executor.py) for more example usages

TODO:
- [x] fix broken unittest testcases.


test/parser/test_parser.py
```diff
@@ -50,7 +50,7 @@ def test_create_statement(self):
                   Frame_ID INTEGER UNIQUE,
                   Frame_Data TEXT(10),
                   Frame_Value FLOAT(1000, 201),
```
gaurav274, line 52: Why do we have these?

gaurav274: Nvm, these are for precision and scale

gaurav274 (changes requested):
Review in progress

src/catalog/schema_utils.py
```diff
@@ -78,3 +82,24 @@ def get_petastorm_schema(name, column_list):
 
         petastorm_schema = Unischema(name, petastorm_column_list)
         return petastorm_schema
+
+    @staticmethod
+    def petastorm_type_cast(schema: Unischema, df: pd.DataFrame) \
```
gaurav274, line 87: Why do we need this?

src/expression/comparison_expression.py
```diff
@@ -32,41 +33,43 @@ def __init__(self, exp_type: ExpressionType, left: AbstractExpression,
...
-        if len(left_values) != len(right_values):
-            if len(left_values) == 1:
-                left_values = pd.DataFrame(np.repeat(left_values.values,
-                                                     len(right_values),
-                                                     axis=0))
-            elif len(right_values) == 1:
-                right_values = pd.DataFrame(np.repeat(right_values.values,
-                                                      len(left_values),
-                                                      axis=0))
+        if len(lvalues) != len(rvalues):
```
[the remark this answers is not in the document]
xzdandy: What do we want to support here? I am a little lost here.

src/catalog/schema_utils.py
xzdandy: The reason is the same input array say [1,2,3,4,5] can be interpreted as different types (UINT8, INT8, INT16 and etc.), so we need to cast the array to the user defined type.

src/expression/comparison_expression.py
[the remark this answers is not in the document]
gaurav274: Code here repeats the left or right values if the length is 1. I'm saying we don't need this right? Since now we have the contains/is_contained operators we don't want to do this.

src/expression/comparison_expression.py
```diff
@@ -32,41 +33,43 @@ def __init__(self, exp_type: ExpressionType, left: AbstractExpression,
...
         elif self.etype == ExpressionType.COMPARE_LEQ:
-            return Batch(pd.DataFrame(
-                left_values.values <= right_values.values))
+            return Batch(pd.DataFrame(lvalues <= rvalues))
         elif self.etype == ExpressionType.COMPARE_NEQ:
-            return Batch(pd.DataFrame(
-                left_values.values != right_values.values))
+            return Batch(pd.DataFrame(lvalues != rvalues))
+        elif self.etype == ExpressionType.COMPARE_CONTAINS:
+            res = [[all(x in p for x in q)
```
gaurav274, line 62: Can we change this to work without for loops? Typically, it is slow if we use python loops. We should be able to perform this using numpy/pandas internal operations.

src/expression/comparison_expression.py
```diff
@@ -32,41 +33,43 @@ def __init__(self, exp_type: ExpressionType, left: AbstractExpression,
...
+                   for p, q in zip(left, right)]
+                   for left, right in zip(lvalues, rvalues)]
+            return Batch(pd.DataFrame(res))
+        elif self.etype == ExpressionType.COMPARE_IS_CONTAINED:
+            res = [[all(x in q for x in p)
```
gaurav274, line 67: Same as above.

src/catalog/schema_utils.py
gaurav274: Got it

src/catalog/models/df_column.py
```diff
@@ -30,6 +30,7 @@ class DataFrameColumn(BaseModel):
     _name = Column('name', String(100))
     _type = Column('type', Enum(ColumnType), default=Enum)
     _is_nullable = Column('is_nullable', Boolean, default=False)
+    _array_type = Column('array_type', Enum(NdArrayType), nullable=True)
```
gaurav274, line 33: We also need to add the array_type parameter to the udfIO.

src/expression/comparison_expression.py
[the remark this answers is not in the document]
xzdandy: Ah,  I got. It is different. Say a predicate UDF(data) = 'car', the UDF(data) will return a dataframe that has `batch_size` numbers of rows, while the 'car' is a single row constant dataframe. So we need to repeat the 'car' dataframe `batch_size` times.

src/expression/comparison_expression.py
xzdandy: Yeah, I was also trying to look for an existing sub array check in numpy. However I do not find it. 
What we can do is using set instead of array. But as far as my knowledge, I do not think we can avoid the for loop to check cell by cell.

src/catalog/models/df_column.py
xzdandy: Will add.

src/catalog/catalog_manager.py
```diff
@@ -234,23 +234,6 @@ def get_dataset_metadata(self, database_name: str, dataset_name: str) -> \
         metadata.schema = df_columns
         return metadata
 
-    def udf_io(
```
gaurav274, line 235: I think we should keep the catalog manager as a single source of access to the catalog models/services. It will help keep things modular.

src/expression/comparison_expression.py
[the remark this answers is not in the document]
gaurav274: Ah, now it makes sense

src/catalog/catalog_manager.py
xzdandy: Add back in 50812db

gaurav274 (approved):
LGTM

PR #154, opened 2021-06-18
Sync db api into master.

xzdandy:
- [x] Code Synchronization
- [x] Test cases update


db_api_example.py
```diff
@@ -0,0 +1,40 @@
...
+    cursor = connection.cursor()
+    for onequery in query:
+        cursor.execute(onequery)
+        response = cursor.fetch_one()
+        print('Query: %s' % onequery)
+        print(response)
+
+if __name__ == '__main__':
+    if sys.argv[1] != 'sync':
+        asyncio.run(run_async(['INVALID QUERY',
```
gaurav274, line 33: These queries are obsolete, right? The Upload statement has changed paths, right?

src/models/server/response.py
```diff
@@ -7,6 +7,7 @@
 class ResponseStatus(str, Enum):
     FAIL = -1
     SUCCESS = 0
+    MORE = 1
```
gaurav274, line 10: Can you add a comment to say why we have MORE? Thanks!

src/server/interpreter.py
```diff
@@ -1,82 +1,68 @@
-# coding=utf-8
```
gaurav274, line 1: Any reason for removing this header?

gaurav274 (changes requested):
Thanks for adding the test cases!

xzdandy:
Closing this PR. Address @gaurav274 's comments in #155

PR #155, opened 2021-07-05
Add support for UPLOAD statement

pchunduri6:
Added support for uploading a video from the client into a custom location in the server.
The syntax is as follows:
```
UPLOAD INFILE 'data/ua_detrac/ua_detrac.mp4' PATH 'test_video.mp4';
LOAD DATA INFILE 'test_video.mp4' INTO MyVideo;
SELECT id, data FROM MyVideo WHERE id < 5;
```

The prefix `/tmp` is added to the location before storing the video on the server.
Appropriate changes have been made to the parser, planner, and executor.

The UPLOAD statement is rewritten by the client as:
`UPLOAD PATH 'test_video.mp4' BLOB b'AAAA......';`

The server _only_ accepts upload command of the above syntax. The client is responsible for rewriting the command provided by the end-user into a syntax understood by the server.


src/executor/load_executor.py
```diff
@@ -42,6 +44,7 @@ def exec(self):
         # We currently use create to empty exsiting table.
         StorageEngine.create(self.node.table_metainfo)
 
-        video_reader = OpenCVReader(self.node.file_path)
+        video_reader = OpenCVReader(
+            os.path.join('/tmp', self.node.file_path))
```
xzdandy, line 48: This join '/tmp' for all load commands. I understand the reason because UPLOAD will join the '\tmp' to the user-given path, but I think this might not be a good design. One alternative can be that UPLOAD will return the generated path on the server, and the user uses the explicit path for the LOAD command?

src/executor/upload_executor.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        Upload the video blob into the location defined by 'path'
+        on the server. The video blob is in base64 format, so
+        it will first be decoded by the executor and then saved
+        at the specified path, prefixed by '/tmp'
+        """
+
+        video_blob = self.node.video_blob
+        path = self.node.file_path
+        video_bytes = base64.b64decode(video_blob[1:])
```
xzdandy, line 40: A classification question: what is video_blob[0]?

src/server/async_protocol.py
```diff
@@ -0,0 +1,151 @@
...
+            self.queue.put_nowait(message)
+
+    @asyncio.coroutine
+    def send_message(self, message):
+
+        LoggingManager().log("[ " + str(self.id) + " ]" +
+                             " Request to server: --|" + str(message) + "|--"
+                             )
+
+        if 'UPLOAD' in message:
```
xzdandy, line 138: @pchunduri6 Would you mind me moving the upload logic (line 138-147) to db_api? So we can keep the async_protocol as a generic communication protocol.

src/executor/load_executor.py
pchunduri6: We had some discussions about this. Having an end-to-end system seemed to be the priority now, so we went with this design choice. With the alternative, I feel that it would not give the user the level of control over the content that he would upload to the server. For example, the user might want to create a directory structure within the server for subsequent analytics. What do you think?

src/executor/upload_executor.py
pchunduri6: video_blob[0] is the byte literal prefix 'b'.

src/executor/load_executor.py
xzdandy: I see. I later noticed that all test cases have videos moved to /tmp, so everything works. That's good.

src/server/async_protocol.py
pchunduri6: Sure, feel free.

xzdandy:
LGTM

src/optimizer/statement_to_opr_convertor.py
```diff
@@ -225,6 +226,18 @@ def visit_load_data(self, statement: LoadDataStatement):
         load_data_opr = LogicalLoadData(table_metainfo, statement.path)
         self._plan = load_data_opr
 
+    def visit_upload(self, statement: UploadStatement):
+        """Convertor for parsed load data statement
```
gaurav274, line 230: Please, change this doc string to reflect the Upload statement. Thanks!

src/parser/upload_statement.py
```diff
@@ -0,0 +1,53 @@
...
+
+from src.parser.statement import AbstractStatement
+
+from src.parser.types import StatementType
+from pathlib import Path
+
+
+class UploadStatement(AbstractStatement):
+    """
+    Load Data Statement constructed after parsing the input query
```
gaurav274, line 24: Please modify doc string. Thanks!

src/planner/upload_plan.py
```diff
@@ -0,0 +1,46 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from src.planner.abstract_plan import AbstractPlan
+from src.planner.types import PlanOprType
+from pathlib import Path
+
+
+class UploadPlan(AbstractPlan):
+    """
+    This plan is used for storing information required for load data
```
gaurav274, line 22: Please, modify the doc string. Thanks!

gaurav274 (changes requested):

src/executor/load_executor.py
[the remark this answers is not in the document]
pchunduri6: `/tmp` is a folder in Ubuntu that hosts temporary files (that are automatically deleted after a certain time). So the name is not a deliberate choice. This design ensures that the videos are in the temporary folder for subsequent consumption by the `LOAD DATA` command. This also ensures that the uploaded videos are temporary on the server, and not taking up real server space. Loading video from any path in the server is not a priority as far as I understand. The `UPLOAD` command which always precedes the `LOAD DATA` command ensures that the video is available on the server. Let me know what you think.

src/executor/upload_executor.py
```diff
@@ -0,0 +1,42 @@
...
+        with open(os.path.join('/tmp', path), 'wb') as f:
```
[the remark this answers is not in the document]
pchunduri6: Addressed in the above response.

src/planner/upload_plan.py
pchunduri6: Fixed.

src/parser/upload_statement.py
pchunduri6: Fixed.

src/optimizer/statement_to_opr_convertor.py
pchunduri6: Fixed.

src/optimizer/rules/rules.py
```diff
@@ -404,6 +407,23 @@ def apply(self, before: LogicalLoadData, context: OptimizerContext):
         return after
 
 
+class LogicalUploadToPhysical(Rule):
+    def __init__(self):
+        pattern = Pattern(OperatorType.LOGICALUPLOAD)
+        # pattern.append_child(Pattern(OperatorType.DUMMY))
```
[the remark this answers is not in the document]
pchunduri6: Fixed.

src/optimizer/operators.py
```diff
@@ -479,3 +480,37 @@ def __eq__(self, other):
         return (is_subtree_equal
                 and self.table_metainfo == other.table_metainfo
                 and self.path == other.path)
+
+
+class LogicalUpload(Operator):
+    """Logical node for load data operation
+
+    Arguments:
+        table_metainfo(DataFrameMetadata): table to load data into
```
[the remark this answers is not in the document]
pchunduri6: Fixed.

src/executor/load_executor.py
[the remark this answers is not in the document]
gaurav274: I understand how `tmp` works :) I was of the opinion that we are persisting the video on the server. Nevertheless, in case we are going by this design, I think there is no harm in making it a config parameter. We can set it to `/tmp` by default. It will help a lot later. @xzdandy what do you think?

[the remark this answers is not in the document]
xzdandy: Yes. I agree. We can make it a configurable variable in the `eva.yml` file.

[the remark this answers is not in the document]
pchunduri6: > I understand how `tmp` works :) I was of the opinion that we are persisting the video on the server. Nevertheless, in case we are going by this design, I think there is no harm in making it a config parameter. We can set it to `/tmp` by default. It will help a lot later. @xzdandy what do you think?

I believe we can't change it to `uploaded` because of how `/tmp` works. Making it a config parameter is trivial :) I will make that change.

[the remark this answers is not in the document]
pchunduri6: The latest commit should resolve this. Please review the changes and let me know @gaurav274 @xzdandy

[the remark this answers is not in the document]
xzdandy: Good for me.

gaurav274 (approved):
LGTM

Issue #160, opened 2021-07-11
Disallow multiple queries on the same cursor to ensure cursor correctness


gaurav274:
We shouldn't allow concurrent select queries on the same cursor to ensure correctness.

`cursor.execute('query1')`
`cursor.execute('query2')`

The above code shouldn't be allowed. The user has to clear the cursor buffer before running the second query. Throw the following error. 
`InternalError: Unread result found`

xzdandy:
Will do

gaurav274:
#171

Issue #158, opened 2021-07-11
Human-readable messages from server


pchunduri6:
Right now, when the client issues a command, the server does not provide a human-readable response to the request. This affects the useability of the system. We need to add request-dependent messages to the response.

1. Connection established - Messages on server and client when a connection is established.
2. Upload - Message on the server with size and name of video being uploaded. Message on the client with the size of the video uploaded.

xzdandy:
Shall we prepare a page or doc that indicates the response we want for every command, then we can start working on them.

pchunduri6:
> Shall we prepare a page or doc that indicates the response we want for every command, then we can start working on them.

Sure, let's do that.

PR #161, opened 2021-07-11
Quickstart tutorial 

gaurav274:
- [x] Need to support Unnest to enable useful queries. #152 
- [x] Disable concurrent queries for cursor correctness. #160 @xzdandy 
- [x] Upload fails because of missing permissions. #162 @pchunduri6 
- [x] Better response messages from the server. #158

### DIFF
--- a/tutorials/object_detection.ipynb
+++ b/tutorials/object_detection.ipynb
@@ -19,7 +19,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 22,
+   "execution_count": 1,
    "id": "62e83a0b",
    "metadata": {},
    "outputs": [],
@@ -31,7 +31,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 23,
+   "execution_count": 2,
    "id": "b78a43e3",
    "metadata": {},
    "outputs": [],
@@ -43,7 +43,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 24,
+   "execution_count": 3,
    "id": "1bc99e88",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


tutorials/object_detection.ipynb
```diff
@@ -0,0 +1,269 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "id": "b18cfe2a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "cursor.execute('LOAD DATA INFILE \"test_video.mp4\" INTO MyVideo;')\n",
```
pchunduri6, line 122: Minor comment: This should be `video.mp4` to be consistent with the uploaded video.

tutorials/object_detection.ipynb
```diff
@@ -0,0 +1,269 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "b78a43e3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# import nest_asyncio\n",
+    "# nest_asyncio.apply()\n",
```
pchunduri6, line 40: Any reason why this is commented? When this is commented, I am facing the `event loop is already running` issue when I run the notebook on `ada-02`. The error goes away when I uncomment this.

tutorials/object_detection.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "    fps = vcap.get(5)\n",
+    "    fourcc = cv2.VideoWriter_fourcc(*'H264') #codec\n",
+    "    video=cv2.VideoWriter(output_path, fourcc, fps, (width,height))\n",
+    "\n",
+    "    frame_id = 0\n",
+    "    # Capture frame-by-frame\n",
+    "    ret, frame = vcap.read()  # ret = 1 if the video is captured; frame is the image\n",
+    "\n",
+    "    while ret:\n",
+    "        df = response.batch.frames\n",
```
pchunduri6, line 200: Might be better to add the variable `response` as a function argument instead of directly accessing the global variable.

tutorials/object_detection.ipynb
```diff
@@ -0,0 +1,269 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "id": "e83e5a44",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "cursor.execute(\"\"\"CREATE UDF IF NOT EXISTS FastRCNNObjectDetector \n",
```
pchunduri6, line 141: Would it be possible to elaborate on the `INPUT` and the `OUTPUT` in this query? Specifically, the purpose of the constructs `frame`, `labels`, and '`bboxes`, and how they affect the construction of the UDF. We could provide a brief explanation as a markdown cell in the notebook itself.

pchunduri6 (changes requested):
Nice work on the tutorial! I just left a few minor comments that need to be addressed, other than that it looks good!

tutorials/object_detection.ipynb
gaurav274: I was not sure if this is only showing up in my environment. Ideally, we should fix the root issue. Anyways, I'll uncomment for now.

tutorials/object_detection.ipynb
gaurav274: Yes, correct. I'll modularize it.

tutorials/object_detection.ipynb
gaurav274: Done

tutorials/object_detection.ipynb
gaurav274: Done. We should add similar information for other queries as well. Probably, will need help with that.

pchunduri6 (approved):

PR #163, opened 2021-07-12
Ndarray UDF Support (ArrayCount, Unnest, etc.)

gaurav274:
1. Add support of Ndarray UDF support to EVA - `Array_Count` and `Unnest` 
2. Bootstrapping queries added to easily load inbuilt UDFs.
3. `ANY_DIM` and `ANY` keywords added to support the declaration of arrays whose dimensions are populated at runtime.

https://georgia-tech-db.atlassian.net/wiki/spaces/EVA/overview#Detailed-Roadmap


src/udfs/ndarray_udfs/array_count.py
```diff
@@ -0,0 +1,74 @@
...
+        # if number of dimensions of searchElement not less by 1 than numpy
+        # array throw
+
+        # loop over and apply count function
+        # convert to pd data frame and then add index as a column (optional)
+
+        # inp (1 column), label: ["A"]
+        # row 0: [["A"], ["A"], ["B"]]
+        # row 1: [["A"], ["B"], ["B"]]
+        # return: [2,
```
xzdandy, line 43: Does `array_count` return one number (sum) or an array of count in each dimension?

eva.py
```diff
@@ -52,3 +51,5 @@ def eva():
 if __name__ == '__main__':
     # execute only if run as the entry point into the program
     eva()
+    mode = ConfigurationManager().get_value('core', 'mode')
+    init_builtin_udfs(mode=mode)
```
xzdandy, line 55: I think this is not working.

xzdandy (changes requested):

eva.py
gaurav274: Oh! I didn't test this. Let me try this.

src/udfs/ndarray_udfs/array_count.py
gaurav274: I'll fix the comments for better understanding. It will return a sum for each tuple. The idea is to flatten along the first dimension and then count the search element in this flattened array. 
For example, a tuple of shape (3,4,5) will be flattened into three (4,5) elements. And the search key is assumed to be of shape (4,5), else we throw an error. and this is done for all the tuples in the batch.

xzdandy: So the return result will be just one integer.

gaurav274: Yes, a single integer for each tuple.

xzdandy (approved):
LGTM

Issue #156, opened 2021-07-11
Reduce video loading time


pchunduri6:
The `LOAD DATA` command currently takes a significant amount of time to load the video into the database. Integrate optimizations from the eva-reuse project into eva to improve loading time.

xzdandy:
We have used coalesce in eva-reuse to improve the loading speed. I will merge into the master.

pchunduri6:
Resolved in #166.

PR #165, opened 2021-07-13
 Coalesce to improve writing speed. 

xzdandy:
Use coalesce in pyspark to improve writing speed (#156).
Default value 4. Can be configured in eva.yml.

Issue:
Skip test_select_and_sample in https://github.com/georgia-tech-db/eva/blob/master/test/integration_tests/test_select_executor.py.
Sample needs ORDER BY support.


xzdandy:
PR from fork can not trigger circleci

PR #169, opened 2021-07-24
Enhance server response

gaurav274:
Enhance the response message seen on the EVA client.
1. Load statement returns the number of frames loaded and the name of the video


gaurav274:
Added some return info for Load

xzdandy (approved):
LGTM

pchunduri6 (approved):
Looks good. We should also add server and client responses for the other commands. A few examples:
1. Connection established - Messages on server and client when a connection is established.
2. Upload - Message on the server with size and name of video being uploaded. Message on the client with the size of the video uploaded.

gaurav274:
Yes, Absolutely. I'll do that via another PR. Will update the issue with these comments.

PR #172, opened 2021-08-01
Load and Read speed improvement

gaurav274:
1. Enabled petastorm caching to speed up data loading.
2. Changed batch_size to batch_mem_size. Rather than using a fixed size batch during execution, Eva infers the batch size based on the video data. The idea is that the number of video frames proceeded together should be configured based on the frame size rather than using a fixed number for all videos.

### DIFF
--- a/src/planner/storage_plan.py
+++ b/src/planner/storage_plan.py
@@ -15,6 +15,7 @@
 from src.catalog.models.df_metadata import DataFrameMetadata
 from src.planner.abstract_plan import AbstractPlan
 from src.planner.types import PlanOprType
+from src.configuration.configuration_manager import ConfigurationManager
 
 
 class StoragePlan(AbstractPlan):
@@ -24,7 +25,7 @@ class StoragePlan(AbstractPlan):
 
     Arguments:
         video (DataFrameMetadata): Required meta-data for fetching data
-        batch_size (int): size of batch frame
+        batch_mem_size (int): memory size of the batch retreived
         skip_frames (int): skip frequency
         offset (int): storage offset for retrieving data
         limit (int): limit on data records to be retrieved
@@ -32,12 +33,17 @@ class StoragePlan(AbstractPlan):
         curr_shard (int): current curr_shard if data is sharded
     """
 
-    def __init__(self, video: DataFrameMetadata, batch_size: int = 1,
+    def __init__(self, video: DataFrameMetadata, batch_mem_size: int = None,
                  skip_frames: int = 0, offset: int = None, limit: int = None,
                  total_shards: int = 0, curr_shard: int = 0):
         super().__init__(PlanOprType.STORAGE_PLAN)
         self._video = video
-        self._batch_size = batch_size
+        self._batch_mem_size = batch_mem_size
+        if self._batch_mem_size is None:
+            self._batch_mem_size = ConfigurationManager(
+            ).get_value("executor", "batch_mem_size")
+            if self._batch_mem_size is None:
+                self._batch_mem_size = 300000000  # 300mb
         self._skip_frames = skip_frames
         self._offset = offset
         self._limit = limit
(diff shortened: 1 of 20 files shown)


src/planner/storage_plan.py
```diff
@@ -24,20 +25,25 @@ class StoragePlan(AbstractPlan):
...
                  total_shards: int = 0, curr_shard: int = 0):
         super().__init__(PlanOprType.STORAGE_PLAN)
         self._video = video
-        self._batch_size = batch_size
+        self._batch_mem_size = batch_mem_size
+        if self._batch_mem_size is None:
+            self._batch_mem_size = ConfigurationManager(
+            ).get_value("executor", "batch_mem_size")
+            if self._batch_mem_size is None:
+                self._batch_mem_size = 300000000  # 300mb
```
xzdandy, line 46: Any reason we need these logics, instead of a default value of 300mb, like what we did for the batch size in the old implementation.

src/readers/abstract_reader.py
```diff
@@ -29,18 +30,25 @@ class AbstractReader(metaclass=ABCMeta):
...
 
         self.file_url = file_url
-        self.batch_size = batch_size
+        self.batch_mem_size = batch_mem_size
+        if self.batch_mem_size is None or self.batch_mem_size < 0:
+            self.batch_mem_size = ConfigurationManager().get_value(
+                "executor",
+                "batch_mem_size")
+            if self.batch_mem_size is None:
+                self.batch_mem_size = 300000000  # 300mb
```
xzdandy, line 51: Same as the above comments

test/readers/test_opencv_reader.py
```diff
@@ -89,7 +89,7 @@ def test_should_start_frame_number_from_two(self):
     def test_should_start_frame_number_from_two_and_offset_from_one(self):
         video_loader = OpenCVReader(
             file_url=os.path.join(PATH_PREFIX, 'dummy.avi'),
-            batch_size=NUM_FRAMES,
+            batch_mem_size=NUM_FRAMES,
```
xzdandy, line 92: Does all these testcases make sense when we have batch_mem_size? They seem to be wield.

src/planner/storage_plan.py
gaurav274: I think it was similar

gaurav274: Ohh wait I changed 300mb to 30mb bcz it didn't pass on circleci. Is that what you are referring to?

xzdandy:
So the idea is that we only check the `batch_mem_size` when reading from a video. During the processing, we will no longer check the batch size.

src/planner/storage_plan.py
xzdandy: I meant we can do
```python
def __init__(self, video: DataFrameMetadata, batch_mem_size: int = 300000000,
                 skip_frames: int = 0, offset: int = None, limit: int = None,
                 total_shards: int = 0, curr_shard: int = 0):
```
Instead of the line 42-46.

gaurav274:
Right now yes, during processing only the storage executor decides the batch size. Every other node assumes that is the maximum batch size.

test/readers/test_opencv_reader.py
gaurav274: You are right. I can get rid of them.

src/planner/storage_plan.py
gaurav274: Ah! makes sense.

xzdandy: I feel it's better that when we create the plan in the optimizer (i.e., rules), we choose the `batch_mem_size`. This entitles optimizer more capability. How do you think?

gaurav274: Interesting. Let me think about it.

gaurav274: Done

test/readers/test_opencv_reader.py
gaurav274: Redesigned the test cases to use batch_mem_size.

xzdandy (approved):

PR #174, opened 2021-08-05
added steps for gpu config in readme

Anirudh58:
Improved readme


README.md
```diff
@@ -61,13 +61,18 @@ Please refer to [official doc](https://docs.docker.com/engine/install/).
...
 ```
-3. Below is a basic query that should work on the client
+
+3. Run the `LOAD` command in the client terminal: (may take a while)
+```mysql
+LOAD DATA INFILE 'test_video.mp4' INTO MyVideo;
+```
+4. Below is a basic query that should work on the client
 ```mysql
 SELECT id, data FROM MyVideo WHERE id < 5;
```
gaurav274, line 77: Personally, I feel this query doesn't look good on the readme. It doesn't accomplish much. How about we replace it with the select query in tutorial? At least that has object detection. What do you think?

gaurav274 (commented):
LGTM. Why isn't the testsuite getting triggered?

README.md
Anirudh58: But when a person is starting off, they need to know how to load a video right? for any select query to even work.

gaurav274: Ah! I missed it is under the client testing. Then, it is fine.

gaurav274 (approved):

PR #181, opened 2021-09-30
Guide for creating and executing UDFs

Anirudh58:


api-docs/guide/adding-a-custom-udf.rst
```diff
@@ -0,0 +1,88 @@
...
+    1. Create a new file under `src/udfs/` and give it a descriptive name. eg: `fastrcnn_object_detector.py`, `midas_depth_estimator.py`. 
+    
+    2. Here, you need to create and implement a class that derives from `PytorchAbstractUDF`.  
+
+        * The `PytorchAbstractUDF` is a parent class that defines and implements standard methods for model inference.
+
+        * `_get_predictions()` - an abstract method that needs to be implemented in your child class.
+
+        * `classify()` - A  method that receives the frames and calls the `_get_predictions()` implemented in your child class. Based on GPU batch size, it also decides whether to split frames into chunks and performs the accumulation.
+
```
gaurav274, line 26: We should also provide a few example code snippets so that readers can follow. Add examples from already existing one.

api-docs/guide/adding-a-custom-udf.rst
```diff
@@ -0,0 +1,88 @@
...
+Introduction to UDFs
+=====================
+
+    * UDFs are user-defined functions, used to achieve a specific functionality on any video loaded to EVA. 
+    * As the name suggests, these functions are defined and implemented by users. Once the UDFs are registered, you can 'call' these functions in your queries. 
+
+Implementing your own UDF
+=========================
+
+    1. Create a new file under `src/udfs/` and give it a descriptive name. eg: `fastrcnn_object_detector.py`, `midas_depth_estimator.py`.
```
gaurav274, line 17: Do you think we should have these udfs outside the src/ directory? We can create a /udf directory to avoid readers from getting confused.

api-docs/guide/adding-a-custom-udf.rst
```diff
@@ -0,0 +1,88 @@
...
+    2. Now, we can register the UDF with the following syntax:
+
+        `CREATE UDF [ IF NOT EXISTS ] <name> 
+        INPUT  ( [ <arg_name> <arg_data_type> ] [ , ... ] )
+        OUTPUT ( [ <result_name> <result_data_type> ] [ , ... ] )
+        TYPE  <udf_type_name>
+        IMPL  '<path_to_implementation>'`
+
+        where,
+
```
gaurav274, line 70: Give an example

api-docs/guide/adding-a-custom-udf.rst
Anirudh58: Got it. I'll do that.

api-docs/guide/adding-a-custom-udf.rst
Anirudh58: I think this looks fine as it is? All core code components going into src/ makes sense for me. Prof. @jarulraj @suryatejreddy can comment.

api-docs/guide/adding-a-custom-udf.rst
Anirudh58: Sure.

api-docs/guide/adding-a-custom-udf.rst
suryatejreddy: I think it should be outside `src/`. Just because we want to show that the UDF is not part of the EVA Codebase and is something external that users can create. However, if we want to package some UDF's as part of the EVA module then they should be inside src, so depending on that, we should take the call I guess. @gaurav274 @Anirudh58 @jarulraj

gaurav274: Yes, I agree with Surya. Users should add udfs outside the `src/ ` directory.

Anirudh58: Makes sense. I'll change accordingly.

gaurav274 (approved):

gaurav274:
@Anirudh58 Please merge the master so that the build passes. Thanks!

Anirudh58:
> @Anirudh58 Please merge the master so that the build passes. Thanks!

Done! I think you can merge it.

PR #197, opened 2021-11-19
Ocr code

devshreebharatia:
The code implementation for optical Character Recognition UDF.
A unit test has also been added in test_pytorch.py to test the OCR UDF.

### DIFF
--- a/src/udfs/gpu_compatible.py
+++ b/src/udfs/gpu_compatible.py
@@ -24,10 +24,14 @@ class GPUCompatible(ABC):
     def device(self,):
         return self._device
 
-    def assign_device(self, device: str):
+    def assign_device(self, device: str) -> object:
         '''
         Assigning the device passed by the UDF to the member variable of the class
-        Returns after enabling GPU for the device
+        This is an internal function used by eva for device allocation.
+        Arguments:
+            device (str): device details
+        Returns:
+            A GPU compatible object
         '''
         self._device = device
         return self.to_device(device)
--- a/src/udfs/gpu_compatible.py
+++ b/src/udfs/gpu_compatible.py
@@ -13,9 +13,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from abc import abstractmethod, ABC
-
+from src.constants import NO_GPU
 
 class GPUCompatible(ABC):
+
+    def __init__(self):
+        self._device = NO_GPU
+
+    @property
+    def device(self,):
+        return self._device
+
+    def assign_device(self, device: str):
+        '''
+        Assigning the device passed by the UDF to the member variable of the class
+        Returns after enabling GPU for the device
+        '''
+        self._device = device
+        return self.to_device(device)
+
     @abstractmethod
     def to_device(self, device: str):
         """
--- a/src/udfs/optical_character_recognition.py
+++ b/src/udfs/optical_character_recognition.py
@@ -67,49 +67,53 @@ class OpticalCharacterRecognition(AbstractClassifierUDF, GPUCompatible):
         
         frames_list = frames.values.tolist()
         frames = np.array(frames_list)
-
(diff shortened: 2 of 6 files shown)


src/udfs/gpu_compatible.py
```diff
@@ -13,9 +13,25 @@
...
+    def __init__(self):
+        self._device = NO_GPU
+
+    @property
+    def device(self,):
+        return self._device
+
+    def assign_device(self, device: str):
+        '''
+        Assigning the device passed by the UDF to the member variable of the class
```
gaurav274, line 29: We can be made clearer. Add comment to say it is an internal function. Thanks!

src/udfs/gpu_compatible.py
```diff
@@ -13,9 +13,25 @@
...
+        Returns after enabling GPU for the device
+        '''
+        self._device = device
+        return self.to_device(device)
+
     @abstractmethod
```
gaurav274, line 35: Please also add a doc string for this function? Since it has to be implemented by the user, we should explicitly state the input and output requirements.

src/udfs/optical_character_recognition.py
```diff
@@ -0,0 +1,115 @@
...
+                    outcome = outcome.append(
+                        {
+                            "labels": list(pred_class),
+                            "scores": list(pred_score),
+                            "boxes": list(pred_boxes)
+                        },
+                        ignore_index=True)
+
+
+        return outcome
```
gaurav274, line 115: Missing new line

src/udfs/optical_character_recognition.py
```diff
@@ -0,0 +1,115 @@
...
+
+        """
+        
+        frames_list = frames.values.tolist()
+        frames = np.array(frames_list)
+
+        outcome = pd.DataFrame()
+        
+        for i in range(frames.shape[0]):
+            prediction = self.model.readtext_batched(frames[i])
```
gaurav274, line 74: Isn't this logic processing one frame at a time? How about we call `self.model.readtext_batched(frames)`?

src/udfs/pytorch_abstract_udf.py
```diff
@@ -38,8 +38,8 @@ def __init__(self):
         AbstractClassifierUDF.__init__(self)
         nn.Module.__init__(self)
 
-    def get_device(self):
-        return next(self.parameters()).device
+    #def get_device(self):
```
gaurav274, line 41: Are these changes required? I remember we decided to revert these back. Please double-check, thanks!

src/udfs/pytorch_abstract_udf.py
```diff
@@ -52,7 +52,7 @@ def transform(self, images: np.ndarray):
 
     def forward(self, frames: List[np.ndarray]):
         tens_batch = torch.cat([self.transform(x) for x in frames])\
-            .to(self.get_device())
+            .to(f"cuda:{self.device}")
```
gaurav274, line 55: Similar as above.

gaurav274 (changes requested):

src/udfs/pytorch_abstract_udf.py
devshreebharatia: Yes, I think these changes are required. We are storing the device object using the assign_device() function in the GPU Compatible class as one of its variables. So we can directly access that object instead of using a get_device() function. I believe that was the reason we decided to eliminate this functionality. What do you think?

src/udfs/optical_character_recognition.py
devshreebharatia: I think there was some confusion due to the image size. I have made a few changes to make prediction on a batch of frames. Once we have the results for the entire batch, we traverse through each of the results and append them to the output dataframe. Creating a new PR for this!

jarulraj:
Resolved in #303

PR #223, opened 2022-02-04
Implement test APPLY_BLUR command

LordDarkula:
`APPLY_BLUR AS SELECT ...`

Apply blur to all frames returned by a SELECT command.


jarulraj:
We should add this as a UDF instead of an operator?

PR #224, opened 2022-02-04
Support for loading datasets

Anirudh58:
Major changes:

1. Support to load CSVs into tables. Split load_executors as load_video_executors and load_csv_executors
2. Added support to specify FORMAT in query
3. Added support to specify columns in LOAD query. 
4. Added/modified all the relevant testcases for this. 
5. Added helper scripts to load datasets.

We can now execute queries like:

`LOAD DATA INFILE 'data/meta.csv' INTO  MyMeta (id, frame_id, video_id, label) WITH FORMAT CSV;`

### DIFF
--- a/eva/executor/create_executor.py
+++ b/eva/executor/create_executor.py
@@ -34,11 +34,28 @@ class CreateExecutor(AbstractExecutor):
         Calls the catalog to create metadata corresponding to the table.
         Calls the storage to create a spark dataframe from the metadata object.
         """
-        if (self.node.if_not_exists):
-            # check catalog if we already have this table
-            return
 
-        table_name = self.node.video_ref.table_info.table_name
+        if (self.node.if_not_exists):
+            
+            # check if the table exists
+            catalog_manager = CatalogManager()
+
+            # TODO: Due to the way get_table_bindings is written, 
+            # an error will be logged here if the table does not exist. 
+            # But that is okay since we are just checking if the table exists. 
+            table_name = self.node.video_ref.table_name
+            metadata_id, _ = catalog_manager.get_table_bindings("eva_catalog",
+                                                                table_name)
+
+            # metadata_id is not None when the table exists, so return
+            if metadata_id is not None:
+                return
+
+        # TODO: I had to comment out the following line,
+        # because it was causing an error. table_name is in video_ref, 
+        # and not in table_info.
+        # table_name = self.node.video_ref.table_info.table_name
+        table_name = self.node.video_ref.table_name
         file_url = str(generate_file_path(table_name))
         metadata = CatalogManager().create_metadata(table_name,
                                                     file_url,
--- a/eva/executor/load_csv_executor.py
+++ b/eva/executor/load_csv_executor.py
@@ -1,0 +1,66 @@
+# coding=utf-8
+# Copyright 2018-2020 EVA
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+import pandas as pd
+
+from eva.configuration.configuration_manager import ConfigurationManager
+from eva.executor.abstract_executor import AbstractExecutor
+from eva.models.storage.batch import Batch
+from eva.planner.load_data_plan import LoadDataPlan
+from eva.storage.storage_engine import StorageEngine
+from eva.readers.csv_reader import CSVReader
+
+
+class LoadCSVExecutor(AbstractExecutor):
+
+    def __init__(self, node: LoadDataPlan):
+        super().__init__(node)
+        config = ConfigurationManager()
+        self.path_prefix = config.get_value('storage', 'path_prefix')
+
+    def validate(self):
+        pass
+
+    def exec(self):
+        """
+        Read the input meta file using pandas and persist data
+        using storage engine
+        """
+
+        # Read the CSV file
+        # converters is a dictionary of functions that convert the values
+        # in the column to the desired type
+        csv_reader = CSVReader(
+            os.path.join(self.path_prefix, self.node.file_path),
+            column_list=self.node.column_list,
+            batch_mem_size=self.node.batch_mem_size
+        )
+
+        # write with storage engine in batches
+        num_loaded_frames = 0
+        for batch in csv_reader.read():
+            StorageEngine.write(self.node.table_metainfo, batch)
+            num_loaded_frames += len(batch)
+
+        # yield result
+        df_yield_result = Batch(
+            pd.DataFrame({
+                'CSV': str(self.node.file_path),
+                'Number of loaded frames': num_loaded_frames
+            }, index=[0]))
+
+        yield df_yield_result
--- a/eva/executor/load_executor.py
+++ b/eva/executor/load_executor.py
@@ -13,15 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-import pandas as pd
-
-from eva.planner.load_data_plan import LoadDataPlan
-from eva.executor.abstract_executor import AbstractExecutor
-from eva.storage.storage_engine import StorageEngine
-from eva.readers.opencv_reader import OpenCVReader
-from eva.models.storage.batch import Batch
 from eva.configuration.configuration_manager import ConfigurationManager
+from eva.executor.abstract_executor import AbstractExecutor
+from eva.executor.load_csv_executor import LoadCSVExecutor
+from eva.executor.load_video_executor import LoadVideoExecutor
+from eva.planner.load_data_plan import LoadDataPlan
+from eva.parser.types import FileFormatType
 
 
 class LoadDataExecutor(AbstractExecutor):
@@ -36,26 +33,17 @@ class LoadDataExecutor(AbstractExecutor):
 
     def exec(self):
         """
-        Read the input video using opencv and persist data
-        using storage engine
+        Use TYPE to determine the type of data to load.
         """
 
-        # videos are persisted using (id, data) schema where id = frame_id
-        # and data = frame_data. Current logic supports loading a video into
-        # storage with the assumption that frame_id starts from 0. In case
-        # we want to append to the existing store we have to figure out the
-        # correct frame_id. It can also be a parameter based by the user.
+        # invoke the appropriate executor
+        if self.node.file_options['file_format'] == FileFormatType.VIDEO:
+            executor = LoadVideoExecutor(self.node)
+        elif self.node.file_options['file_format'] == FileFormatType.CSV:
+            executor = LoadCSVExecutor(self.node)
 
-        # We currently use create to empty existing table.
-        StorageEngine.create(self.node.table_metainfo)
-        num_loaded_frames = 0
-        video_reader = OpenCVReader(
-            os.path.join(self.path_prefix, self.node.file_path),
-            batch_mem_size=self.node.batch_mem_size)
-        for batch in video_reader.read():
-            StorageEngine.write(self.node.table_metainfo, batch)
-            num_loaded_frames += len(batch)
+        # for each batch, exec the executor
+        for batch in executor.exec():
+            yield batch
 
-        yield Batch(pd.DataFrame({'Video': str(self.node.file_path),
-                                  'Num Loaded Frames': num_loaded_frames},
-                                 index=[0]))
+
--- a/eva/optimizer/operators.py
+++ b/eva/optimizer/operators.py
@@ -456,10 +456,14 @@ class LogicalLoadData(Operator):
     """
 
     def __init__(self, table_metainfo: DataFrameMetadata,
-                 path: Path, children=None):
+                 path: Path, 
+                 column_list: List[AbstractExpression] = None,
+                 file_options: dict = None, children=None):
         super().__init__(OperatorType.LOGICALLOADDATA, children=children)
         self._table_metainfo = table_metainfo
         self._path = path
+        self._column_list = column_list
+        self._file_options = file_options
 
     @property
     def table_metainfo(self):
@@ -469,9 +473,21 @@ class LogicalLoadData(Operator):
     def path(self):
         return self._path
 
+    @property
+    def column_list(self):
+        return self._column_list
+
+    @property
+    def file_options(self):
+        return self._file_options
+
     def __str__(self):
-        return 'LogicalLoadData(table: {}, path: {})'.format(
-            self.table_metainfo, self.path)
+        return "LogicalLoadData(table: {}, path: {}, \
+                column_list: {}, \
+                file_options: {})".format(self.table_metainfo, 
+                                          self.path, 
+                                          self.column_list, 
+                                          self.file_options)
 
     def __eq__(self, other):
         is_subtree_equal = super().__eq__(other)
@@ -479,7 +495,9 @@ class LogicalLoadData(Operator):
             return False
         return (is_subtree_equal
                 and self.table_metainfo == other.table_metainfo
-                and self.path == other.path)
+                and self.path == other.path
+                and self.column_list == other.column_list
+                and self.file_options == other.file_options)
 
 
 class LogicalUpload(Operator):
--- a/eva/optimizer/statement_to_opr_convertor.py
+++ b/eva/optimizer/statement_to_opr_convertor.py
@@ -35,6 +35,7 @@ from eva.optimizer.optimizer_utils import (bind_table_ref, bind_columns_expr,
                                            create_video_metadata)
 from eva.parser.table_ref import TableRef
 from eva.utils.logging_manager import LoggingLevel, LoggingManager
+from eva.expression.tuple_value_expression import TupleValueExpression
 
 
 class StatementToPlanConvertor:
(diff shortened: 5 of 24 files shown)


eva/executor/create_executor.py
```diff
@@ -34,11 +34,20 @@ def exec(self):
...
         Calls the storage to create a spark dataframe from the metadata object.
         """
+
+        print(f"CreateExecutor: inside exec")
+        print(f"CreateExecutor: dir(node) = {dir(self.node)}")
+        print(f"CreateExecutor: node.if_not_exists = {self.node.if_not_exists}")
+
+        # TODO: Disabling the below code for now. Check what actually needs to be done here.
+        '''
         if (self.node.if_not_exists):
```
Anirudh58, line 44: I had to disable this code to get the "CREATE TABLE" query working. Need to check what exactly has to be done here

eva/executor/create_executor.py
```diff
@@ -34,11 +34,20 @@ def exec(self):
         Calls the catalog to create metadata corresponding to the table.
         Calls the storage to create a spark dataframe from the metadata object.
         """
+
+        print(f"CreateExecutor: inside exec")
+        print(f"CreateExecutor: dir(node) = {dir(self.node)}")
+        print(f"CreateExecutor: node.if_not_exists = {self.node.if_not_exists}")
+
+        # TODO: Disabling the below code for now. Check what actually needs to be done here.
+        '''
```
Anirudh58, line 43: I had to disable this code segment to get the "CREATE TABLE" query working. Need to check what exactly has to be done here

eva/executor/create_executor.py
```diff
@@ -34,11 +34,20 @@ def exec(self):
...
         if (self.node.if_not_exists):
             # check catalog if we already have this table
             return
+        '''
 
-        table_name = self.node.video_ref.table_info.table_name
+        #table_name = self.node.video_ref.table_info.table_name
+        table_name = self.node.video_ref.table_name
```
Anirudh58, line 50: Minor discrepancies in the way we are loading videos/csv files

eva/executor/load_executor.py
```diff
@@ -36,26 +34,24 @@ def validate(self):
...
     def exec(self):
         """
-        Read the input video using opencv and persist data
-        using storage engine
+        Use TYPE to determine the type of data to load. 
         """
+        
+        print(f"AbstractLoadDataExecutor: inside exec")
+
+        # TODO: Currently using file extension to decide which executor to invoke. Figure out how to get TYPE from the query
```
Anirudh58, line 42: I was not able to figure out how exactly a query is getting converted to a node plan, so I was not able to incorporate the "TYPE" keyword :/ I will need some more guidance here. Temporarily using file extension and invoking the corresponding executors.

eva/executor/load_meta_executor.py
```diff
@@ -0,0 +1,76 @@
...
+        meta_df = pd.read_csv(csv_file_path, converters = {"bbox" : convert_bbox})
+        meta_df_len = len(meta_df)
+        print(meta_df)
+        print(f"LoadDataExecutor: read {meta_df_len} rows from CSV file")
+
+        # Create a batch
+        batch = Batch(meta_df)
+
+        # Store the batch
+        # TODO: Get the column types from the metadata
```
Anirudh58, line 65: Need to do this change

eva/readers/abstract_reader.py
```diff
@@ -52,7 +52,10 @@ def read(self) -> Iterator[Batch]:
         row_size = None
         for data in self._read():
             if row_size is None:
-                row_size = data['data'].nbytes
+                if 'data' in data:
```
Anirudh58, line 55: We are assuming that we always have an attribute 'data' in the result of SELECT queries. But this may not be the case. For example: SELECT queries on meta table. I had to add this change, to avoid errors.

Anirudh58: This is fixed in @gaurav274 's latest PR

eva/executor/create_executor.py
Anirudh58: Fixed

eva/executor/load_csv_executor.py
```diff
@@ -0,0 +1,76 @@
...
+from eva.storage.storage_engine import StorageEngine
+from eva.readers.csv_reader import CSVReader
+
+
+class LoadCSVExecutor(AbstractExecutor):
+
+    def __init__(self, node: LoadDataPlan):
+        super().__init__(node)
+        config = ConfigurationManager()
+        self.path_prefix = config.get_value('storage', 'path_prefix')
```
xzdandy, line 34: I suggest doing the path join in the operator node instead of the executor. The executor can simply execute based on the plan.

eva/executor/load_csv_executor.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Read the CSV file
+        # converters is a dictionary of functions that convert the values
+        # in the column to the desired type
+        # column_list is a list of columns to be read from the CSV file. 
+        # This should match the table schema
+        csv_reader = CSVReader(
+            os.path.join(self.path_prefix, self.node.file_path),
+            converters={"bbox": convert_bbox},
+            column_list=['id', 'frame_id', 'video_id', 'dataset_name',
```
xzdandy, line 58: Why the column_list is hard coded?

eva/executor/load_csv_executor.py
```diff
@@ -0,0 +1,76 @@
...
+            return np.array([np.float32(coord) for coord in bbox.split(",")])
+
+        # Read the CSV file
+        # converters is a dictionary of functions that convert the values
+        # in the column to the desired type
+        # column_list is a list of columns to be read from the CSV file. 
+        # This should match the table schema
+        csv_reader = CSVReader(
+            os.path.join(self.path_prefix, self.node.file_path),
+            converters={"bbox": convert_bbox},
```
xzdandy, line 57: It's better if converters can be configurable instead of hard coded.

eva/optimizer/operators.py
```diff
@@ -456,10 +457,13 @@ class LogicalLoadData(Operator):
...
 
     def __init__(self, table_metainfo: DataFrameMetadata,
-                 path: Path, children=None):
+                 path: Path, file_format: FileFormatType, 
+                 file_options: dict, children=None):
         super().__init__(OperatorType.LOGICALLOADDATA, children=children)
         self._table_metainfo = table_metainfo
         self._path = path
+        self._file_format = file_format
+        self._file_options = file_options
```
xzdandy, line 466: What is `file_option`? And how it is used? I don't see it in the executor. Is that resevered for future use? If so, what is the example use case?

eva/parser/evaql/evaql_parser.g4
```diff
@@ -154,10 +154,16 @@ updateStatement
...
+
 loadStatement
     : LOAD DATA
       INFILE fileName
       INTO tableName
+      (WITH fileOptions)?
+    ;
+
+fileOptions
+    : FORMAT fileFormat=(CSV|VIDEO)
```
xzdandy, line 166: It seems that the design is that file format is one possible file options. Then we may only need file options across node, plan, and executor instead of file format and file options.

eva/parser/parser_visitor/_load_statement.py
```diff
@@ -18,11 +18,28 @@
...
 
 class Load(evaql_parserVisitor):
     def visitLoadStatement(self, ctx: evaql_parser.LoadStatementContext):
         file_path = self.visit(ctx.fileName()).value
         table = TableRef(self.visit(ctx.tableName()))
-        stmt = LoadDataStatement(table, file_path)
+
+        # Set defualt as Video
+        file_format = FileFormatType.VIDEO
+        other_options = None
```
xzdandy, line 31: I will suggest just having a file_options dictionary with file_format as a key. We may have many possible options in future (e.g., the convert can be option). This saves us from changing the data structure of node and plan everytime when we add one option.

eva/readers/abstract_reader.py
```diff
@@ -52,7 +52,10 @@ def read(self) -> Iterator[Batch]:
...
+                    row_size = data['data'].nbytes
```
xzdandy, line 56: Can we use `from sys import getsizeof` to estimate? So it works for general tables.

xzdandy (changes requested):
We also need integration test cases (for CSV).

eva/optimizer/operators.py
gaurav274: Yes, the design assumes file_format is one example key in file options. I followed the design used by Postgres. Other examples can be DELIMITER 'delimiter_character', HEADER [ boolean ] etc.

gaurav274: @Anirudh58 we should consider only passing the file_options as a dictionary. Thanks!

eva/readers/abstract_reader.py
gaurav274: Not sure if it provides the size of nested objects. I remember testing something similar but it didn't give the correct size. We need to check if there is a better way to do it.

eva/executor/load_meta_executor.py
```diff
@@ -0,0 +1,76 @@
...
+import pandas as pd
+
+from eva.configuration.configuration_manager import ConfigurationManager
+from eva.executor.abstract_executor import AbstractExecutor
+from eva.models.storage.batch import Batch
+from eva.planner.load_data_plan import LoadDataPlan
+from eva.storage.storage_engine import StorageEngine
+
+
+class LoadMetaExecutor(AbstractExecutor):
```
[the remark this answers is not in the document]
Anirudh58: Oh yes there is a new LoadCSVExecutor. Forgot to delete this file on remote.

eva/executor/load_csv_executor.py
Anirudh58: Hmm in the original code, we were joining the path from executor only right? as below. 

```
        # read the video with opencv
        video_reader = OpenCVReader(
            os.path.join(self.path_prefix, self.node.file_path),
            batch_mem_size=self.node.batch_mem_size)
```

Anirudh58: Do you want me to change this code in the video_executor also?

eva/executor/load_csv_executor.py
Anirudh58: I gave this some thought too. When you say 'configurable', you want it to be configured through the query itself? Can you give an example? 

The reason why I had to add this converter was because, by default, when pandas reads a CSV file, it inputs columns values as string values.

eva/executor/load_csv_executor.py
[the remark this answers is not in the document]
Anirudh58: So the new load query would be something like: 

```
LOAD DATA INFILE 'test.csv' INTO meta_bdd WITH FORMAT CSV COLUMNS ('id', 'frame_id', 'video_id', 'dataset_name', 'label', 'bbox', 'object_id')
```

Anirudh58: For loading videos, we know the columns are always 'data' and 'id'. But for csv files, it can be anything right? How else can I handle this?

eva/executor/load_executor.py
```diff
@@ -36,26 +33,17 @@ def validate(self):
...
-
-        yield Batch(pd.DataFrame({'Video': str(self.node.file_path),
-                                  'Num Loaded Frames': num_loaded_frames},
-                                 index=[0]))
+        # invoke the appropriate executor
+        if self.node.file_format == FileFormatType.VIDEO:
+            executor = LoadVideoExecutor(self.node)
+        elif self.node.file_format == FileFormatType.CSV:
+            executor = LoadCSVExecutor(self.node)
+
```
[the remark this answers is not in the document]
Anirudh58: Yes right makes sense, I will make the change.

eva/optimizer/operators.py
Anirudh58: Ok got it so file_format should now be a key in file_options

eva/parser/evaql/evaql_parser.g4
Anirudh58: Got it got it. I will make this change thanks

eva/parser/parser_visitor/_load_statement.py
Anirudh58: Got it.

eva/readers/abstract_reader.py
Anirudh58: ya get sizeof does not seem to return correct values for nested dictionaries.

eva/executor/load_csv_executor.py
xzdandy: I suggest doing here https://github.com/georgia-tech-db/eva/blob/c2eb49c522f6ab9086a69c9ac2f6bf3747f64ea7/eva/optimizer/statement_to_opr_convertor.py#L226, though we can discuss.

eva/executor/load_csv_executor.py
xzdandy: I got the problems. Can we get the columns from the csv file? I assume that we need the column_list because we don't want always load all columns. If that's the case, we can have column_list as a file option to the query.

eva/executor/load_csv_executor.py
xzdandy: Yeah, the idea is that we can have converter as something like UDF, and in query when we specify the columns list, we can also specify the converter if needed.  

Like if the format you pick for columns is LOAD DATA INFILE 'test.csv' INTO meta_bdd WITH FORMAT CSV COLUMNS ('id', 'frame_id', 'video_id', 'dataset_name', 'label', 'bbox', 'object_id'). Then we can have LOAD DATA INFILE 'test.csv' INTO meta_bdd WITH FORMAT CSV COLUMNS ('id', 'frame_id', 'video_id', 'dataset_name', 'label', 'bbox': 'bbox_converter', 'object_id')

eva/executor/load_executor.py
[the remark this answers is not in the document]
Anirudh58: Function in the optimizer to make changes to: 

https://github.com/georgia-tech-db/eva/blob/c2eb49c522f6ab9086a69c9ac2f6bf3747f64ea7/eva/optimizer/rules/rules.py#L406

eva/executor/load_csv_executor.py
Anirudh58: Oh so bbox_converter should be a registered UDF then?

xzdandy: uhm. This is good point. UDF actually inputs dataframe and outputs dataframe. I believe converter inputs string and outputs numpy array. Now i think about it, the better way to do that is have this convert as a built-in function like what you did. And have a syntax sugar on the query to indicate which column needs to be converted (so it doesn't work only on the bbox column.). For example  LOAD DATA INFILE 'test.csv' INTO meta_bdd WITH FORMAT CSV COLUMNS ('id', 'frame_id', 'video_id', 'dataset_name', 'label', NDARRAY('bbox'), 'object_id'). @gaurav274 What is your opinion?

eva/readers/abstract_reader.py
xzdandy: I see. Will we have the nested dictionaries?

eva/optimizer/statement_to_opr_convertor.py
```diff
@@ -222,8 +222,11 @@ def visit_load_data(self, statement: LoadDataStatement):
         if table_metainfo is None:
             # Create a new metadata object
             table_metainfo = create_video_metadata(table_ref.table.table_name)
-
-        load_data_opr = LogicalLoadData(table_metainfo, statement.path)
+            
+        load_data_opr = LogicalLoadData(table_metainfo, 
+                                        statement.path, 
+                                        statement.column_list,
```
gaurav274, line 228: Get the column_metadata object using  column_list.
`column_metadata_list = create_column_metadata(statement.column_list)`

eva/executor/load_csv_executor.py
```diff
@@ -0,0 +1,78 @@
...
+
+        # TODO: Should we move this to utils?
+        # utility function to convert bbox to lisst of floats
+        def convert_bbox(bbox):
+            return np.array([np.float32(coord) for coord in bbox.split(",")])
+
+        # fetch columns to read from meta file
+        column_list = None
+        if self.node.column_list:
+            column_list = [col.col_name for col in self.node.column_list]
```
gaurav274, line 53: Once we replace the column_list with column_metadata list, we can get the column type etc here

gaurav274 (changes requested):

eva/optimizer/statement_to_opr_convertor.py
Anirudh58: Ohh sure sure

eva/optimizer/statement_to_opr_convertor.py
```diff
@@ -222,9 +223,29 @@ def visit_load_data(self, statement: LoadDataStatement):
...
+        self._populate_column_map(table_metainfo)
+
+        # if query had columns specified, we just copy them
+        if statement.column_list is not None:
+            column_list = statement.column_list
+        
+        # else we copy all the columns
+        else:
+            column_list = []
+            for key in self._column_map:
```
gaurav274, line 236: Just realized, we have a bug here. The column_map might have columns from other tables as well. So, you can change the column_map to be a table_metadata_id -> column_name -> columns.

gaurav274 (changes requested):

eva/readers/csv_reader.py
```diff
@@ -23,29 +24,47 @@
...
         """
 
-        self._converters = converters
         self._column_list = column_list
         super().__init__(*args, **kwargs)
 
     def _read(self) -> Iterator[Dict]:
         
+        # TODO: What is a good location to put this code?
+        def convert_csv_string_to_ndarray(row_string):
```
gaurav274, line 41: You can keep it here for now

eva/readers/csv_reader.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # read the csv file
+        df = pd.read_csv(self.file_url)
+        for index, row in df.iterrows():
+            result = {}
+
+            for column in self._column_list:
+
+                # TODO: Is there a better way to do this?
+                if isinstance(row[column.col_name], str) \
```
xzdandy, line 61: Use `pd.read_csv(self.file_url, nrows=0)` to get the column name and then use `pd.read_csv(self.file_url, converters={#column name: convert_csv_string_to_ndarray})`

gaurav274 (approved):

Anirudh58:
> We also need integration test cases (for CSV).

Added in later commit.

xzdandy (approved):
LGTM

PR #229, opened 2022-02-23
Add Gaussian Blur

LordDarkula:
Add filter UDF abstract classes and Gaussian blur filter.


eva/udfs/pytorch_abstract_udf.py
```diff
@@ -113,3 +113,92 @@ def __call__(self, *args, **kwargs):
         if isinstance(frames, pd.DataFrame):
             frames = frames.transpose().values.tolist()[0]
         return nn.Module.__call__(self, frames, **kwargs)
+
+
+class PytorchAbstractFilterUDF(AbstractFilterUDF, nn.Module, GPUCompatible, ABC):
```
gaurav274, line 118: Create new abstract files for such filters

eva/udfs/gaussian_blur.py
```diff
@@ -0,0 +1,59 @@
...
+        self.padding_mode = padding_mode
+
+    @property
+    def input_format(self) -> FrameInfo:
+        # TODO: figure this out
+        return FrameInfo(-1, -1, 3, ColorSpace.RGC)
+
+    def _get_frames(self, frames: Tensor) -> Tensor:
+
+        model = nn.Conv2d(frames[1], frames[1],
```
gaurav274, line 51: Move this to init

eva/udfs/abstract_udfs.py
```diff
@@ -67,3 +67,42 @@ def classify(self, frames: np.ndarray) -> pd.DataFrame:
 
     def __call__(self, *args, **kwargs):
         return self.classify(*args, **kwargs)
+
+
+class AbstractFilterUDF(metaclass=ABCMeta):
```
gaurav274, line 72: Move to new file

eva/udfs/abstract_udfs.py
```diff
@@ -67,3 +67,42 @@ def classify(self, frames: np.ndarray) -> pd.DataFrame:
...
+    """
+    Abstract class for UDFs. All the UDFs which apply filters
```
gaurav274, line 74: Can we not call these filters? Someone might think of them as filtering data

gaurav274 (changes requested):

eva/udfs/gaussian_blur.py
LordDarkula: Isn't the resolution of the frames unknown at initialization? If not, how do I get it?

eva/udfs/abstract_udfs.py
LordDarkula: What should I call them? Effects?

jarulraj:
Redundant PR.

PR #234, opened 2022-03-18
Video stored as compressed files

gaurav274:
1. Video stored as a compressed file and read using OpenCV


eva/eva.yml
```diff
@@ -15,7 +15,10 @@ executor:
   gpus: {'127.0.0.1': [0]}
 storage:
   engine: "eva.storage.petastorm_storage_engine.PetastormStorageEngine"
+  video_engine: "eva.storage.video_storage_engine.VideoStorageEngine"
+  video_engine_version: 0
```
xzdandy, line 19: What is the use for this version?

eva/executor/insert_executor.py
```diff
@@ -49,3 +46,5 @@ def exec(self):
         batch.frames = SchemaUtils.petastorm_type_cast(
             metadata.schema.petastorm_schema, batch.frames)
         StorageEngine.write(metadata, batch)
+        """
+        assert NotImplemented
```
xzdandy, line 50: raise NotImplementedError

eva/storage/video_storage_engine.py
```diff
@@ -0,0 +1,108 @@
...
+            version = int.from_bytes(f.read(4), byteorder='big')
+            if version > self.curr_version:
+                LoggingManager().log('Invalid metadata version {}'
+                                     .format(version), LoggingLevel.ERROR)
+                return False
+            length = int.from_bytes(f.read(4), byteorder='big')
+            path = f.read(length)
+            return Path(path.decode())
+
+    def _create_video_metadata(self, dir_path, video_file):
```
xzdandy, line 90: What is the use for the metadata?

eva/storage/video_storage_engine.py
```diff
@@ -0,0 +1,108 @@
...
+        # File structure
+        # <version> <length> <file_name>
+        with open(dir_path / self.metadata, 'wb') as f:
+            # write version number
+            f.write((self.curr_version).to_bytes(4, byteorder='big'))
+            file_path_bytes = str(video_file).encode()
+            length = len(file_path_bytes)
+            f.write((length).to_bytes(4, byteorder='big'))
+            f.write(file_path_bytes)
```
xzdandy, line 99: use struct.pack and struct.unpack?

eva/storage/video_storage_engine.py
gaurav274: Added this so that in the future if we plan to add support for splitting the video into smaller chunks or even tiling video we can store that info in the metadata

eva/eva.yml
xzdandy: Not sure how this is supposed to work. Right now in `_get_video_file_path`, we compare the version in the metadata and version in the yml file, and reject the request if the version in the metadata is larger then version in the yml file. What you suggested is that say later we have `video_engine_version: 1`, and for metadata having `version 0`, we will still use the version 0 storage engine.

xzdandy: One way is to have a `switch case` in StorageEngine.read and StorageEngine.write.

gaurav274: So the idea is that later as we have another version, we handle them differently.

gaurav274: yml file keeps track of the latest version number

eva/storage/video_storage_engine.py
gaurav274: done

xzdandy (approved):
LGTM

PR #236, opened 2022-04-03
Lateral Join (and Hash Join)

suryatejreddy:
The following type of queries are supported:
- `SELECT id FROM MyVideo, 
LATERAL Unnest(FastRCNNObjectDetector(data))`
- `SELECT id FROM MyVideo, 
LATERAL Unnest(FastRCNNObjectDetector(data)) 
WHERE labels = 'car'`;
- `SELECT id, labels FROM MyVideo,
LATERAL Unnest(FastRCNNObjectDetector(data)) 
WHERE labels = 'car' and Color(data, boxes) = 'BLUE'`;

The PR also adds support for Hash Join on two structured tables based on common columns.


eva/parser/parser_visitor/_table_sources.py
```diff
@@ -33,6 +34,20 @@ def visitTableSources(self, ctx: evaql_parser.TableSourcesContext):
...
+        #For Join Operator
+        if table_sources_count > 1: # Do Not Support More Than 2 Sources
+            left_child = table_list[0]
+            right_source = table_list[1]
+            
+            join_type = None 
+            if right_source.is_func_expr():
+                join_type = JoinType.LATERAL_JOIN
+            elif right_source.is_table_atom():
+                join_type = JoinType.HASH_JOIN
```
gaurav274, line 47: Join can be one of these https://www.techonthenet.com/postgresql/joins.php. Don't call hash a separate type.

gaurav274 (changes requested):

gaurav274 (changes requested):

xzdandy (changes requested):
Need integration test.

PR #240, opened 2022-04-18
Major design modifications to the optimizer

gaurav274:
1. Binder: Consumes a group expression and generates a new copy of the matching operator tree. This operator tree is later converted to the group expression and added back to the memo. 
2. If the pattern has a DummyPattern, the binder tries to match it with any operator type. If a match is found, we return it as a DummyOperator node. The dummy operator node captures the group id of the matching expression. 
3. Operators encapsulated inside group expressions should never have children. Therefore, conversion from the operator tree (created by the parser or returned by binder) to group expression transfers only the content of the operator node to the group expression. The children of the operator are dropped. The parent/child dependency is managed via memo. 
4. Conversion from operator to group expression creates an expression with UNDEFINED_ID and tries to find a duplicate in the memo before inserting a new copy. 
5. Dummy operator nodes are not converted to group expression. Instead, the group id they were matched to is added as a child of the parent group expression. 
6. Memo avoids duplicate inserts via hashing. All the operators, plan nodes, expressions, etc., now supports hashing. 
7. Added new stages in the optimizer tasks. ApplyRule applies the rule to the expression. Explore group generates more equivalent logical expressions (required to handle join-reordering).
8. Things to fix in the future: The operator should have a way to expose only the content. In the current implementation, we expose the content by overriding the `__copy__` to do a shallow copy of all the attributes except children.


eva/expression/comparison_expression.py
```diff
@@ -77,3 +77,6 @@ def __eq__(self, other):
             return False
         return (is_subtree_equal
                 and self.etype == other.etype)
+
+    def __hash__(self) -> int:
+        return super().__hash__()
```
xzdandy, line 82: Is this right?

eva/expression/logical_expression.py
```diff
@@ -55,3 +55,6 @@ def __eq__(self, other):
             return False
         return (is_subtree_equal
                 and self.etype == other.etype)
+
+    def __hash__(self) -> int:
+        return super().__hash__()
```
xzdandy, line 60: Same for logical expression.

eva/optimizer/rules/rules.py
```diff
@@ -563,15 +584,18 @@ def __new__(cls):
...
-            PushdownFilterThroughSample(),
             PushdownProjectThroughSample()
         ]
 
+        self._rewrite_rules = [
+            EmbedFilterIntoGet(),
+            EmbedFilterIntoDerivedGet(),
+            PushdownFilterThroughSample()
+        ]
+
```
xzdandy, line 598: What is the difference between logical rules and rewrite rules?

eva/optimizer/memo.py
```diff
@@ -34,92 +37,78 @@ def groups(self):
     def group_exprs(self):
         return self._group_exprs
 
-    def get_group_id(self, expr: GroupExpression) -> int:
-        """
-        Find whether expr is in any exising group.
-        """
+    def find_duplicate_expr(self, expr: GroupExpression) -> GroupExpression:
         if expr in self.group_exprs:
```
xzdandy, line 41: This should return int.

eva/optimizer/optimizer_tasks.py
```diff
@@ -62,19 +53,17 @@ def task_type(self):
...
+                 rule_set: List[Rule],
                  optimizer_context: OptimizerContext):
-        super().__init__(root_expr, root_expr.group_id,
-                         optimizer_context, OptimizerTaskType.TOP_DOWN_REWRITE)
+        self._root_expr = root_expr
+        self.rule_set = rule_set
+        super().__init__(optimizer_context, OptimizerTaskType.TOP_DOWN_REWRITE)
 
     @property
     def root_expr(self):
```
xzdandy, line 69: Only TopDownRewrite uses @property and @setter. Shall we get rid of it for consistency?

eva/optimizer/plan_generator.py
```diff
@@ -48,37 +49,29 @@ def build_optimal_physical_plan(self,
...
+                           optimizer_context))
         self.execute_task_stack(optimizer_context.task_stack)
-        # Update the group expression
-        grp_expr = memo.groups[root_grp_id].logical_exprs[0]
+
         # BottomUp Rewrite
+        root_expr = memo.groups[root_grp_id].logical_exprs[0]
         optimizer_context.task_stack.push(
-            BottomUpRewrite(grp_expr, optimizer_context))
+            BottomUpRewrite(root_expr, RulesManager().logical_rules,
```
xzdandy, line 67: Why top down rewrite_rules and bottom up logical_rules?

eva/optimizer/optimizer_tasks.py
```diff
@@ -85,172 +74,188 @@ def root_expr(self, expr: GroupExpression):
...
-                )
+            # explore the input group if necessary
+            for idx, child in enumerate(rule.pattern.children):
+                if len(child.children):
+                    child_grp_id = self.root_expr.children[idx]
+                    group = self.optimizer_context.memo.get_group_by_id(
+                        child_grp_id)
+                    self.optimizer_context.task_stack.push(
+                        ExploreGroup(group, self.optimizer_context)
+                    )
```
xzdandy, line 200: Need some help to understand the ExploreGroup.

eva/expression/comparison_expression.py
gaurav274: Abstract Expression will take care of those

xzdandy: Got. Thanks.

eva/optimizer/rules/rules.py
gaurav274: I'll fix the naming.

eva/optimizer/optimizer_tasks.py
gaurav274: Sure, got it

eva/optimizer/plan_generator.py
gaurav274: Will fix this.

eva/optimizer/optimizer_tasks.py
gaurav274: This is to ensure that the child groups are explored. For example, cases where you add a new logical equivalent expression.

PR #241, opened 2022-04-28
Adding Binding Support 

gaurav274:
1. Binding added as a new state in the query processing. The goal is to resolve columns to the source tables. It also aids in validating the query, catching ambiguous columns (same column name in 2 table sources), etc.
2. Each query level maintains 2 key-value maps. One to hold the mapping from table alias to the table object and the other to hold the mapping from child subquery alias to output columns. 
3.  In the current architecture, we resolve each tuple value access to a column alias. In the future, we will replace it with a column index.
4. Another drawback of the current setting: we need to add the alias to the columns after reading from storage and drop the alias before storing.


eva/binder/statement_binder.py
```diff
@@ -0,0 +1,159 @@
...
+from eva.utils.logging_manager import LoggingLevel, LoggingManager
+
+
+class StatementBinder:
+    def __init__(self, binder_context: StatementBinderContext):
+        self._binder_context = binder_context
+        self._catalog = CatalogManager()
+
+    @singledispatchmethod
+    def bind(self, node: AbstractStatement):
```
xzdandy, line 41: Is this type hint correct? It can be other class like tableref.

eva/binder/statement_binder.py
```diff
@@ -0,0 +1,159 @@
...
+        pass
+
+    @bind.register(AbstractExpression)
+    def _bind_abstract_expr(self, node: AbstractExpression):
+        for child in node.children:
```
xzdandy, line 46: To make sure my understanding is correct. For completeness (say type check), we should also bind constant, arithmetic, comparison and etc. It's just they are not in this pr.

gaurav274: Yes. You are correct. I haven't added the data validation in this PR

gaurav274: Yes

eva/binder/statement_binder.py
gaurav274: Fixed

eva/binder/statement_binder.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @bind.register(LoadDataStatement)
+    def _bind_load_data_statement(self, node: LoadDataStatement):
+        table_ref = node.table_ref
+        if node.file_options['file_format'] == FileFormatType.VIDEO:
+            # Create a new metadata object
+            create_video_metadata(table_ref.table.table_name)
+            self.bind(table_ref)
+        else:
+            self.bind(table_ref)
```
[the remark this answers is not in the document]
gaurav274: Fixed The else part is supposed to take care of it.

eva/models/storage/batch.py
```diff
@@ -270,3 +276,27 @@ def reverse(self):
...
         self._frames.reset_index(drop=True, inplace=True)
+
+    def modify_column_alias(self, alias: str) -> NoReturn:
+        # a, b, c -> table1.a, table1.b, table1.c
+        # t1.a -> t2.a
+        new_col_names = []
+        for col_name in self.frames.columns:
+            if '.' in col_name:
+                new_col_names.append('{}.{}'.format(
+                    alias, col_name.split('.')[1]))
```
xzdandy, line 287: We will change this to be index-based later right? This feels like a hack method.

eva/expression/function_expression.py
```diff
@@ -13,63 +13,46 @@
...
-        is_temp (bool, default:False): In case of EXEC type, decides if the
-        outcome needs to be stored in BatchFrame temporarily.
+    `output`: If the user wants only subset of ouputs. Eg,
+    ObjDetector.lables the parser with set output to 'labels'
 
-        output(str): The column to return after executing function
-
-        output_obj(UdfIO): The catalog object corresponding to the func_output.
-        To be populated by optimizer.
+    `ourput_col_aliases`: It is populated by the binder. In case the
```
xzdandy, line 34: typo, output_col_aliases

eva/server/command_handler.py
```diff
@@ -32,6 +34,10 @@ def execute_query(query) -> Iterator[Batch]:
     Execute the query and return a result generator.
     """
     stmt = Parser().parse(query)[0]
+    try:
+        StatementBinder(StatementBinderContext()).bind(stmt)
+    except Exception:
+        raise RuntimeError('Invalid Query')
```
xzdandy, line 40: The binder's exception contains why it fails. Shall we include that instead of only "Invalid Query"?

eva/models/storage/batch.py
gaurav274: Yeah, I know. There was a lot to change so went with this design. Surely, we got to change it to index. I haven't thought of how we would do it.

gaurav274:
It is just a line. So should be fine I guess

xzdandy (approved):
Overall LGTM.

xzdandy:
> It is just a line. So should be fine I guess

In the test coverage page: https://coveralls.io/builds/49204364

0 | -95.83% | eva/udfs/ndarray_udfs/unnest.py

PR #242, opened 2022-05-12
Update config.yml

gaurav274:
Bump up the python version for testing


xzdandy:
Shall we also update the conda and pip stuff for consistency?

gaurav274:
Do we have python version dependency in conda and pip?

xzdandy (approved):

PR #243, opened 2022-05-15
Join Support Added

gaurav274:
1. Lateral and hash join supported.
2. Join predicates needs to be specified using `on` keyword
3. Only Equi joins are supported.
4. Dummy Cost model added
5. Join associativity rule added


eva/executor/function_scan_executor.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+
+    def __init__(self, node: FunctionScanPlan):
+        super().__init__(node)
+        self.func_expr = node.func_expr
+
+    def validate(self):
+        pass
+
+    def exec(self, lateral_input: Batch, *args, **kwargs) -> Iterator[Batch]:
```
xzdandy, line 37: Consider the following plan:
```mermaid
graph TD;
  APPLY-->SELECTION;
  SELECTION-->FUNCTIONSCAN;
```
Now the Selection's executor also needs to pass `lateral_input`to its children's `exec()`. In other words, all operators between APPLY and FUNCTIONSCAN have to call their children's `exec()` differently. Can we make it anonymously to other operators? E.g., having `lateral_input` in `**kwargs` and passing `**kwargs` to children's `exec()` always.

eva/executor/join_build_executor.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def exec(self, *args, **kwargs) -> Iterator[Batch]:
+        child_executor = self.children[0]
+        # build in memory hash table and pass to the probe phase
+        # Assumption the hash table fits in memory
+        # Todo: Implement a partition based hash join (grace hash join)
+        cumm_batches = []
+        for batch in child_executor.exec():
+            if not batch.empty():
+                cumm_batches.append(batch)
```
xzdandy, line 27: This can be simplified  `cumm_batches = [batch for batch in child_executor.exec() if not batch.empty()]`

eva/executor/hash_join_executor.py
```diff
@@ -0,0 +1,60 @@
...
+                if not join_batch.empty() and self.predicate:
+                    outcomes = self.predicate.evaluate(join_batch).frames
+                    filtered_df = join_batch.frames[(outcomes > 0).to_numpy()]
+                    join_batch = Batch(filtered_df.reset_index(drop=True))
+                # Then do project
+                if not join_batch.empty() and self.join_project:
+                    batches = [expr.evaluate(join_batch)
+                               for expr in self.join_project]
+                    join_batch = Batch.merge_column_wise(batches)
+
```
xzdandy, line 59: What is the benefit of having the selection and projection in the join? Instead of say a seqscan's executor above the Join (better modularity). If we later want to make optimizations to the selection and projection's executor , we don't need to change at multiple places.

eva/executor/lateral_join_executor.py
```diff
@@ -0,0 +1,54 @@
...
+                    outcomes = self.predicate.evaluate(result_batch).frames
+                    result_batch = Batch(result_batch.frames
+                                         [(outcomes > 0).to_numpy()]
+                                         .reset_index(drop=True))
+                # Then do project
+                if not result_batch.empty() and self.join_project:
+                    batches = [expr.evaluate(result_batch)
+                               for expr in self.join_project]
+                    result_batch = Batch.merge_column_wise(batches)
+
```
xzdandy, line 52: Similar to the hash join, we are having the same selection and projection code multiple times.

eva/optimizer/group.py
```diff
@@ -49,6 +50,10 @@ def __init__(self, group_id: int):
     def group_id(self):
         return self._group_id
 
+    @property
+    def aliases(self):
+        return self._aliases
```
xzdandy, line 55: aliases are more like output objects of the operators (given operators are logically equivalent, the all should have the same output objects) in the group?

eva/optimizer/memo.py
```diff
@@ -54,12 +56,27 @@ def get_group_by_id(self, group_id: int) -> GroupExpression:
...
+        Collects table aliases of all the children
+        """
+        aliases = []
+        for child_grp_id in expr.children:
+            child_grp = self._groups[child_grp_id]
+            aliases.extend(child_grp.aliases)
+        if (expr.opr.opr_type == OperatorType.LOGICALGET or
+                expr.opr.opr_type == OperatorType.LOGICALQUERYDERIVEDGET):
+            aliases.append(expr.opr.alias)
+
```
xzdandy, line 70: If my previous comments are valid, can we get the aliases from the operator instead of collecting from the children?

xzdandy:
Shall we add a unit test cases for cost model? A simple one can be one group has two physical plans, and the optimizer will return the lower one. A complex one can be a tree, the optimizer will return a path from root has the lowest cost?

eva/executor/function_scan_executor.py
gaurav274: Sure, It should be that way

eva/optimizer/group.py
gaurav274: What do you mean?

eva/executor/hash_join_executor.py
gaurav274: Seq scan operator?

suryatejreddy: I'm not sure if I understand this completely, but I believe you added the predicate and project executor and not using them here?

eva/optimizer/group.py
suryatejreddy: @xzdandy do you mean that we need to extract it from whichever operator has the TableRef?

eva/optimizer/operators.py
```diff
@@ -621,6 +625,123 @@ def __hash__(self) -> int:
...
+    Attributes:
+        join_type: JoinType
+            Join type provided by the user - Lateral, Inner, Outer
+        join_predicate: AbstractExpression
+            condition/predicate expression used to join the tables
+    """
+
+    def __init__(self,
+                 join_type: JoinType,
+                 join_predicate: AbstractExpression = None,
```
suryatejreddy, line 672: Shouldn't we have a LogicalPredicate Operator if we have a plan and executor?

suryatejreddy: Same for project I guess.

suryatejreddy (approved):

eva/executor/hash_join_executor.py
xzdandy: Yes, 
```mermaid
flowchart TD;
   subgraph ide2 [option two]
   PROJECTION --- SELECTION
   SELECTION --- B(JOIN)
   end
   subgraph ide1 [option one]
   SEQSCAN --- A(JOIN)
   end
```
Merging Projection and Selection into Get can be beneficial because storage engine may optimize the reading performance given the predicate. 
But for hash join and lateral join, I am not sure how knowing the projection columns and predicates can benefit.

eva/optimizer/group.py
xzdandy: I think that after binder, every operator should have output objects binded (called aliases?). Correct me if I am wrong.

eva/executor/hash_join_executor.py
gaurav274: Option one does not make sense. Seqscan is always done on a table. I understand we have a separate node for storage but by definition, it should be part of seqscan.
 
Right now we are using option 2.

eva/optimizer/operators.py
gaurav274: Do you mean the type of join_predicate should be LogicalPredicate?
If yes, then no. Right now we don't have the rule to push predicates into join so the optimised plan will always have a PredicateNode above join. I kept join_predicate in case we plan to push them later.

eva/optimizer/group.py
gaurav274: I should change it from list to set to avoid keeping duplicates

gaurav274: Does that answer the concern?

gaurav274:
Do you have any simple ideas on how we can write the cost model test cases? I'm thinking of mocking the cost model to control the cost.

eva/executor/hash_join_executor.py
gaurav274: I thought about it. To make things modular we can move predicate and projection logic to a util function. What do you think? I hope that should be fine with the ray right?

xzdandy: For ray, it prefers option 2, since it has more options to insert the exchange operator for different parralision configuration. PROJECTION and SELECTOIN needs to be different operator nodes. They can not be within the JOIN exectutor. In other words, JOIN needs to only do the join job. Otherwise we can not inject the exchange operator.  
Regarding SEQSCAN, I am a bit confused. Is SEQSCAN === PROJECTION + SELECTION?

xzdandy:
> Do you have any simple ideas on how we can write the cost model test cases? I'm thinking of mocking the cost model to control the cost.

I was thinking of mocking the plans. So create a cost model, for planA, return cost 1, for planB, return cost 2. Then have a group with physical_expr contains [planA, planB]. Then run optmizeInputs, and check the output plan tree.

eva/executor/hash_join_executor.py
gaurav274: Sounds good.
SEQSCAN **not equal** PROJECTION + SELECTION. It means iterating over the data. Ideally, our storage node + our seqscan node == true SeqScan

xzdandy (approved):
LGTM

PR #251, opened 2022-05-19
Prod packaging

suryatejreddy:


setup.cfg
```diff
@@ -1,8 +1,8 @@
 [metadata]
-name = evatestdb
-version = 0.0.55
+name = evadb
+version = 0.0.1
 author = Georgia Tech Database Group
-author_email = arulraj@gatech.edu
+author_email = georgia.tech.db@gmail.com
```
gaurav274, line 5: Is this a valid email?

gaurav274 (approved):

setup.cfg
xzdandy: We can consider creating a group such as eva@groups.gatech.edu that includes all contributors.

PR #254, opened 2022-06-02
Removes Torch Dependencies

suryatejreddy:


gaurav274:
1. Also add the `try catch` changes
2. The build doesn't seem to work

jarulraj:
Being addressed in -- #277

PR #259, opened 2022-07-01
EVA Website

EDGAhab:


gaurav274:
@EDGAhab Why do we have the entire eva repository inside the website?
And is it possible to directly use the Readme on GitHub rather than making a copy inside the website folder? If not, please update the readme. Thanks!

jarulraj:
Can we get rid of the eva source code?

_config.yml
```diff
@@ -0,0 +1,60 @@
...
+logo: Logo.png
+
+title: EVA
+blank: 
+description: Visual data management system. Features include declarative query language, storage management, deep learning, and hardware acceleration. Developed at Georgia Institute of Technology.
+
+baseurl: "/eva" # the subpath of your site, e.g. /blog
+url: "https://georgia-tech-db.github.io" # the base hostname & protocol for your site, e.g. http://example.com
+
+show_downloads: true
```
jarulraj, line 29: We can disable this option.

EDGAhab:
> Can we get rid of the eva source code?

I just got rid of the source code!

EDGAhab:
> @EDGAhab Why do we have the entire eva repository inside the website? And is it possible to directly use the Readme on GitHub rather than making a copy inside the website folder? If not, please update the readme. Thanks!

I deleted the eva repo and updated the readme.md!

_config.yml
EDGAhab: Okay! Disabled

PR #260, opened 2022-07-01
Update README.md

xzdandy:
More example queries based on today's discussion.

TODO: Example query for JOIN LATERAL.


jarulraj (approved):
Looks good to me

PR #262, opened 2022-07-02
Error Message Ehancement

xzdandy:
- fix error message so it is not truncated at the client
- enable parser error listener


gaurav274 (approved):

eva/parser/parser.py
```diff
@@ -35,20 +35,20 @@ def syntaxError(self, recognizer, offendingSymbol, line, column, msg, e):
                     str(column) + " - " + str(msg)
         raise Exception(error_str)
 
-    def reportAmbiguity(self, recognizer, dfa, startIndex, stopIndex,
-                        exact, ambigAlts, configs):
-        error_str = "ERROR: Ambiguity -" + str(configs)
-        raise Exception(error_str)
+    # def reportAmbiguity(self, recognizer, dfa, startIndex, stopIndex,
```
gaurav274, line 38: Can you share the reason for removing these?

gaurav274: Given we are changing things, please also change the class name from MyErrorListener to something better maybe AntlrErrorListener

eva/parser/parser.py
```diff
@@ -74,7 +74,7 @@ def parse(self, query_string: str) -> list:
 
         parser = evaql_parser(stream)
         # Attach error listener for debugging parser errors
-        # parser._listeners = [self._error_listener]
+        parser._listeners = [self._error_listener]
```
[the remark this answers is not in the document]
xzdandy:
```
def reportAmbiguity(self, recognizer, dfa, startIndex, stopIndex,
                        exact, ambigAlts, configs):
        error_str = "ERROR: Ambiguity -" + str(configs)
        raise Exception(error_str)

    def reportAttemptingFullContext(self, recognizer, dfa, startIndex,
                                    stopIndex, conflictingAlts, configs):
        error_str = "ERROR: Attempting Full Context -" + str(configs)
        raise Exception(error_str)

    def reportContextSensitivity(self, recognizer, dfa, startIndex,
                                 stopIndex, prediction, configs):
        error_str = "ERROR: Context Sensitivity -" + str(configs)
        raise Exception(error_str)
``` 
These three breaks the testcases. The first one syntax is not.

eva/parser/parser.py
xzdandy: I think our parser definition is incrementally defined over time by different people. There are duplication and ambiguity. 
This breaks almost all testcases involving parser.

eva/parser/parser.py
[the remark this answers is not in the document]
jarulraj: Should we remove those three functions instead of commenting them out? Is it better to fix the parser as opposed to disabling them?

eva/parser/parser.py
gaurav274: Ah! That is bad. How severe is it? Maybe create an issue and we can skip it for now.

xzdandy: Create an issue: #264

eva/parser/parser.py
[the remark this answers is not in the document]
xzdandy: I think it is better to fix the parser which may take some efforts. That's why I commented them out, so we can re-enable them later.

PR #267, opened 2022-07-04
feat(binder): add support for select *

xzdandy:
Extend the binder for `SELECT * FROM ...`

TODO: More testcases needed.

### DIFF
--- a/eva/binder/binder_utils.py
+++ b/eva/binder/binder_utils.py
@@ -20,6 +20,9 @@ from eva.catalog.models.df_metadata import DataFrameMetadata
 from eva.parser.create_statement import ColConstraintInfo, ColumnDefinition
 from eva.parser.table_ref import TableInfo, TableRef
 from eva.utils.generic_utils import generate_file_path
+from eva.binder.statement_binder_context import StatementBinderContext
+from eva.expression.abstract_expression import AbstractExpression
+from eva.expression.tuple_value_expression import TupleValueExpression
 from eva.utils.logging_manager import logger
 
 
@@ -125,3 +128,24 @@ def handle_if_not_exists(table_ref: TableRef, if_not_exist=False):
             raise RuntimeError(err_msg)
     else:
         return False
+
+
+def extend_star_in_target_list(
+    target_list: List[AbstractExpression],
+    alias: str,
+    binder_context: StatementBinderContext
+):
+    idx = 0
+    while idx < len(target_list):
+        expr = target_list[idx]
+        if isinstance(expr, TupleValueExpression) and expr.col_name == '*':
+            del target_list[idx]
+            col_names = binder_context._search_all_col_name(alias)
+            for col_name in col_names:
+                target_list.insert(
+                    idx,
+                    TupleValueExpression(col_name=col_name, table_alias=alias)
+                )
+                idx += 1
+        else:
+            idx += 1
--- a/eva/binder/statement_binder.py
+++ b/eva/binder/statement_binder.py
@@ -14,7 +14,8 @@
 # limitations under the License.
 import sys
 from eva.binder.statement_binder_context import StatementBinderContext
-from eva.binder.binder_utils import bind_table_info, create_video_metadata
+from eva.binder.binder_utils import bind_table_info, create_video_metadata, \
+    extend_star_in_target_list
 from eva.catalog.catalog_manager import CatalogManager
 from eva.expression.abstract_expression import AbstractExpression
 from eva.expression.function_expression import FunctionExpression
@@ -69,6 +70,11 @@ class StatementBinder:
         if node.where_clause:
             self.bind(node.where_clause)
         if node.target_list:
+            extend_star_in_target_list(
+                node.target_list,
+                node.from_table.alias,
+                self._binder_context
+            )
             for expr in node.target_list:
                 self.bind(expr)
         if node.orderby_list:
--- a/eva/binder/statement_binder_context.py
+++ b/eva/binder/statement_binder_context.py
@@ -91,6 +91,17 @@ class StatementBinderContext:
             for obj in col_objs:
                 if obj.name.lower() == col_name:
                     return obj
+    
+    def _search_all_col_name(self, alias: str) -> List[str]:
+        table_obj = self._table_alias_map.get(alias, None)
+        if table_obj:
+            return [col.name for col in table_obj.columns]
+        col_objs = self._derived_table_alias_map.get(alias, None)
+        if col_objs:
+            return [obj.name for obj in col_objs]
+        err_msg = 'Cannot find alias = {}'.format(alias)
+        logger.error(err_msg)
+        raise RuntimeError(err_msg)
 
     def _search_all_alias_maps(self,
                                col_name: str) -> Tuple[str, CatalogColumnType]:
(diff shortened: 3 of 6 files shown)


jarulraj:
Here are some more test cases. Is it ever possible that both base table and derived tables don't exist -- that code is leading to coverage drop?

```
+        sub_query = """SELECT * FROM MyVideo"""
+        nested_query = """SELECT * FROM ({}) AS T;""".format(sub_query)
+        actual_batch = execute_query_fetch_all(nested_query)
+        actual_batch.sort()
+        expected_batch = list(create_dummy_batches())
+        self.assertEqual(actual_batch.batch_size, expected_batch[0].batch_size)
+
+        with self.assertRaises(Exception):
+            invalid_select_query = "SELECT * FROM MyVideo5;"
+            actual_batch = execute_query_fetch_all(invalid_select_query)
```

eva/binder/statement_binder_context.py
```diff
@@ -91,6 +91,17 @@ def _check_derived_table_alias_map(self,
             for obj in col_objs:
                 if obj.name.lower() == col_name:
                     return obj
+    
+    def _search_all_col_name(self, alias: str) -> List[str]:
```
jarulraj, line 95: We should add some docstrings to functions in this file.

test/integration_tests/test_select_executor.py
```diff
@@ -88,6 +88,13 @@ def test_should_load_and_select_in_table(self):
         actual_batch.sort()
         expected_batch = list(create_dummy_batches())
         self.assertEqual([actual_batch], expected_batch)
+    
+    def test_should_select_star_in_table(self):
```
jarulraj, line 92: Gave suggestions for two more test cases.

jarulraj (approved):

eva/binder/statement_binder_context.py
```diff
@@ -91,6 +91,17 @@ def _check_derived_table_alias_map(self,
...
+        table_obj = self._table_alias_map.get(alias, None)
+        if table_obj:
+            return [col.name for col in table_obj.columns]
+        col_objs = self._derived_table_alias_map.get(alias, None)
+        if col_objs:
+            return [obj.name for obj in col_objs]
+        err_msg = 'Cannot find alias = {}'.format(alias)
+        logger.error(err_msg)
```
gaurav274, line 103: @jarulraj With the new logger, does the `logger.exception(error)` also raise the exception (so that we can merge lines 103-104)?

test/integration_tests/test_select_executor.py
gaurav274: Maybe add test with nested query and join query

eva/binder/statement_binder_context.py
jarulraj: Not sure. We could add that functionality.

xzdandy:
> 

It should not, if it works as expected.

test/integration_tests/test_select_executor.py
xzdandy: I will add nested query and join.

eva/binder/statement_binder.py
```diff
@@ -69,6 +70,11 @@ def _bind_select_statement(self, node: SelectStatement):
         if node.where_clause:
             self.bind(node.where_clause)
         if node.target_list:
+            extend_star_in_target_list(
```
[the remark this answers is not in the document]
xzdandy: I assume this is also valid `SELECT *, YOLO(data) FROM MyVideo?` If not, we can just replace instead of extending.

eva/binder/binder_utils.py
```diff
@@ -125,3 +128,24 @@ def handle_if_not_exists(table_ref: TableRef, if_not_exist=False):
             raise RuntimeError(err_msg)
     else:
         return False
+
+
+def extend_star_in_target_list(
```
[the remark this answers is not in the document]
xzdandy: See above

eva/binder/binder_utils.py
```diff
@@ -125,3 +128,24 @@ def handle_if_not_exists(table_ref: TableRef, if_not_exist=False):
...
+    target_list: List[AbstractExpression],
+    alias: str,
+    binder_context: StatementBinderContext
+):
+    idx = 0
+    while idx < len(target_list):
+        expr = target_list[idx]
+        if isinstance(expr, TupleValueExpression) and expr.col_name == '*':
```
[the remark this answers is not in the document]
xzdandy: Would you elaborate? `target_list` is  a list of `AbstractExpression`. What is the enum?

eva/binder/statement_binder_context.py
xzdandy: I don't think so. `logger.exception(err_msg)` is equivalent to `logger.error(err_msg, exc_info=True)`

eva/binder/statement_binder_context.py
```diff
@@ -60,6 +80,15 @@ def add_derived_table_alias(self,
     def get_binded_column(self,
                           col_name: str,
                           alias: str = None) -> Tuple[str, CatalogColumnType]:
```
xzdandy, line 82: @gaurav274 The return columns object is `CatalogColumnType` or `DataFrameColumn`? Below:
- `_check_table_alias_map` and `_check_derived_table_alias_map` returns `DataFrameColumn`
- `_search_all_alias_maps` returns `CatalogColumnType`

eva/binder/binder_utils.py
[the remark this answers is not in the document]
gaurav274: Ignore

eva/binder/statement_binder_context.py
gaurav274: _check_derived_table_alias_map should have been CatalogColumnType

eva/binder/statement_binder.py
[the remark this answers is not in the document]
gaurav274: Yes, * is typically used alone

eva/binder/statement_binder_context.py
xzdandy: How about `_check_table_alias_map`?  `get_binded_column` directly returns the `col_object` from the  `_check_table_alias_map`. So their type should be same?

jarulraj:
@gaurav274 Is this good to merge?

test/integration_tests/test_select_executor.py
```diff
@@ -202,7 +240,7 @@ def test_lateral_join(self):
         self.assertEqual(actual_batch.batch_size, 5)
 
     def test_aahash_join_with_one_on(self):
-        select_query = """SELECT table1.a2 FROM table1 JOIN
+        select_query = """SELECT * FROM table1 JOIN
```
gaurav274, line 243: We should also keep the existing test cases

test/integration_tests/test_select_executor.py
```diff
@@ -88,21 +88,59 @@ def test_should_load_and_select_in_table(self):
...
+
+        select_query = """SELECT * FROM (SELECT id FROM MyVideo) AS T;"""
+        actual_batch = execute_query_fetch_all(select_query)
+        actual_batch.sort()
+        expected_rows = [{"T.id": i} for i in range(NUM_FRAMES)]
+        expected_batch = Batch(frames=pd.DataFrame(expected_rows))
+        self.assertEqual(actual_batch, expected_batch)
+
+    @unittest.skip('Not supported in current version')
+    def test_select_star_in_lateral_join(self):
```
gaurav274, line 120: Shouldn't it return all the columns (including columns from UDF)? SO basically 4 columns?

xzdandy: The column binding for function or lateral join is not correct. The assert is a placeholder. But the query does not run. So I am not able to figure out the actual output column names.

PR #269, opened 2022-07-05
feat: SHOW UDFS;

gaurav274:


eva/parser/evaql/evaql_parser.g4
```diff
@@ -358,6 +358,10 @@ helpStatement
     : HELP STRING_LITERAL
     ;
 
+showStatement
+    : SHOW (UDFS | TABLES)
```
jarulraj, line 362: We could add support for DATASETS as well

jarulraj (approved):

PR #270, opened 2022-07-08
Combine UPLOAD and LOAD commands

pchunduri6:
1. Combine UPLOAD and LOAD statements to support queries of the format:

`UPLOAD INFILE 'data/ua_detrac/ua_detrac.mp4' INTO MyVideo;`

Statements of the following type are also supported:
```
UPLOAD INFILE 'data/ua_detrac/ua_detrac.mp4'
INTO MyVideo
WITH FORMAT VIDEO;
```

All the options associated with `LOAD` (including `file_options` and `column_list`) are also supported.

2. The dependency on `/tmp` is removed. Any uploaded file is now stored in `~/.eva/tmp/` by default.
3. `path_prefix` has been renamed as `upload_dir` and can be specified in the `eva.yml` config file.


eva/binder/statement_binder.py
```diff
@@ -129,6 +130,42 @@ def _bind_load_data_statement(self, node: LoadDataStatement):
 
         node.column_list = column_list
 
+    @bind.register(UploadStatement)
```
gaurav274, line 133: If the upload binder is mostly reusing the code from the load binder, you can do something like this
```suggestion
    @bind.register(UploadStatement)
    @bind.register(LoadDataStatement)
```

eva/eva.yml
```diff
@@ -17,7 +17,7 @@ storage:
   engine: "eva.storage.petastorm_storage_engine.PetastormStorageEngine"
   video_engine: "eva.storage.opencv_storage_engine.OpenCVStorageEngine"
   video_engine_version: 0
-  path_prefix: "/tmp"
+  upload_dir: ""
```
gaurav274, line 20: We can avoid exposing upload_dir and always assume `tmp` implicit to our system.

eva/server/db_api.py
```diff
@@ -76,15 +77,20 @@ def _upload_transformation(self, query: str) -> str:
          the content into a string.
         """
         if 'UPLOAD' in query:
-            file_path = query.split()[2][1:-1]
-            dst_path = query.split()[-1][1:-2]
+            query_list = query.split()
```
gaurav274, line 80: Please drop a comment here. We might forget what is happening here.

eva/server/db_api.py
```diff
@@ -76,15 +77,20 @@ def _upload_transformation(self, query: str) -> str:
...
+            file_path = query_list[2][1:-1]
+            dst_path = os.path.basename(file_path)
             with open(file_path, "rb") as f:
                 bytes_read = f.read()
                 b64_string = str(base64.b64encode(bytes_read))
                 query = 'UPLOAD PATH ' + \
```
gaurav274, line 86: And also maybe use `'{}'.format` to clean it up

test/executor/test_load_executor.py
```diff
@@ -135,3 +135,10 @@ def test_should_call_csv_reader_and_storage_engine(
 
         # remove the dummy.csv
         file_remove('dummy.csv')
+
+
+if __name__ == '__main__':
```
gaurav274, line 140: We can remove this from all the test cases

eva/eva.yml
gaurav274: Internally, we refer to it as the Upload directory so the logic still remains the same

gaurav274 (commented):
Overall LGTM. 
1. Fix the readme with the new changes

test/integration_tests/test_upload_executor.py
```diff
@@ -16,26 +16,98 @@
...
-        expected_blob = "b'AAAA'"
-        with open(os.path.join(PATH_PREFIX, 'dummy.avi'), 'rb') as f:
+        expected_blob = self.video_blob
+        with open(os.path.join(UPLOAD_DIR, 'dummy.avi'), 'rb') as f:
             bytes_read = f.read()
             actual_blob = str(base64.b64encode(bytes_read))
         self.assertEqual(actual_blob, expected_blob)
+
+    # integration test for video
+    def test_should_upload_video_to_table(self):
```
gaurav274, line 53: Add a test case without the `FORMAT VIDEO` keyword

jarulraj:
Refer #315

eva/server/db_api.py
jarulraj: Resolved.

eva/server/db_api.py
jarulraj: Needs to be resolved.

jarulraj:
@pchunduri6  Please resolve these two pending issues in the master branch.

PR #273, opened 2022-07-27
Predicate Pushdown for video table

gaurav274:
1. Push down the predicate on the video table to avoid decoding all the frames.
2. Supports simple range-based and equality-based predicates.


eva/optimizer/rules/rules.py
```diff
@@ -238,22 +241,44 @@ def __init__(self):
...
-            lget.video,
-            lget.dataset_metadata,
-            alias=lget.alias,
-            predicate=predicate,
-            target_list=lget.target_list,
-            children=lget.children,
+        # System only supports pushing basic range predicates on id
+        video_alias = lget.video.alias
+        col_alias = f"{video_alias}.id"
+        pushdown_pred, unsupported_pred = extract_pushdown_predicate(
```
xzdandy, line 263: Shall we try avoiding doing the same computation in rule check and apply?

eva/expression/expression_utils.py
```diff
@@ -28,3 +35,234 @@ def expression_tree_to_conjunction_list(expression_tree):
...
+        upper_bound (int): upper bound of the comparison predicate
+
+    Returns:
+        List[Tuple(int)]: list of valid ranges
+
+    Raises:
+        RuntimeError: Invalid expression
+
+    Example:
+        extract_range_from_comparison_expr(id < 10, 0, inf): True, [(0,9)]
```
xzdandy, line 79: How do we input `inf`?

xzdandy (approved):
Overall LGTM

eva/optimizer/rules/rules.py
gaurav274: I thought of it but the only way to do it is to keep the state across rule check and apply, which feels wierd to me and might creep in some bugs. Do you any other thoughts?

gaurav274: I thought of it. But, the only way to avoid it is by keeping the state across check and apply which might creep in bugs.

eva/expression/expression_utils.py
gaurav274: Inf is just an example. For actual code, we would always know the upper and lower bound. The ultimate goal is to get it from catalog and pass it down the pipeline.

eva/optimizer/rules/rules.py
gaurav274: Btw, do you have any thoughts on it?

xzdandy: We can either keep the state in rule or in optimizer context. I think both we need to clean up afterwards. To enforce the cleaning up, we may can implement a step in the base class, though not sure it is a good idea or not.

gaurav274: I like the `optimizer context` idea. But it is too much engineering for now. I do not think the computation is expensive enough for us to spend time on this optimization.

PR #276, opened 2022-07-30
Website updates

jarulraj:
-- Created a basic Jekyll based website
-- Links are in `_templates/default.html`


docs/Gemfile.lock
```diff
@@ -0,0 +1,299 @@
+GEM
```
gaurav274, line 1: Do we need this file?

gaurav274 (approved):

docs/Gemfile.lock
jarulraj: Looks like it is needed -- https://stackoverflow.com/questions/4151495/should-gemfile-lock-be-included-in-gitignore

PR #287, opened 2022-07-31
Documentation updates

gaurav274:


jarulraj:
It seems like only the docs for the latest version is shown.

gaurav274:
Can we merge this? It has a bunch of other changes to the readme and documentation.

jarulraj (approved):

PR #290, opened 2022-08-01
Ray executor

xzdandy:
1. Add exchange operator
2. Add execution plan for exchange operator (parallelism)

TODO: CLEAN UP
TODO: create a rule to generate a default plan with exchange operator


gaurav274:
Merged with #382

PR #296, opened 2022-08-05
Restructure Abstract UDF classes to allow for the future addition of EVA transforms

LordDarkula:
Create `AbstractUDF` base class and update `AbstractClassifierUDF` base class to inherit from it
Create `AbstractTransformationUDF` base class
Create `PytorchAbstractTransformationUDF` to combine torchvision transforms and  `AbstractTransformationUDF`
Rename `PytorchAbstractUDF` to `PytorchClassifierAbstractUDF`


eva/udfs/pytorch_abstract_udf.py
```diff
@@ -112,3 +110,11 @@ def __call__(self, *args, **kwargs):
         if isinstance(frames, pd.DataFrame):
             frames = frames.transpose().values.tolist()[0]
         return nn.Module.__call__(self, frames, **kwargs)
+
+
+class PytorchAbstractTransformationUDF(AbstractTransformationUDF, Compose):
```
LordDarkula, line 115: I am somewhat if `PytorchAbstractTransformationUDF` should subclass torchvision transforms or nn.Module. A neural network can also be used to transform a frame. Let me know what you think.

jarulraj: Doesn't `PytorchAbstractUDF` itself inherit from `nn.Module`?

jarulraj:
do we need `input format` everywhere -- that is leading to a slight drop in coverage in #303 ?

jarulraj:
@LordDarkula @gaurav274 Is it possible to use some kind of `reflection` to use torchvision transforms instead of writing a separate UDF for each transform? -- https://stackoverflow.com/questions/37628/what-is-reflection-and-why-is-it-useful

LordDarkula:
> @LordDarkula @gaurav274 Is it possible to use some kind of `reflection` to use torchvision transforms instead of writing a separate UDF for each transform? -- https://stackoverflow.com/questions/37628/what-is-reflection-and-why-is-it-useful

I assume you mean allow the user to just use a torchvision transform directly in EVA instead of having to explicitly subclass PytorchAbstractUDF. Assuming we want EVA to be type-safe, we could use a Protocol (https://peps.python.org/pep-0544/). However, this feature was introduced in Python 3.8 and EVA supports earlier versions of Python.

jarulraj:
This looks more like the approach I suggested:

1. https://docs.bentoml.org/en/v1.0.0-a7/frameworks/easyocr.html
2. https://docs.bentoml.org/en/0.13-lts/frameworks.html#bentoml.frameworks.easyocr.EasyOCRArtifact

jarulraj (approved):

PR #303, opened 2022-08-09
(feat) Face Detector

jarulraj:
-- Added UDF and an integration test case

### DIFF
--- a/eva/udfs/face_detector.py
+++ b/eva/udfs/face_detector.py
@@ -1,0 +1,93 @@
+# coding=utf-8
+# Copyright 2018-2022 EVA
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


jarulraj:
Do we need `labels` and `input format` properties for all UDFs?

gaurav274:
No, we don't need them. In fact, we are not yet enforcing the `input_format`.

LordDarkula:
I added defaults for `name` and `input_format` in #296

eva/udfs/face_detector.py
```diff
@@ -0,0 +1,80 @@
...
+        Performs predictions on input frames
+        Arguments:
+            frames (np.ndarray): Frames on which predictions need
+            to be performed
+        Returns:
+            face boxes (List[List[BoundingBox]])
+        """
+
+        copy = torch.squeeze(frames)
+        transform = T.ToPILImage()
```
gaurav274, line 67: This converison is wierd. Are we doing it frame by frame?

jarulraj: Should we do it in some other manner?

```
        pil_image = T.ToPILImage(torch.squeeze(frames))
        bboxes, scores = self.model.detect(img=pil_image)
```

jarulraj: model takes in ndarray or PIL image

gaurav274: We need to provide a cleaner interface.

PR #309, opened 2022-08-10
feat: Add support for unnest with lateral join

gaurav274:


eva/catalog/schema_utils.py
```diff
@@ -40,15 +40,15 @@ def get_petastorm_column(df_column):
...
+                np.int64,
                 (),
                 ScalarCodec(IntegerType()),
                 column_is_nullable,
             )
         elif column_type == ColumnType.FLOAT:
             petastorm_column = UnischemaField(
                 column_name,
-                np.float64,
+                np.float32,
```
xzdandy, line 51: What is the motivation of these changes?

eva/parser/parser_visitor/_table_sources.py
```diff
@@ -176,3 +186,12 @@ def visitFromClause(self, ctx: evaql_parser.FromClauseContext):
             where_clause = self.visit(ctx.whereExpr)
 
         return {"from": from_table, "where": where_clause}
+
+    def visitAliasClause(self, ctx: evaql_parser.AliasClauseContext):
+        alias_name = self.visit(ctx.uid())
+        column_list = []
+        if ctx.uidList():
+            column_list = self.visit(ctx.uidList())
```
xzdandy, line 194: I feel it becomes harder over time to figure out which `visit` function it actually calls. Any idea to improve the readability?

eva/catalog/schema_utils.py
gaurav274: I changed it to python's default `int` and `float` types.

eva/executor/executor_utils.py
```diff
@@ -18,6 +18,10 @@
 from eva.models.storage.batch import Batch
 
 
+class ExecutorError(Exception):
```
xzdandy, line 21: Where is the `ExecutorError` used?

eva/parser/parser_visitor/_table_sources.py
gaurav274: It is actually quite intuitive, thanks to antlr.
`self.visit(ctx.uidList()) -> visitUidList`

eva/executor/executor_utils.py
gaurav274: Not being used right now. I added it in the hope of using it for reporting error messages during execution.

test/expression/test_function_expression.py
```diff
@@ -57,7 +57,7 @@ def test_should_execute_same_function_if_no_gpu(self, context):
 
         context_instance.gpu_device.return_value = NO_GPU
 
-        expression = FunctionExpression(mock_function, name="test")
+        expression = FunctionExpression(mock_function, name="test", alias="func_expr")
```
xzdandy, line 60: alias is not `str`?

test/integration_tests/test_select_executor.py
```diff
@@ -246,24 +245,120 @@ def test_select_and_sample(self):
...
+                {
+                    "myvideo.id": np.array([0, 0, 1, 1]),
+                    "T.label": np.array(["person", "person", "bicycle", "bicycle"]),
+                }
+            )
+        )
+
+        self.assertEqual(unnest_batch, expected)
+
+    def test_should_raise_error_with_missing_alias_in_lateral_join(self):
```
xzdandy, line 312: So `JOIN LATERAL` much have `AS`. However, it seems from https://github.com/georgia-tech-db/eva/blob/3c16181b140bc3b8cd26e99212d73f14a479ba30/eva/binder/statement_binder.py#L281 , when the node does not have alias, it will use the default node name and column name.

eva/parser/parser_visitor/_table_sources.py
xzdandy: I see it is somehow in the `_insert_statements.py`. I missed it.

test/integration_tests/test_select_executor.py
gaurav274: Yes, I have added the code to handle it. But later felt that avoiding it, for now, is better. We can later remove the check from the parser and test it more.

test/expression/test_function_expression.py
gaurav274: yeah, it should be `Alias`

xzdandy (approved):
LGTM

gaurav274:
#144

PR #310, opened 2022-08-10
feat: Implement Gaussian Blur EVA Transform

LordDarkula:
Add `torchvision.transform` gaussian blur as Transformation UDF

Here is the query to add the UDF to EVA.
```sql
CREATE UDF IF NOT EXISTS GaussianBlur
      INPUT  (Frame_Array_In NDARRAY UINT8(3, ANYDIM, ANYDIM))
      OUTPUT (Frame_Array_Out NDARRAY UINT8(3, ANYDIM, ANYDIM))
      TYPE  Transformation
      IMPL  '<eva-dir>/udfs/gaussian_blur.py';
```


test/udfs/test_utils.py
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+
+import os
+
+import cv2
+import numpy as np
+
+
+def get_udf_data_path():
+    return os.path.join(os.path.dirname(os.path.abspath(__file__)), "data")
```
jarulraj, line 23: Is this path guaranteed to remain unchanged or is it part of the configuration file?

jarulraj: Should we store all data at the same location -- like mnist.mp4 and image files?

test/udfs/test_gaussian_blur.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def test_gaussian_blur_PIL():
+    img_path = os.path.join(get_udf_data_path(), "dog.jpeg")
+    img = PIL.Image.fromarray(load_image(img_path))
+
+    blur = GaussianBlur(kernel_size=3, sigma=0.5)
+
+    blurred_img = blur.transform(img)
+    assert type(img) == type(blurred_img)
```
jarulraj, line 18: Is there a better way to check if the image is blurred?

LordDarkula: I did not want to couple the test to the exact implementation of torch's gaussian blur.
That is not the point.
Instead I wanted to make sure the EVA UDF handles the frame correctly and the output is something we expect.

test/udfs/test_utils.py
LordDarkula: In my opinion, yes.
However it should be a separate PR, which I would be more than happy to create.
Since this PR was about Gaussian blurs I did not want to unnecessarily change unrelated parts of the code base.
I instead pulled this function out because I was using the exact same logic in my own tests.

gaurav274:
LGTM. Please add an end-to-end test case.

jarulraj:
#451

PR #311, opened 2022-08-11
feat: Profile queries

jarulraj:
-- Added a basic timer
-- Need to check timing measurements

### DIFF
--- a/eva/server/command_handler.py
+++ b/eva/server/command_handler.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 import asyncio
 from typing import Iterator, Optional
-from xmlrpc.client import Boolean
 
 from eva.binder.statement_binder import StatementBinder
 from eva.binder.statement_binder_context import StatementBinderContext
@@ -28,7 +27,7 @@ from eva.utils.logging_manager import logger
 from eva.utils.timer import start_timer, end_timer, elapsed_time
 
 
-def execute_query(query, report_time: Boolean = False) -> Iterator[Batch]:
+def execute_query(query, report_time: bool = False) -> Iterator[Batch]:
     """
     Execute the query and return a result generator.
     """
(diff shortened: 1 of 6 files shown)


eva/server/command_handler.py
```diff
@@ -24,20 +25,37 @@
 from eva.optimizer.statement_to_opr_convertor import StatementToPlanConvertor
 from eva.parser.parser import Parser
 from eva.utils.logging_manager import logger
+from eva.utils.timer import start_timer, end_timer, elapsed_time
 
 
-def execute_query(query) -> Iterator[Batch]:
+def execute_query(query, report_time: Boolean = False) -> Iterator[Batch]:
```
gaurav274, line 31: Boolean -> bool

eva/server/command_handler.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 import asyncio
 from typing import Iterator, Optional
+from xmlrpc.client import Boolean
```
gaurav274, line 17: Don't need this

eva/server/command_handler.py
```diff
@@ -63,7 +81,8 @@ def handle_request(transport, request_message):
         logger.warn(e)
         response = Response(status=ResponseStatus.FAIL, batch=None, error=str(e))
     else:
-        response = Response(status=ResponseStatus.SUCCESS, batch=output_batch)
+        response = Response(status=ResponseStatus.SUCCESS,
```
gaurav274, line 84: Looks like you have a linter setting with a different max line length. Black uses 88

test/utils/test_timer.py
```diff
@@ -0,0 +1,52 @@
...
+
+        self.assertTrue(t2 < 10)
+
+    def test_timer_with_query(self):
+
+        CatalogManager().reset()
+        create_sample_video(NUM_FRAMES)
+        load_query = """LOAD FILE 'dummy.avi' INTO MyVideo;"""
+        execute_query_fetch_all(load_query)
+        file_remove('dummy.avi')
```
gaurav274, line 51: There is nothing tested here. Can you verify if the elapsed time attribute is populated in the response and has a reasonable value? Thanks!

eva/server/command_handler.py
jarulraj: Good catch. removed.

test/utils/test_timer.py
jarulraj: Yes, added it in updated test cases.

PR #316, opened 2022-08-12
feat: Predicate Pushdown across join and Fix Materialization Bugs

gaurav274:


gaurav274:
#144

test/util.py
```diff
@@ -217,7 +256,8 @@ def create_sample_video_as_blob(num_frames=NUM_FRAMES):
 
 def copy_sample_videos_to_upload_dir():
     shutil.copyfile(
-        "data/ua_detrac/ua_detrac.mp4", os.path.join(UPLOAD_DIR, "ua_detrac.mp4")
+        "data/ua_detrac/ua_detrac.mp4",
```
jarulraj, line 259: Will these hard-coded paths work on Windows?

jarulraj (approved):

test/util.py
gaurav274: They might not. I’ll fix these. Looks like we have a lot of instances in the this file. Can be done in a separate PR.

PR #322, opened 2022-08-13
Improved Timer 

gaurav274:
Add a context manager based timer class


eva/utils/timer.py
```diff
@@ -17,38 +17,29 @@
...
+    def __enter__(self):
+        assert self._start_time is None, "Concurrent calls are not supported"
+        self._start_time = time.perf_counter()
 
-def end_timer():
-    global _START_TIME
-    global _END_TIME
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        assert self._start_time is not None, "exit called with starting the context"
+        time_elaped = time.perf_counter() - self._start_time
```
jarulraj, line 36: Minor typo: elapsed

eva/utils/timer.py
```diff
@@ -17,38 +17,29 @@
...
-_START_TIME = None
-_END_TIME = None
 
+class Timer:
+    """Class used for logging time metrics.
 
-def start_timer():
-    global _START_TIME
-    global _END_TIME
+    This is not thread safe"""
```
jarulraj, line 24: What are the implications of this -- like when not to use this? We could maybe add that as a comment.

jarulraj:
This is much better 👍 Does not corrupt the CI log with timing log records.

eva/utils/timer.py
gaurav274: There isn't an example in the system right now. I added a comment so that in the future when we bring ray, we are careful.

PR #323, opened 2022-08-13
Fix EVA Configuration

LordDarkula:
I was getting periodic nondeterministic test failures due to how broken the paths are on system setup.

Use pathlib library for calculating paths when starting eva_server

Building paths by appending strings is error-prone and will not work on other platforms. Python pathlib takes
care of that automatically.

Remove Python reference to python `gettempdir()` in bootstrap_environment

On ada-01, I do not have permission to access `/tmp`, so no matter what I do, the unit tests will fail periodically and need to be fixed. Since eva already creates a `~/.eva/` directory we can use `/.eva/tmp` as a temporary directory and clean it up in the cleanup script.

`EVA_UPLOAD_DIR` in `dictionary.py` was also unused previously which caused confusion.

### DIFF
--- a/eva/configuration/bootstrap_environment.py
+++ b/eva/configuration/bootstrap_environment.py
@@ -27,37 +27,38 @@ from eva.configuration.dictionary import (
     EVA_DATASET_DIR,
     EVA_DEFAULT_DIR,
     EVA_INSTALLATION_DIR,
+    EVA_UPLOAD_DIR,
 )
 
 
-def get_base_config():
-    ymlpath = None
+def get_base_config() -> Path:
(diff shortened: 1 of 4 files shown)


jarulraj:
Important change. Is it to possible to quickly check for other hard-coded paths?

LordDarkula:
> Important change. Is it to possible to quickly check for other hard-coded paths?

There are a couple instances of hard-coded paths used in tests, but they are all used with Mocks so changing them is not really necessary. I could not find any other instances of hard-coded paths. Maybe @gaurav274 can help me on this.

jarulraj:
We should mock the local dev environment in a test case (that is why coverage is dropping a bit). Otherwise, LGTM.

PR #341, opened 2022-08-21
Cancel the asyncio tasks in db_api when interruptted.

xzdandy:
This pr fixes when ctrl-c `cursor.fetch_one()`, the ayncio tasks `queue.wait()` needs to be cancelled.

### DIFF
--- a/eva/server/db_api.py
+++ b/eva/server/db_api.py
@@ -17,6 +17,7 @@ import base64
 import os
 import random
 
+from signal import SIGINT, SIGTERM
 from eva.models.server.response import Response
 from eva.server.async_protocol import EvaClient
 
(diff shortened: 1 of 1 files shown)


gaurav274 (approved):

jarulraj:
@xzdandy Will users on MacOS/Windows be able to use EVA (without support for concurrent queries)?

xzdandy:
> @xzdandy Will users on MacOS/Windows be able to use EVA (without support for concurrent queries)?

Unfortunately no. `multiprocessing.Process` will not work with `asycnio` on MacOS/Windows (This is inferred from the doc [contexts-and-start-methods](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods). I haven't tried myself on these OSs.).

To support killing a query execution in the server, we need a "container" for the execution. Right now, we consider that the process fits the best. Other ideas are welcome.

jarulraj:
Working in https://github.com/georgia-tech-db/eva/pull/453

PR #342, opened 2022-08-30
Support for NOT NULL

Anirudh58:
- Added support for NOT NULL 
- Added testcases for partial LOAD (LOAD only some columns for a CSV)

### DIFF
--- a/eva/parser/parser_visitor/_create_statements.py
+++ b/eva/parser/parser_visitor/_create_statements.py
@@ -101,7 +101,7 @@ class CreateTable(evaql_parserVisitor):
                 column_constraint_information.unique = True
                 column_constraint_information.nullable = False
                 not_null_set = True
-            elif return_type == ColumnConstraintEnum.NULLNOTNULL:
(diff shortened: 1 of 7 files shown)


eva/parser/parser_visitor/_create_statements.py
```diff
@@ -94,11 +94,19 @@ def visitColumnDefinition(self, ctx: evaql_parser.ColumnDefinitionContext):
...
 
+        not_null_set = False
         for i in range(constraint_count):
             return_type = self.visit(ctx.columnConstraint(i))
             if return_type == ColumnConstraintEnum.UNIQUE:
-
                 column_constraint_information.unique = True
+                column_constraint_information.nullable = False
+                not_null_set = True
+            elif return_type == ColumnConstraintEnum.NULLNOTNULL:
```
gaurav274, line 104: Can we call it `NOTNULL`?

Anirudh58: Yeah I was thinking the same why it was named this way originally.

gaurav274 (approved):

PR #351, opened 2022-09-09
Add typing to Batch, AbstractExecutor, and AbstractUDF

LordDarkula:
Add typing information
Remove mutable default parameters
Remove unnecessary properties


eva/udfs/pytorch_abstract_udf.py
```diff
@@ -37,17 +37,15 @@ class PytorchAbstractClassifierUDF(AbstractClassifierUDF, nn.Module, GPUCompatib
     def __init__(self):
         AbstractClassifierUDF.__init__(self)
         nn.Module.__init__(self)
+        self.transforms = [transforms.ToTensor()]
 
     def get_device(self):
         return next(self.parameters()).device
 
-    @property
```
gaurav274, line 44: Any reason to remove it? We can rather have a dummy function that each class can overload. Maybe call it preprocess.

LordDarkula: The point of a property is to encapsulate internal class attributes to control their usage by the outside world. Transforms is just a class field that is not used by any external actors. Eventually we should create a preprocess function but that is not within the scope of this PR.

gaurav274 (approved):

PR #355, opened 2022-09-12
Docs: Adding instructions for setup on M1 Mac

adarsh2397:
This PR adds onto the current documentation with instructions for setting up on an M1 Mac by mainly pointing out the changes in the setup instructions for M1.
Existing instructions are left unchanged as they are still working on Linux systems.

Should we directly replace `opencv-python` version to `>= 4.5.1.48` in the `setup.py` file itself? I have verified with the latest version of `opencv-python==4.6.0.66` and all tests are passing so far.

TODO:
- [ ] Instead of **Step 4**, should we directly replace opencv-python version in setup.py to opencv-python>=4.5.1.48 
- [ ] Verify the same installation on an M2 Mac and update the document to specify the same.


jarulraj (approved):

jarulraj:
Bumping opencv to the latest version does not work -- https://app.circleci.com/pipelines/github/georgia-tech-db/eva/1109/workflows/f6976bc8-3642-4d20-803f-96cb0c82d6ee/jobs/1145

adarsh2397:
> Bumping opencv to the latest version does not work -- https://app.circleci.com/pipelines/github/georgia-tech-db/eva/1109/workflows/f6976bc8-3642-4d20-803f-96cb0c82d6ee/jobs/1145

I was looking into the pip installation and this is what I found:
image

Turns out we are installing `opencv-python==4.6.0.66` and easyOCR is installing another version of `opencv-python-headless<=4.5.4.60` and they are somehow not compatible with each other.
One way to resolve this as given by the below link is to just uninstall `opencv-python` and `opencv-python-headless` and then reinstall `opencv-python` which is what worked for me.
https://stackoverflow.com/questions/72343944/cv2-imshow-doesnt-work-when-easyocr-installed

But since the current version of opencv isn't causing any issues for us, we can keep the current versioning for now and not complicate the setup. Incase we face some dependancy issues with `opencv` and some other package in the future, this thread might be good to revisit.

PR #365, opened 2022-09-16
EVA UDF Registration

adarsh2397:
This PR aims to replace the existing UDF registration method to a more simpler method avoiding inheritance and unnecessary function definitions.
See https://github.com/georgia-tech-db/eva/discussions/345 for more details.

Tasks:

- [ ] Create Single UDF Abstract Class (without conflicting with existing AbstractUDF)
- [ ] Implement 3 abstract methods in UDF Class: `setup`, `preprocess` and `forward`
- [ ] Convert existing UDFs to the new format by implementing the above methods (setup, preprocess and forward)
- [ ] Modify `create_udf_executor.py`, `function_expression.py` and any other file that plays a role in UDF registration and execution to use the new UDF Class


eva/udfs/udf.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @property
+    def name(self) -> stt
+
+    def setup(
+        self,
+        setup_func=None,
+        cache=True,
+        requirements=["tqdm"],
+        model_path=["path/to/model"],
```
LordDarkula, line 43: Do not use mutable values as default parameters. Instead use `None` and set it in the method body

eva/udfs/udf.py
```diff
@@ -0,0 +1,66 @@
...
+    ):
+        # TODO: perform the basic setup instructions here
+        pass
+
+    def preprocess(
+        self,
+        preprocess_func=None,
+        input_type=PILImage,
```
LordDarkula, line 51: Create enum with these types

LordDarkula (changes requested):

PR #366, opened 2022-09-16
Replace external usage of configuration.constants

LordDarkula:
Rename `configuration.dictionary` to `configuration.constants` for consistency.

All configuration should be accessed through the `ConfigurationManager` because it is the source of truth. Accessing the constants file directly bypasses the `eva.yml` file and makes configuration very messy.


eva/configuration/bootstrap_environment.py
```diff
@@ -86,12 +87,25 @@ def bootstrap_environment(eva_config_dir: Path, eva_installation_dir: Path):
     logger.setLevel(level)
     logger.debug("Setting logging level to: " + str(level))
 
-    # fill default values for dataset, database and upload loc if not present
+    # fill default values for eva installation dir,
+    # dataset, database and upload loc if not present
+    eva_installation_dir_from_file = read_value_config(
```
gaurav274, line 92: Let us fix the name and document what is happening in this function. Honestly, I've lost track of things, and it is better to write a brief high-level idea of the function (just in case we revisit it later).  Thanks!

LordDarkula: I will have a future PR in which I redo this function completely in preparation for making it concurrent-friendly. I am just taking care of these issues so they do not cause bugs in the future.

gaurav274 (approved):

PR #367, opened 2022-09-16
Dataset support 

xzdandy:
Add dataset support through storage engine, and leave the other parts of the system untouched.

### DIFF
--- a/eva/storage/opencv_storage_engine.py
+++ b/eva/storage/opencv_storage_engine.py
@@ -88,7 +88,7 @@ class OpenCVStorageEngine(AbstractStorageEngine):
             )
             for batch in reader.read():
                 column_name = table.columns[0].name
-                batch.frames[column_name] = video_file_name
+                batch.frames[column_name] = str(video_file_name)
                 yield batch
 
     def _get_video_file_path(self, metadata_file):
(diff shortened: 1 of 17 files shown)


eva/storage/opencv_storage_engine.py
```diff
@@ -57,7 +60,18 @@ def drop(self, table: DataFrameMetadata):
...
 
     def write(self, table: DataFrameMetadata, rows: Batch):
-        pass
+        df = rows.frames
+        try:
+            dir_path = Path(table.file_url)
+            for video_file_path in df["video_file_path"]:
+                video_file = Path(video_file_path)
+                shutil.copy2(str(video_file), str(dir_path))
+                self._create_video_metadata(dir_path, video_file.name)
```
gaurav274, line 69: To be sure, we are only writing the file's name in the metadata and ignoring all the relative paths.

gaurav274 (approved):
LGTM - I'm adding a few testcases and then will add the filtering based on video name optimization.

eva/storage/opencv_storage_engine.py
xzdandy: Yes, relative path is associated with dataset.

PR #368, opened 2022-09-17
feat: Refactoring Notebooks

jarulraj:
-- Refactoring notebooks to use the same notebook (00-start-server.ipynb) for starting the server
-- We need to test all the notebooks as part of CI using this command:

# particular notebook
`pytest --nbmake "./tutorials/03-movie-analysis.ipynb"  -s`

# all notebooks
`pytest --nbmake "./tutorials/"  -s`

cc @kaushikravichandran @gaurav274 @LordDarkula @Aryan-Rajoria


jarulraj:
Another problem to solve is that the folder locations are relative to the "tutorials" sub-folder. CREATE UDF statements also take a relative path. Colab might have a different folder structure (e.g., `/content`). Not sure how to make it work both locally and on Colab.

gaurav274:
What do you think about creating a `get_dataset` function inside the notebook to `wget` all the required data (videos, udfs, etc.) and put it relative to the notebook?

PR #372, opened 2022-09-17
bugfix: Make ConfigurationManager read and update operate on eva.yml

LordDarkula:
`ConfigurationManager` is a singleton that loads the contents of `~/.eva/eva.yml` into memory and uses that to read and write configuration values. The issue with that is if the source file changes, the stale values will continue to be used. This causes confusion if the user wants to change configuration values. When EVA becomes distributed, this system breaks down entirely as different processes will operate with different configuration values.

This PR simply reads directly from `~/.eva/eva.yml` when a read is needed and writes back to the file when `update_value` is called.

### DIFF
--- a/eva/configuration/config_utils.py
+++ b/eva/configuration/config_utils.py
@@ -24,7 +24,15 @@ def read_value_config(config_path: Path, category: str, key: str) -> Any:
         return config_obj[category][key]
 
 
-def update_value_config(cfg: Any, category: str, key: str, value: str):
-    category_data = cfg.get(category, None)
-    if category_data:
-        category_data[key] = value
+def update_value_config(config_path: Path, category: str, key: str, value: str):
+    # read config file
+    with config_path.open("r") as yml_file:
+        config_obj = yaml.load(yml_file)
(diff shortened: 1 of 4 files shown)


eva/configuration/config_utils.py
```diff
@@ -12,14 +12,28 @@
...
+        if config_obj is None:
+            raise ValueError(f"Invalid path to config file {config_path}")
+        return config_obj[category][key]
 
-def update_value_config(cfg: Any, category: str, key: str, value: str):
-    category_data = cfg.get(category, None)
-    if category_data:
-        category_data[key] = value
+
+def update_value_config(config_path: Path, category: str, key: str, value: str):
```
gaurav274, line 29: Do we need to make it atomic?

gaurav274:
Can we remove the singleton pattern?

eva/configuration/config_utils.py
LordDarkula: Yes, but that will be in the next to next PR

gaurav274 (approved):

PR #375, opened 2022-09-19
Update notebook and adding them to api-docs

Aryan-Rajoria:
1) Adding the new notebook names to `api-docs/_toc.yaml`
2) running the google colab notebook for object detection and found that `x1,y1 = bbox[0]` throws error, changed it to `x1,y1,x2,y2 = bbox`


Aryan-Rajoria:
I have tried running the `movie-analysis` notebook, I believe updating the version of `evadb` on pypi, should solve the issue hence leaving the PR here.

gaurav274 (approved):

PR #387, opened 2022-09-22
Notebook update

Aryan-Rajoria:


eva/udfs/emotion_detector.py
```diff
@@ -104,13 +103,13 @@ def __init__(self, threshold=0.85):
 
         # load model
         self.model = VGG("VGG19")
-        output_directory = os.path.join(EVA_DEFAULT_DIR, "udfs", "models")
+        output_directory = os.path.join(os.getcwd(), "models")
```
gaurav274, line 106: We shouldn't need to change this. This should work across platforms

Aryan-Rajoria: Yes, made the change.

gaurav274 (approved):

PR #399, opened 2022-09-25
Convert FrameInfo and Response use to dataclass

LordDarkula:
- use `auto()` to define enum states in `properties.py`
- Convert `FrameInfo` and `Reponse` to immutable dataclasses


xzdandy (approved):
LGTM

PR #401, opened 2022-09-27
feat: sample push down

gaurav274:


eva/readers/opencv_reader.py
```diff
@@ -44,12 +53,26 @@ def _read(self) -> Iterator[Dict]:
...
+                    yield {"id": frame_id, "data": frame}
+                    _, frame = video.read()
+                    frame_id += 1
+        else:
+            for begin, end in range_list:
+
+                # align begin with sampling rate
+                if begin % self._sampling_rate:
+                    begin += self._sampling_rate - (begin % self._sampling_rate)
+                print(begin, end + 1, self._sampling_rate)
```
xzdandy, line 71: Remove print

eva/readers/opencv_reader.py
```diff
@@ -44,12 +53,26 @@ def _read(self) -> Iterator[Dict]:
...
+                while frame is not None and frame_id <= end:
+                    yield {"id": frame_id, "data": frame}
+                    _, frame = video.read()
+                    frame_id += 1
+        else:
+            for begin, end in range_list:
+
+                # align begin with sampling rate
+                if begin % self._sampling_rate:
+                    begin += self._sampling_rate - (begin % self._sampling_rate)
```
xzdandy, line 70: If sampling has higher precedence, should it happen before predicate? Is that due to the fact that we only push down range id predicates, so the order does not matter after we align the `begin`?

eva/readers/abstract_reader.py
```diff
@@ -57,6 +57,7 @@ def read(self) -> Iterator[Batch]:
                 row_size = get_size(data)
             data_batch.append(data)
             if len(data_batch) * row_size >= self.batch_mem_size:
+                print("yielding: ", len(data_batch), self.batch_mem_size, row_size)
```
pchunduri6, line 60: Remove print

eva/readers/opencv_reader.py
gaurav274: Yes. And I also removed the PushFilterBelowSample Rule to enforce it

xzdandy (approved):

PR #415, opened 2022-10-02
Remove AbstractNdarrayUDF

LordDarkula:
- Add `setup` and `forward` abstract methods to `AbstractUDF`
- Convert NDArray UDFs to this format


.coveragerc
```diff
@@ -1,6 +1,7 @@
 [run]
 omit = 
     eva/parser/evaql/*
+    eva/udfs/abstract_udf.py
```
xzdandy, line 4: Why omit the `abstract_udf`?

xzdandy (approved):
Overall LGTM

.coveragerc
LordDarkula: Abstract classes cannot be directly tested so including them in coverage reports is misleading. Adding/removing abstract methods does not actually alter test coverage.

PR #426, opened 2022-10-04
Simplify PytorchAbstractClassifierUDF interface

LordDarkula:
If the user wants to use a pytorch model, they can inherit from `PytorchAbstractClassifierUDF` and implement `setup(*args, **kwargs)` and `forward(frames: Tensor) -> pd.DataFrame`. They can choose to define `self.transforms` in `setup` if they wish to use pytorch transforms.


xzdandy (approved):
LGTM

PR #432, opened 2022-10-10
docs: Adding header to emotion palette

Aryan-Rajoria:


jarulraj:
@kaushikravichandran Any idea on why Jenkins failed?

kaushikravichandran:
> @kaushikravichandran Any idea on why Jenkins failed? Merging is blocked because of that.

Some intermittent test failure, will take a look. Jenkins is not a mandatory check now, so it shouldn't be blocking. I think an approving review is required.

PR #434, opened 2022-10-12
Encapsulate Batch

LordDarkula:
Right now many places in the query execution process access the underlying `DataFrame` in a `Batch`. This tightly couples the code to `Batch` and the pandas library. This code decouples the two and builds a larger API in `Batch` to allow indirect access to `Batch` functionality.


eva/models/storage/batch.py
```diff
@@ -175,6 +218,29 @@ def sort_orderby(self, by, sort_type=None) -> None:
...
+        self._frames = ~self._frames
+
+    def all_true(self) -> bool:
+        return self._frames.all().bool()
+
+    def all_false(self) -> bool:
+        inverted = ~self._frames
+        return inverted.all().bool()
+
+    def create_mask(self) -> List:
```
gaurav274, line 231: What is this?

eva/models/storage/batch.py
```diff
@@ -191,6 +257,9 @@ def project(self, cols: None) -> Batch:
             )
         return Batch(self._frames[verfied_cols], self._identifier_column)
 
+    def repeat(self, times: int) -> None:
```
gaurav274, line 260: Why do we need this?

eva/models/storage/batch.py
```diff
@@ -249,18 +318,90 @@ def concat(cls, batch_list: Iterable[Batch], copy=True) -> Batch:
 
         return Batch(frame)
 
+    @classmethod
+    def inner_join(cls, first: Batch, second: Batch) -> Batch:
```
gaurav274, line 322: Can we make it join and pass `inner` as an argument? We will need `left` join in an upcoming feature.

eva/models/storage/batch.py
```diff
@@ -249,18 +318,90 @@ def concat(cls, batch_list: Iterable[Batch], copy=True) -> Batch:
...
+        if expression == ExpressionType.ARITHMETIC_ADD:
+            return Batch(pd.DataFrame(first._frames + second._frames))
+        elif expression == ExpressionType.ARITHMETIC_SUBTRACT:
+            return Batch(pd.DataFrame(first._frames - second._frames))
+        elif expression == ExpressionType.ARITHMETIC_MULTIPLY:
+            return Batch(pd.DataFrame(first._frames * second._frames))
+        elif expression == ExpressionType.ARITHMETIC_DIVIDE:
+            return Batch(pd.DataFrame(first._frames / second._frames))
+
+    def hash_items(self, indexes) -> None:
```
gaurav274, line 345: Rename it to something reflecting that hash values are assigned to `index`

eva/models/storage/batch.py
```diff
@@ -249,18 +318,90 @@ def concat(cls, batch_list: Iterable[Batch], copy=True) -> Batch:
...
+
     def empty(self):
         """Checks if the batch is empty
         Returns:
             True if the batch_size == 0
         """
         return len(self) == 0
 
+    @classmethod
+    def compare_contains(cls, batch1: Batch, batch2: Batch) -> None:
```
gaurav274, line 368: Move it up with other class methods

gaurav274 (changes requested):

gaurav274:
Great job with the rewrite! I'll take another closer look. 
Please also profile the performance of queries with the new changes. Doing it for integration test cases should be sufficient. I want to ensure we don't add unnecessary copy operations. Thanks!

eva/models/storage/batch.py
LordDarkula: It is used when evaluating logical expressions.

xzdandy:
Very nice work! Thanks for the efforts.

jarulraj (approved):

PR #441, opened 2022-10-20
feat: Enable Support for Querying UDF through Logical Type

adarsh2397:
Tagging Discussion: [Support for Querying using UDF Type](https://github.com/georgia-tech-db/eva/discussions/440)

Changes Made:
- Added a file to store the Enum of different UDF Logical Types
- Added methods to get UDF by type in `catalog_manager.py` and `udf_service.py`
- Modified a existing UDF `FastRCNNObjectDetector` and a test `test_show_info_executor`

Pending Tasks:

- [ ] Should we add a check in `CREATE UDF` statement to check if inputs and outputs match?
- [ ] Add test cases to query using the Logical UDF Type instead of the name

### DIFF
--- a/eva/binder/statement_binder.py
+++ b/eva/binder/statement_binder.py
@@ -37,6 +37,7 @@ from eva.parser.statement import AbstractStatement
 from eva.parser.table_ref import TableRef
 from eva.parser.types import FileFormatType
 from eva.parser.upload_statement import UploadStatement
+from eva.udfs.udf_types import ModelType
 from eva.utils.generic_utils import path_to_class
 from eva.utils.logging_manager import logger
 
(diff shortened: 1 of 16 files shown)


xzdandy:
Can we add an integration test that uses the Logical UDF?

adarsh2397:
> Can we add an integration test that uses the Logical UDF?

Yep. Working on it. Will update on this thread once I push the changes.

adarsh2397:
Added integration tests for using UDF in a simple select query and as part of a LATERAL JOIN. The PR is ready to merge. Not sure why the coverage has dropped though. I tried investigating, but could not find the issue.

jarulraj:
@IshSiva @kaushikravichandran  @gaurav274 Do we need this PR given the recent changes in decorators?

PR #442, opened 2022-10-24
Alias mapping for group by operator

IshSiva:


eva/optimizer/column_deriver.py
```diff
@@ -0,0 +1,119 @@
...
+
+#Lateral Join: Children traverse sequential and join.
+#Seq Scan: 
+
+
+import sys
+from eva.planner import SeqScanPlan, LateralJoinPlan, StoragePlan, FunctionScanPlan, ProjectPlan
+
+
+if sys.version_info >= (3, 8):
```
gaurav274, line 16: Let us move it to utils.py and reuse across repo

eva/optimizer/column_deriver.py
```diff
@@ -0,0 +1,119 @@
...
+            return dispatcher.dispatch(args[1].__class__)(*args, **kw)
+
+        wrapper.register = dispatcher.register
+        update_wrapper(wrapper, func)
+        return wrapper
+
+
+
+
+class ColumnDeriver:
```
gaurav274, line 35: The logic looks good. However, you need to use better function names and variable names. It is hard to understand what is going on. Thanks!

gaurav274 (changes requested):

PR #444, opened 2022-10-26
feat: Explain command support

jiashenC:
Add `EXPLAIN` support in the system.


jiashenC:
Can you @gaurav274 give a proofread of current implementation from parser to statement before I go to the optimizer? As we discussed, the explainable statement/operator/plan will always be embedded as an attribute in the current design.

eva/binder/statement_binder.py
```diff
@@ -75,6 +76,10 @@ def _bind_abstract_expr(self, node: AbstractExpression):
         for child in node.children:
             self.bind(child)
 
+    @bind.register(ExplainStatement)
+    def _bind_explain_statement(self, node: ExplainStatement):
+        self.bind(node.explainable_stmt)
```
xzdandy, line 81: Right now, bind will create catalog entries. @gaurav274 This can be a consideration to move catalog operations to executors.

xzdandy:
> Can you @gaurav274 give a proofread of current implementation from parser to statement before I go to the optimizer? As we discussed, the explainable statement/operator/plan will always be embedded as an attribute in the current design.

I think the idea works

gaurav274 (approved):

PR #445, opened 2022-10-27
Replace Spark+Petastorm with Sqlite+SqlAlchemy

gaurav274:


eva/storage/sqlite_storage_engine.py
```diff
@@ -0,0 +1,148 @@
...
+        result = self._sql_engine.execute(new_table.select())
+        data_batch = []
+        row_size = None
+        for row in result:
+            # Todo: Verfiy the order of columns in row matches the table.columns
+            # ignore the first dummy (_row_id) primary column
+            data_batch.append(self._sql_row_to_dict(row[1:], table.columns))
+            if row_size is None:
+                row_size = 0
+                row_size = get_size(data_batch)
```
jarulraj, line 143: get_row_size. Should this be cached in the schema?

eva/storage/sqlite_storage_engine.py
```diff
@@ -0,0 +1,148 @@
...
+            if len(data_batch) * row_size >= batch_mem_size:
+                yield Batch(pd.DataFrame(data_batch))
+                data_batch = []
+        if data_batch:
```
jarulraj, line 147: We should add a pretty print function for a Batch as well

eva/catalog/schema_utils.py
```diff
@@ -12,89 +12,47 @@
...
         elif column_type == ColumnType.NDARRAY:
-            np_type = NdArrayType.to_numpy_type(column_array_type)
-            petastorm_column = UnischemaField(
-                column_name,
-                np_type,
-                column_array_dimensions,
-                NdarrayCodec(),
-                column_is_nullable,
-            )
+            sqlalchemy_column = Column(LargeBinary)
```
jarulraj, line 37: We could add support for DATETIME maybe in a new PR.

jarulraj (approved):

eva/storage/sqlite_storage_engine.py
```diff
@@ -0,0 +1,148 @@
...
+from eva.catalog.sql_config import SQLConfig
+from eva.models.storage.batch import Batch
+from eva.storage.abstract_storage_engine import AbstractStorageEngine
+from eva.utils.generic_utils import PickleSerializer, get_size
+from eva.utils.logging_manager import logger
+
+# Leveraging Dynamic schema in SQLAlchemy
+# https://sparrigan.github.io/sql/sqla/2016/01/03/dynamic-tables.html
+
+Base = declarative_base()
```
gaurav274, line 35: Not being used. Remove it

eva/storage/sqlite_storage_engine.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self):
+        """
+        Grab the existing sql session
+        """
+        self._sql_session = SQLConfig().session
+        self._sql_engine = SQLConfig().engine
+        self._serializer = PickleSerializer()
+
+    def _dict_to_sql_row(self, dict_row: dict, columns: List[DataFrameColumn]):
+        # Serialize numpy data
```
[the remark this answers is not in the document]
gaurav274: We were using `dict` to ingest data to spark. I followed the same terminology. Let me try if we can get away with just values.

eva/storage/sqlite_storage_engine.py
gaurav274: Yes, we should do it future.

eva/storage/sqlite_storage_engine.py
```diff
@@ -0,0 +1,148 @@
...
+        # Deserialize numpy data
+        dict_row = {}
+        for idx, col in enumerate(columns):
+            if col.type == ColumnType.NDARRAY:
+                dict_row[col.name] = self._serializer.deserialize(sql_row[idx])
+            else:
+                dict_row[col.name] = sql_row[idx]
+        return dict_row
+
+    def create(self, table: DataFrameMetadata, **kwargs):
```
xzdandy, line 69: What happen if the table is already existed? This raised an issue that I did not notice before. I was using `if_not_exists` in the [load executor]( https://github.com/georgia-tech-db/eva/blob/b8cfc51259cadc68b184b4ae2ef6da077cdcde6f/eva/executor/load_video_executor.py#L59). In [mat executor](https://github.com/georgia-tech-db/eva/blob/b8cfc51259cadc68b184b4ae2ef6da077cdcde6f/eva/executor/create_mat_view_executor.py#L34), we can use `handle_if_not_exists` because the catalog entry is created in the mat executor. But this is not the case for the load operator, the catalog entry is created before the load executor. I think we shall choose one way for both of them. And update the mat, load design and this `create` (e.g., doc what happens the table exists, remove `**kwargs`, add a call to check whether table exists or can we use read to do it?)

xzdandy (changes requested):

eva/storage/sqlite_storage_engine.py
gaurav274: Can you raise it as an issue? We can fix it later as I want to get rid of Pyspark asap

PR #446, opened 2022-10-27
fix: importing files in /abstract/ dir

kaushikravichandran:
eva_server doesn't run because some modules (particularly "fastrcnn_object_detector.py") import files in the "eva/udfs/abstract/" dir.
"__init__.py" file (in the "eva/udfs/abstract/" directory - where this file resides) is required for python to treat this directory as a package


gaurav274:
Will this fix the jupyter notebook issue?

gaurav274 (approved):

PR #453, opened 2022-10-31
feat: Support for killing long running queries

jarulraj:
Without relying on multiprocessing


eva/server/server.py
```diff
@@ -79,9 +80,21 @@ def data_received(self, data):
...
+                    self.pending_task = None
             else:
                 logger.debug("Handle request")
-                asyncio.create_task(handle_request(self.transport, request_message))
+                # self.pending_task = multiprocessing.Process(
+                #    target=handle_request,
+                #    args=(self.transport, request_message),
+                # )
+                # self.pending_task.start()
+                self.pending_task = asyncio.create_task(
```
xzdandy, line 95: This will not work. The reason is because the [handle_request](https://github.com/georgia-tech-db/eva/blob/c832a21359d0515d7a5346bdc03740545f7bd65a/eva/server/command_handler.py#L58) is a coroutine generated by `@asyncio.coroutine` instead of a native asyncio function. So the `asyncio protocol` at the server do not have a time point to even receive the `interrupt` message from the client, because once the query execution starts, the internal is a sync function (no asyncio at all). That is the reason we need multiprocess processing, or we need to change the whole eva processing pipeline to use asyncio io, so other asyncio functions (i.e., the asyncio network protocol) can kick in when a query is being processed.  The easy way to see this problem is to add a long delay (e.g., time.sleep, while True loop) in handle_request to fake a long running query, and we won't be able to kill the query and start the another one by ctrl-c from the client.

xzdandy (changes requested):

eva/server/server.py
jarulraj: I see. I added a sleep in execute_query_fetch_all().

jarulraj: Moved to multiprocessing.

kaushikravichandran:
Does circleCI have a limit on time to execute an ipynb notebook? Looks like that's the reason for the test failure on CircleCI

For future reference:c
CircleCI: https://app.circleci.com/pipelines/github/georgia-tech-db/eva/1570/workflows/8121e9bb-8ea5-4f95-a510-fe42743b06c3/jobs/1607
Jenkins: https://ada-00.cc.gatech.edu/jenkins/job/EVA%20Jenkins%20CI/35/

PR #469, opened 2022-11-02
wip: Allow UDF Decorators to create AbstractUDF

LordDarkula:
- Add `UDFService` class that allows user to create a UDF with decorators and generates a subcless of `AbstractUDF`
- Rename `Array_Count` to `ArrayCount`
- Add pickle file read capability to `path_to_class`

TODO: Add command for decorator UDF registration


eva/udfs/udf_service.py
```diff
@@ -0,0 +1,81 @@
...
+        def wrapper_forward(self, frames: ArrayLike):
+            return func(frames)
+
+        self._forward = wrapper_forward
+        return wrapper_forward
+
+    def __call__(self, *args: Any, **kwds: Any) -> Any:
+        return self._forward(args[0])
+
+    def create_udf(self) -> Type[AbstractUDF]:
```
gaurav274, line 60: Are we getting rid of create UDF command?

LordDarkula: `create_udf` outputs an in-memory UDF class that can be instantiated and used. Right now it is the only way of interacting with a UDF created with decorators. Maybe in the future when I figure out registration this method will become unnecessary.

gaurav274: Wait, I'm confused. Are we replacing the current CREATE UDF command with some new command?

test/integration_tests/test_udf_decorator_executor.py
```diff
@@ -0,0 +1,83 @@
...
+    CatalogManager().reset()
+    create_sample_video(NUM_FRAMES)
+    load_query = """LOAD FILE 'dummy.avi' INTO MyVideo;"""
+    execute_query_fetch_all(load_query)
+    yield
+    file_remove("dummy.avi")
+
+
+@pytest.fixture(scope="session")
+def dummy_object_detector():
```
gaurav274, line 40: What is going on in this test case?

eva/udfs/udf_service.py
LordDarkula: Does the updated name and docstring clear up confusion? This method has nothing to do with the `CREATE UDF` SQL command.

eva/udfs/udf_service.py
```diff
@@ -0,0 +1,81 @@
...
+import functools
+from enum import Enum, auto
+from typing import Any, Callable, List, Optional, Type
+
+from numpy.typing import ArrayLike
+
+from eva.udfs.abstract.abstract_udf import AbstractClassifierUDF, AbstractUDF
+
+
+class FrameType(Enum):
```
gaurav274, line 25: Is there a plan to use this Enum?

eva/udfs/udf_service.py
gaurav274: Which docstring are you referring to? 
And in my understanding, `create_udf` should not be exposed to the user. They should only specify forward and setup functions and we should be able to create the UDF from it.

jarulraj:
@LordDarkula @IshSiva Is this PR still needed with the recent changes? (like #607)

IshSiva:
I think this can be closed as we have changed the implementation of decorators.

PR #472, opened 2022-11-03
Replace Json with Pickle Serialization

gaurav274:
1. Replace JSON serialization with pickle serialization.
2. Remove unnecessary logging, which was the bottleneck.
3. Reduce the `display.pprint_nest_depth` of ndarray objects in batch


eva/models/storage/batch.py
```diff
@@ -75,18 +58,14 @@ def columns(self):
...
 
-    def to_json(self):
-        obj = {
-            "frames": self._frames,
-            "batch_size": len(self),
-            "identifier_column": self._identifier_column,
-        }
-        return json.dumps(obj, cls=BatchEncoder)
+    def serialize(self):
+        obj = {"frames": self._frames, "batch_size": len(self)}
```
xzdandy, line 62: What is the usage for `batch_size`? Can we `return PickleSerializer.serialize(self)`?

eva/server/networking_utils.py
```diff
@@ -79,3 +81,10 @@ def set_socket_io_timeouts(transport, seconds, useconds=0):
         return True
     except OSError:
         return False
+
+
+def serialize_message(message: str):
```
xzdandy, line 86: Shall we use `typing.Any` for the type hint of `message`?

xzdandy (approved):
Please check the comments. Overall LGTM

eva/models/storage/batch.py
gaurav274: For cases where we omit some rows, at least batch_size tells the row count

eva/server/networking_utils.py
gaurav274: Done

PR #474, opened 2022-11-04
Support for action queries in Eva

pchunduri6:
1. Added GROUP BY operator for video inputs - GROUP BY "16f" stacks 16 frames into a segment.
2. Added Aggregate Expressions:
    * FIRST - Return the first element in the group
    * LAST - Return the last element in the group
    * SEGMENT - Stack 3d frame inputs into 4d.
3. Added the MVIT (multi-vision transformer-based) ActionRecognition UDF with SOTA action classification accuracy.

Queries supported are of the form:
```
SELECT FIRST(id), MVITActionRecognition(SEGMENT(data)) FROM Actions GROUP BY '16f';
```


eva/executor/groupby_executor.py
```diff
@@ -0,0 +1,63 @@
...
+import pandas as pd
+
+from eva.executor.abstract_executor import AbstractExecutor
+from eva.models.storage.batch import Batch
+from eva.planner.groupby_plan import GroupByPlan
+
+
+class GroupByExecutor(AbstractExecutor):
+    """
+    Samples uniformly from the rows.
```
LordDarkula, line 27: Put an example of groupby in docstring

eva/executor/groupby_executor.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Arguments:
```
LordDarkula, line 29: Move to init docstring

eva/executor/groupby_executor.py
```diff
@@ -0,0 +1,63 @@
...
+            match_string = pattern.group(0)
+            if match_string[-1] == "f":
+                self._segment_length = int(match_string[:-1])
+                # TODO ACTION condition on segment length?
+            else:
+                err_msg = "Only grouping by frames (f) is supported"
+                raise ValueError(err_msg)
+        else:
+            err_msg = "Incorrect GROUP BY pattern: {}".format(node.groupby_clause.value)
+            raise ValueError(err_msg)
```
LordDarkula, line 49: handle error case first, ie

```python
if not pattern:
    raise ValueError()

# continue logic

```

eva/models/storage/batch.py
```diff
@@ -349,6 +349,23 @@ def concat(cls, batch_list: Iterable[Batch], copy=True) -> Batch:
 
         return Batch(frame)
 
+    @classmethod
+    def stack(cls, batch: Batch, copy=True) -> Batch:
```
LordDarkula, line 353: `Batch` already contains `merge_column_wise`, `__add__`, `concat`, `join`, and `combine_batches`. I am not sure if `stack` reproduces the functionality of any of these functions but either way this is a problem. @gaurav274

eva/optimizer/rules/rules.py
```diff
@@ -610,6 +612,25 @@ def apply(self, before: LogicalUnion, context: OptimizerContext):
...
+        pattern.append_child(Pattern(OperatorType.DUMMY))
+        super().__init__(RuleType.LOGICAL_GROUPBY_TO_PHYSICAL, pattern)
+
+    def promise(self):
+        return Promise.LOGICAL_GROUPBY_TO_PHYSICAL
+
+    def check(self, before: Operator, context: OptimizerContext):
+        return True
+
+    def apply(self, before: LogicalGroupBy, context: OptimizerContext):
```
LordDarkula, line 627: unclear what is happening. A brief explanation would be nice regarding the intent would be nice

LordDarkula (dismissed):
Left a couple comments. Good for the most part.

eva/models/storage/batch.py
pchunduri6: What problem are you referring to?

eva/executor/groupby_executor.py
```diff
@@ -0,0 +1,63 @@
...
+    def exec(self) -> Iterator[Batch]:
+        child_executor = self.children[0]
+
+        buffer = Batch(pd.DataFrame())
+        for batch in child_executor.exec():
+            new_batch = buffer + batch
+            while len(new_batch) >= self._segment_length:
+                yield new_batch[: self._segment_length]
+                new_batch = new_batch[self._segment_length :]
+            buffer = new_batch
```
gaurav274, line 63: Missed yielding the buffer outside the loop

pchunduri6: Do you mean for the remainder segment? We discard the final segment (of size < segment_length) because it cannot be processed by the UDF. The assumption is that all the segments are of equal size, and `GROUP BY '8f'` is a hard constraint (standard practice in action localization).

pchunduri6: Done.

eva/models/storage/batch.py
pchunduri6: Stack is different from the functions that already exist. We're adding a new dimension to the input arrays, not concatenating batches. A batch of size 8 will become a batch of size 1 with `stack`.

eva/optimizer/operators.py
```diff
@@ -291,6 +292,25 @@ def __hash__(self) -> int:
...
+        super().__init__(OperatorType.LOGICALGROUPBY, children)
+        self._groupby_clause = groupby_clause
+
+    @property
+    def groupby_clause(self):
+        return self._groupby_clause
+
+    def __eq__(self, other):
+        is_subtree_equal = super().__eq__(other)
+        if not isinstance(other, LogicalGroupBy):
```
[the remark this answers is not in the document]
pchunduri6: Follows rest of the operators for consistency.

eva/executor/groupby_executor.py
pchunduri6: Done.

eva/executor/groupby_executor.py
pchunduri6: Followed other executors for consistency.

eva/optimizer/rules/rules.py
pchunduri6: Consistent with other optimizer rules.

jarulraj (approved):

PR #476, opened 2022-11-06
feat: GPU jenkins support

gaurav274:


docker/jenkins.Dockerfile
```diff
@@ -0,0 +1,49 @@
+# https://www.jenkins.io/doc/book/installing/docker/ - Follow these steps for starting jenkins.
+FROM nvidia/cuda:11.4.2-cudnn8-devel-ubuntu20.04
```
kaushikravichandran, line 2: ada-00 is using 18.04 right?

kaushikravichandran:
Also, please share successful build URL in description. Would be helpful for future reference

kaushikravichandran (approved):

docker/jenkins.Dockerfile
gaurav274: It is the docker image so it does not matter.

pchunduri6: Any reason why you're using cuda 11.4?

gaurav274: Nope. Didn't try others. The current GPU on ada-00 works with 11.4 so I went with it.

pchunduri6: The latest torchvision works best with cuda 11.6, I'm not entirely sure it works with 11.4. Can we try using 11.6 here for better compatibility?

gaurav274: Done!

PR #482, opened 2022-11-11
Benchmarking config

aavhad1910:


script/test/test_benchmark.sh
```diff
@@ -0,0 +1,30 @@
...
+# Run black, isort, linter 
+sh script/formatting/pre-push.sh
+return_code=$?
+if [ $return_code -ne 0 ];
+then
+    exit $return_code
+fi
+
+# Run only benchmark tests
+PYTHONPATH=./ pytest test/ --cov-report term --cov-config=.coveragerc --cov=eva/ -s -v --log-level=WARNING ${1:-} -m "benchmark"
```
gaurav274, line 20: Is it fine if we only run the benchmark command here? We can avoid doing linting etc

test/integration_tests/conftest.py
```diff
@@ -0,0 +1,36 @@
+# coding=utf-8
```
gaurav274, line 1: We can create a new folder `benchmark_tests`

aavhad1910: Sure

script/test/test_benchmark.sh
aavhad1910: I think its better for the linting check to be there, in case we need to run these tests separately. In that case better to have linting in there

Jenkinsfile_weekly
```diff
@@ -0,0 +1,69 @@
...
+    stage('Coverage Check') {
+    
+      steps {
+        sh '''. env37/bin/activate
+          coveralls'''
+      }
+    }
+  }
+}
+x
```
gaurav274, line 69: What is x?

script/test/test_benchmark.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+# This script should return a non-zero value if either
```
gaurav274, line 3: Please add relevant comments

Jenkinsfile_weekly
gaurav274: Btw, why do we need a separate Jenkinsfile? Can we add it to the same file?

aavhad1910: I think it would be better to have a separate file as there an additional verification step in here.

gaurav274 (approved):

PR #485, opened 2022-11-12
feat: Index creation

jiashenC:


eva/parser/evaql/evaql_parser.g4
```diff
@@ -55,7 +55,7 @@ createDatabase
 createIndex
     : CREATE
       INDEX uid indexType?
```
gaurav274, line 57: How about if we move `USING HNSW` to the end? 
`CREATE INDEX name ON tableName(feature) USING HNSW`?

eva/executor/create_index_executor.py
```diff
@@ -0,0 +1,160 @@
...
+        df_metadata = CatalogManager().get_dataset_metadata(
+            table_ref.table.database_name, table_ref.table.table_name
+        )
+
+        # Add features to index.
+        # TODO: batch size is hardcoded for now.
+        index = None
+        input_dim, feat_size = -1, 0
+        for batch in StorageEngine.read(df_metadata, 1):
+            feat = batch.column_as_numpy_array(self.node.col_list[0].name)[0]
```
gaurav274, line 58: What is going on here?

eva/executor/create_index_executor.py
```diff
@@ -0,0 +1,160 @@
...
+                index = self._create_index(self.node.index_type, input_dim)
+            index.add(feat)
+            feat_size += 1
+
+        # Input dimension is inferred from the actual feature.
+        input_index_io = CatalogManager().index_io(
+            "input_feature",
+            ColumnType.NDARRAY,
+            NdArrayType.FLOAT32,
+            [2],
```
gaurav274, line 70: What does 2 stand for?

eva/executor/create_index_executor.py
```diff
@@ -0,0 +1,160 @@
...
+            True,
+        )
+
+        # Output dimension can be hardcoded.
+        id_index_io = CatalogManager().index_io(
+            "logical_id",
+            ColumnType.NDARRAY,
+            NdArrayType.INT64,
+            [2],
```
gaurav274, line 79: Similar as above

eva/executor/create_index_executor.py
```diff
@@ -0,0 +1,160 @@
...
+            ),
+            col_metadata,
+            identifier_column="logical_id",
+            is_video=False,
+        )
+        StorageEngine.create(tb_metadata)
+
+        # Write exact mapping for now.
+        logical_id = [i for i in range(feat_size)]
+        secondary_index = Batch(
```
gaurav274, line 132: Why is it identity mapping?

gaurav274 (commented):
Dropped a few comments. I will do another review later.

eva/executor/create_index_executor.py
jiashenC: 2D dimension.

eva/executor/create_index_executor.py
jiashenC: 1. Get the feature column name.
2. Read the feature vector for that column.

eva/executor/create_index_executor.py
jiashenC: Logical ID is created according to the insertion order. We are currently building index row by row, so it is identity mapping for now. It may change if there is deletion or index update.

eva/executor/create_index_executor.py
jiashenC: Maybe I misunderstand this by looking at UDF test cases. Do we just specify the dimension of input here or we specify the exact input shape?

eva/executor/create_index_executor.py
gaurav274: See if you can use this function. 
https://github.com/georgia-tech-db/eva/blob/c35a1990d4716ca0d7cba6d6b857ded6fec6e656/eva/models/storage/batch.py#L241

eva/executor/create_index_executor.py
gaurav274: We store the shape

eva/executor/create_index_executor.py
gaurav274: I fear that the rows read from SQLite are not guaranteed to be ordered. So shouldn't we use some column from the table?

gaurav274:
Do you think we should also add the use case of the index with this PR? Like how will we use it in queries? Or do you think it should be a separate PR?

xzdandy (commented):
How are we supposed to use this index? I am trying to understand the usage of index_io catalog table.

jiashenC:
@gaurav274 I feel it would be better that we demonstrate how to use index in a separate PR. 

@xzdandy This index is used to search for similar bounding box in a video based on a given input feature. So, it comes with a index_io, which specifies the expected dimension of the input feature.

eva/executor/create_index_executor.py
jiashenC: I think the code stays pretty much the same if I use project. I did rework the code piece and add some comments. Pls check if that improves the readability.

xzdandy:
> @gaurav274 I feel it would be better that we demonstrate how to use index in a separate PR.
> 
> @xzdandy This index is used to search for similar bounding box in a video based on a given input feature. So, it comes with a index_io, which specifies the expected dimension of the input feature.

@jiashenC If my understanding is correct, the index_io stores the column that the index is built on. For example, in the integration tests, that will be the feat column. I have two questions regarding this.
1. Can we use the existing df_column? The index_io looks very similar. We probably can store a foreign key to the df_column.
2. What if the bbox is the output of a detection UDF (the intended use of the index)? Then there will no "actual" column in the system.

jiashenC:
> the index_io stores the column that the index is built on

@xzdandy I am not so sure about this. Is this always the case? I was following udf_io, which seems not specify the which column the udf is built on (correct me if I am wrong). I mainly add index_io to specify the input/output dimensions. 

I think 1 approach also works. I am just not sure whether we will need to expand index_io later, in which df_column is not sufficient.

xzdandy:
> > the index_io stores the column that the index is built on
> 
> @xzdandy I am not so sure about this. Is this always the case? I was following udf_io, which seems not specify the which column the udf is built on (correct me if I am wrong). I mainly add index_io to specify the input/output dimensions.
> 
> I think 1 approach also works. I am just not sure whether we will need to expand index_io later, in which df_column is not sufficient.

The difference here is: say we have two videos V1 and V2. When we create a UDF (say object detector), we specify the input argument of the object detector which is frame (np array). Then UDF works on either V1's frame column or V2's frame column. The UDF is not tied to any videos (more generally tables). 
However, when we create an index for V1's frame column, then the index is tied to the V1's frame column (query on V2 can not use the created index). In other words, we need a relationship between the table's column and the index, so the system can fetch the right index given a query on a table's column. That's why I think we can store the index_io as a foreign key to the df_column. Notice this does not prevent us expanding new metadata in index_io later. 
 
However, the problem of this approach is if you want to build the index on some intermediate results say bbox from the object detector. Then bbox will not be a column in the catalog. On the other hand, it is not clear to me whether and how the current design tackles this situation.  For example, we can not just store bbox as index_io. Instead, we need store the source of the bbox (e.g., the video table and object detector).

eva/executor/create_index_executor.py
jiashenC: Fixed this.

eva/executor/create_index_executor.py
```diff
@@ -0,0 +1,213 @@
...
+            input_index_io = catalog_manager.index_io(
+                "input_feature",
+                ColumnType.NDARRAY,
+                NdArrayType.FLOAT32,
+                [Dimension.ANYDIM, input_dim],
+                True,
+            )
+
+            # Output dimension can be hardcoded.
+            id_index_io = catalog_manager.index_io(
```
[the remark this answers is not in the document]
jiashenC: I think it is better to leave it as ndarray type here, because users can possibly search multiple input feature vectors together or top N matches (instead of top 1).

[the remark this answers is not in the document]
gaurav274: Ah! This corresponds to the output of the index.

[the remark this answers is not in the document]
gaurav274: We should document the input outputs somewhere. The best place would be the catalog file `index.`

eva/catalog/models/index_io.py
```diff
@@ -0,0 +1,157 @@
...
+
+class IndexIO(BaseModel):
+    __tablename__ = "index_column"
+
+    _name = Column("name", String(100))
+    _type = Column("type", Enum(ColumnType), default=Enum)
+    _is_nullable = Column("is_nullable", Boolean, default=False)
+    _array_type = Column("array_type", Enum(NdArrayType), nullable=True)
+    _array_dimensions = Column("array_dimensions", String(100))
+    _is_input = Column("is_input", Boolean, default=True)
```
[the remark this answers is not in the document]
jiashenC: I guess this is clear now. I will add more documentations here.

eva/catalog/models/index_io.py
```diff
@@ -0,0 +1,168 @@
...
+class IndexIO(BaseModel):
+    __tablename__ = "index_column"
+
+    _name = Column("name", String(100))
+    _type = Column("type", Enum(ColumnType), default=Enum)
+    _is_nullable = Column("is_nullable", Boolean, default=False)
+    _array_type = Column("array_type", Enum(NdArrayType), nullable=True)
+    _array_dimensions = Column("array_dimensions", String(100))
+
+    # Faiss index has both inputs and outputs as well. Both inputs and outputs
```
gaurav274, line 35: Better if you move it to the top of the file.  Can you please document it as a schema? I cannot fully understand it. Thanks!

jiashenC: Should I move to catalog index_io function or just top of file?

eva/executor/create_index_executor.py
```diff
@@ -0,0 +1,220 @@
...
+        ]
+        col_metadata = [
+            catalog_manager.create_column_metadata(
+                col.name, col.type, col.array_type, col.dimension, col.cci
+            )
+            for col in col_list
+        ]
+        catalog_manager = CatalogManager()
+        df_metadata = catalog_manager.create_metadata(
+            "secondary_index_{}_{}".format(self.node.index_type, self.node.name),
```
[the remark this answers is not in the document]
jiashenC: Do you mean store it as a foreign key of index?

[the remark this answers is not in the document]
gaurav274: I think he means adding it as a column in the table. I agree with this. This can lead to bugs in the future (lessons learned from load command).

eva/binder/statement_binder.py
```diff
@@ -75,6 +77,19 @@ def _bind_abstract_expr(self, node: AbstractExpression):
...
 
+    @bind.register(CreateIndexStatement)
+    def _bind_create_index_statement(self, node: CreateIndexStatement):
+        self.bind(node.table_ref)
+
+        # TODO: create index currently only supports single numpy column
+        assert len(node.col_list) == 1, "Index cannot be created on more than 1 column"
+
+        col_def = node.col_list[0]
+        table_ref_obj = node.table_ref.table.table_obj
```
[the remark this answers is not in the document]
gaurav274: Sure, let us add an assertion

eva/executor/create_index_executor.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _get_index_io_list(self, input_dim):
+        # Input dimension is inferred from the actual feature.
+        catalog_manager = CatalogManager()
+        input_index_io = catalog_manager.index_io(
+            "input_feature",
+            ColumnType.NDARRAY,
+            NdArrayType.FLOAT32,
+            [Dimension.ANYDIM, input_dim],
+            True,
```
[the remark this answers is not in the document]
jiashenC: After some brainstorming, I think it is better to just store those information in `IndexMetadata` as opposed to `IndexIO`. `IndexIO` can be both input and output, so it does not make sense to store those info for output as well. What do you guys think?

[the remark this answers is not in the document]
gaurav274: Yes, makes sense. Quick question, how do you plan to handle multiple-column cases?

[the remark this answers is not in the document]
jiashenC: Fix this. Ready for another review. Only thing I left out is udf reference. Since this PR assumes the feature vector has already been materialized in a table, we can add that reference in future PRs.

xzdandy (approved):
LGTM

PR #497, opened 2022-11-20
Load syntax change

gaurav274:
1. Change LOAD Command - `LOAD VIDEO "path" INTO MyVideo;`
2. Changed Storage engine to Factory Design Pattern to reduce code coupling.
3. Replace `is_video` with `TableType` to simplify adding new datatypes.


eva/catalog/models/df_metadata.py
```diff
@@ -25,19 +25,19 @@ class DataFrameMetadata(BaseModel):
     _name = Column("name", String(100), unique=True)
     _file_url = Column("file_url", String(100))
     _unique_identifier_column = Column("identifier_column", String(100))
-    _is_video = Column("is_video", Boolean)
+    _table_type = Column("table_type", Integer)
```
xzdandy, line 28: Shall we use the enum type? https://stackoverflow.com/questions/2676133/best-way-to-do-enum-in-sqlalchemy

xzdandy (approved):

eva/executor/storage_executor.py
```diff
@@ -28,15 +30,21 @@ def validate(self):
...
+
+        if self.node.video.table_type == TableType.VIDEO_DATA:
+            return storage_engine.read(
                 self.node.video,
                 self.node.batch_mem_size,
                 predicate=self.node.predicate,
                 sampling_rate=self.node.sampling_rate,
             )
+        elif self.node.video.table_type == TableType.STRUCTURAL_DATA:
+            return storage_engine.read(self.node.video, self.node.batch_mem_size)
```
pchunduri6, line 43: Should we change the `node.video` in `StoragePlan` to something more generic?

eva/parser/load_statement.py
```diff
@@ -44,7 +44,7 @@ def __init__(
         self._file_options = file_options
 
     def __str__(self) -> str:
-        print_str = "LOAD FILE {} INTO {}({}) WITH {}".format(
+        print_str = "LOAD {} INTO {}({}) WITH {}".format(
```
pchunduri6, line 47: 1. Should remove the `WITH` option here?
2. Is `self._file_options` used after this PR?

eva/storage/storage_engine.py
```diff
@@ -12,10 +12,26 @@
...
 
-StorageEngine = str_to_class(ConfigurationManager().get_value("storage", "engine"))()
-VideoStorageEngine = str_to_class(
-    ConfigurationManager().get_value("storage", "video_engine")
-)()
+
+class StorageEngine:
+    storages = {
+        TableType.STRUCTURAL_DATA: str_to_class(
+            ConfigurationManager().get_value("storage", "engine")
```
pchunduri6, line 25: Nit: Would it make sense to change the "engine" here to `structured_data_engine`? I'm assuming that the other modalities will also be added as `audio_engine` etc.

eva/catalog/catalog_type.py
```diff
@@ -19,6 +19,11 @@ class Dimension(IntEnum):
     ANYDIM = -1
 
 
+class TableType(IntEnum):
+    STRUCTURAL_DATA = auto()
```
pchunduri6, line 23: Nit: I think this should be `STRUCTURED_DATA`?

eva/parser/load_statement.py
gaurav274: 1. Sure, I will fix this.
2. I still kept the `file_options`so that in the future if we plan to support more functionality, we can easily extend it.

eva/storage/storage_engine.py
gaurav274: Yes, makes sense. Will do it

PR #503, opened 2022-11-22
test: adding new test case

Aryan-Rajoria:
There are already test cases for most functions in EVA. However, some code which is only run on `import` and is not part of any function can cause error.
like the previous 
`FileNotFoundError: [Errno 2] No such file or directory: 'eva/version.py'` 
which passed the the test_cases because it is not part of any `function` or `class` and caused when running `eva_client` in any directory other than the `eva` repository.


jarulraj (changes requested):
This PR does not seem to increase coverage. If it is testing some previously untested functionality, any idea on why is coverage is not increasing?

/ cc @gaurav274

jarulraj (approved):

PR #507, opened 2022-11-24
Populate row id

jiashenC:


xzdandy (commented):
The pr looks good to me. Please create an issue regarding fixing the video reader to populate row id. 
However, I am a bit confused why currently `SELECT * ` queries ([example](https://github.com/georgia-tech-db/eva/blob/80f428ccc720159677fae28623702e8007c4cab3/test/integration_tests/test_select_executor.py#L108)) in the integration tests work without populating the `_row_id` column for videos.

gaurav274:
This design will require us to handle `_row_id` separately everywhere (different readers and storage engines). But I think we don't have any option. I tried another approach (conditional add `_row_id` when required), but it gets ugly. 

@jiashenC Can we move `_row_id` to a MACRO and let's merge it?

jiashenC:
> The pr looks good to me. Please create an issue regarding fixing the video reader to populate row id. However, I am a bit confused why currently `SELECT * ` queries ([example](https://github.com/georgia-tech-db/eva/blob/80f428ccc720159677fae28623702e8007c4cab3/test/integration_tests/test_select_executor.py#L108)) in the integration tests work without populating the `_row_id` column for videos.

@xzdandy I think for videos, the code always creates an explicit `id` column ([code](https://github.com/georgia-tech-db/eva/blob/80f428ccc720159677fae28623702e8007c4cab3/eva/binder/binder_utils.py#L64)). The implicit `_row_id` is never used.

jiashenC:
#511

jiashenC:
@xzdandy @gaurav274 Created an issue and move `_row_id` to MACRO. Should be ready now.

PR #509, opened 2022-11-25
Move load and upload out of binder.

xzdandy:
Due to the explain command, the catalog operations required by load and upload can not be completed in the binder. Move them to executor instead.


gaurav274 (approved):

eva/catalog/catalog_manager.py
```diff
@@ -116,6 +117,73 @@ def create_metadata(
...
+            name (str): name of the metadata to be added to the catalog
+
+        Returns:
+            DataFrameMetadata:  corresponding metadata for the input table info
+        """
+        columns = [
+            ColumnDefinition(
+                "name", ColumnType.TEXT, None, [], ColConstraintInfo(unique=True)
+            ),
+            ColumnDefinition("id", ColumnType.INTEGER, None, []),
```
jiashenC, line 136: Do we still want to use a explicit `id` column or change to `_row_id` column?

xzdandy: @jiashenC The function is moved out from the binder for clarification (since it is not used in binder anymore). No modifications are made. Regarding the id vs _row_id, I do not have preference, either works.

jiashenC: Sure, just get your opinion if we can get rid of this redundant column. Others LGTM.

xzdandy: Just come to my mind, under the current implementation. id and _row_id are different when a dataset has more than 1 video. id restates from 0/1 for ever video. _row_id probably should not.

PR #513, opened 2022-12-04
feat: Load regex support

gaurav274:
1. Add regex support to load video commands.
2. We roll back in case a load fails in between. This helps avoid an ugly table state. 
3. Add support to persist the loaded file names in RDBMS (SQLite DB)
4. Extensive error handling to prevent loading duplicate file paths, corrupt files, invalid file paths, etc.
5. Current implementation fails on the first error. In the future, we can add an optional flag to let the user control what to do in such cases, skip is one option supported in other DBs.


eva/executor/rename_executor.py
```diff
@@ -29,4 +31,12 @@ def exec(self):
 
         Calls the catalog to modified metadata corresponding to the table.
         """
-        CatalogManager().rename_table(self.node.new_name, self.node.old_table.table)
+        try:
+            obj = self.node.old_table.table.table_obj
+            print(obj)
```
jiashenC, line 36: Extra print here.

eva/parser/drop_statement.py
```diff
@@ -15,32 +15,28 @@
...
 
-    Attributes:
-        TableRef: table reference in the drop table statement
-    """
-
-    def __init__(self, table_refs: List[TableRef], if_exists: bool):
+    def __init__(self, table_infos: List[TableInfo], if_exists: bool):
         super().__init__(StatementType.DROP)
-        self._table_refs = table_refs
+        self._table_infos = table_infos
```
jiashenC, line 27: Just trying to understand the difference here. Why is this changed to `TableInfo` from `TableRef`?

eva/executor/rename_executor.py
gaurav274: Sure, will remove it

gaurav274: Sure, will remove it

eva/parser/drop_statement.py
gaurav274: The idea is that `drop` will always be on a table, so it is unnecessary to use the abstraction `TableRef` which can be a join, subquery etc

eva/storage/opencv_storage_engine.py
```diff
@@ -50,28 +65,68 @@ def create(self, table: DataFrameMetadata, if_not_exists=True):
...
+            )
+
+        except CatalogError:
+            raise
+        except FileExistsError as e:
+            logger.exception(str(e))
+            raise RuntimeError(str(e))
+        except Exception as e:
+            logger.exception(str(e))
+            raise RuntimeError(str(e))
```
xzdandy, line 126: Duplicate error handler?

xzdandy (approved):

PR #514, opened 2022-12-05
Add support open command

jiashenC:
Currently support open jpeg image and return a numpy array for it.


jiashenC:
> How will we use it in a query?

@gaurav274 I think it can be just embedded in any statement that needs to get a numpy array from the source image (as we discussed here https://github.com/georgia-tech-db/eva/discussions/478). 

`SELECT id, data FROM table WHERE similarity(OPEN('../path/to/img.jpg'), data, FeatureExtractor)  TOP 5;`

xzdandy (commented):
My assumption is that `'../path/to/img.jpg'` locates on the client side (i.e., not server, otherwise we will use the load command to load the image?)
Than this logic should be at the client library where, for example, we turn this image into a base64 encoded string and the expression at the server side decodes it.

gaurav274:
> My assumption is that `'../path/to/img.jpg'` locates on the client side (i.e., not server, otherwise we will use the load command to load the image?) Than this logic should be at the client library where, for example, we turn this image into a base64 encoded string and the expression at the server side decodes it.

We assume it is on the server side and the user does not want to use an existing file. In theory, we can replace `open` with a `select query` but not for now.

jiashenC:
Decide to implement open as functional expression after discussion.

PR #518, opened 2022-12-07
Animal detection

sashiko-345:
1. AnimalDetector UDF
2. simple cache for UDF FunctionalExpression


jarulraj:
@sashiko-345 Any idea on why the model is buggy: https://github.com/jarulraj/animal-detection

[](https://github.com/jarulraj/animal-detection/blob/main/README_files/README_12_2.png)
README_12_2

jarulraj:
Marking this closed as the model does not seem to working.

PR #519, opened 2022-12-07
Add support for COUNT, SUM, AVG, MIN, MAX 

tylcos:
Adds support for COUNT, SUM, AVG, MIN, MAX aggregate functions.

This allows for queries such as `SELECT  SUM(id), COUNT(label),
        MIN(score), MAX(score),
        AVG(score)
FROM    ObjectDetectionVideos 
JOIN    LATERAL FastRCNNObjectDetector
        (data) 
AS      V(label, bbox, score);`


jarulraj:
Integrated into #551

PR #520, opened 2022-12-07
feat: Support timestamps and querying for timestamps

raw-chicken:
Timestamp Integration (Group 4 Fall 2022)

Originally we had stored the column as a string, however in eva/binder/binder_utils.py, we changed the column definition to hold integers of strings and created a UDF for formatting the timestamp.

1. Include a new column for timestamps in seconds
2. Ways to create UDF to use function to view formatted timestamp

Run following command:
```
CREATE UDF IF NOT EXISTS Timestamp
INPUT (seconds INTEGER)
OUTPUT (timestamp NDARRAY STR(8))
TYPE NdarrayUDF
IMPL "eva/udfs/timestamp.py";
```

To test this function, please follow these steps:

1. LOAD FILE "./data/mnist/mnist.mp4" INTO MNISTVid;
2. SELECT id, Timestamp(seconds) FROM MNISTVid WHERE Timestamp(seconds)="0:00:10"; (Find all frames associated with a certain second)
3. SELECT id, Timestamp(seconds) FROM MNISTVid WHERE Timestamp(seconds)<"0:00:20"; (Get all frames before 20 seconds)

### DIFF
--- a/eva/binder/binder_utils.py
+++ b/eva/binder/binder_utils.py
@@ -54,6 +54,7 @@ def create_video_metadata(name: str) -> DataFrameMetadata:
         ColumnDefinition(
             "data", ColumnType.NDARRAY, NdArrayType.UINT8, [None, None, None]
         ),
+        ColumnDefinition("timestamp", ColumnType.TEXT, None, []),
     ]
     col_metadata = create_column_metadata(columns)
     uri = str(generate_file_path(name))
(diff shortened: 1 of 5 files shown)


jarulraj:
Good PR! Integrated into https://github.com/georgia-tech-db/eva/pull/551

PR #521, opened 2022-12-07
Add Car plate detection// CS 4420 Group 6

Eric-Shang:
For simplicity we put all files under tutorials, please put this model under tutorials folder: https://www.dropbox.com/s/x677jwtae0elm6h/model.pth?dl=0 and name it as model.pth
move data from data copy to data under tutorials, and run car_plate_detection pipeline.ipynb


jarulraj:
Good PR! Subsumed into an application -- https://github.com/georgia-tech-db/eva-application-template.

PR #523, opened 2022-12-07
Implement Open as a new UDF

jiashenC:


eva/udfs/ndarray/open.py
```diff
@@ -0,0 +1,53 @@
...
+import numpy as np
+import pandas as pd
+
+from eva.executor.executor_utils import ExecutorError
+from eva.udfs.abstract.abstract_udf import AbstractUDF
+
+
+class Open(AbstractUDF):
+    # cache data to avoid expensive open files on disk
+    _DATA_CACHE = dict()
```
gaurav274, line 27: Any reason for making it a class member and not an instance member?

gaurav274 (approved):

PR #524, opened 2022-12-10
Audio index

jarulraj:


jarulraj:
@tushar-97 This PR is no longer needed, right?

tushar-97:
Correct. We can close this PR in favour of https://github.com/georgia-tech-db/eva/pull/586

PR #527, opened 2022-12-12
Register open fe

jiashenC:
* Register Open as UDF
* Fix Open. The system passes panda DataFrame as input by default even for str.


test/integration_tests/test_open.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @unittest.skip(
+        "Skip because evaluate condition on multi-dimensional array is ambiguious."
+    )
+    def test_open_should_open_image(self):
+        # Test query runs successfully with Open function call.
+        config = ConfigurationManager()
+        upload_dir_from_config = config.get_value("storage", "upload_dir")
+        img_path = os.path.join(upload_dir_from_config, "dummy.jpg")
+        select_query = """SELECT * FROM testOpenTable WHERE Open("{}");""".format(
```
gaurav274, line 63: You can test it with open in the select list

jiashenC: Sure

test/integration_tests/test_open.py
```diff
@@ -0,0 +1,76 @@
...
+
+        num = batch_res.column_as_numpy_array("testopentable.num")[0]
+        self.assertEqual(1, num)
+        actual_img = batch_res.column_as_numpy_array("open.data")[0]
+        self.assertEqual(np.sum(expected_img), np.sum(actual_img))
+
+        num = batch_res.column_as_numpy_array("testopentable.num")[1]
+        self.assertEqual(2, num)
+        actual_img = batch_res.column_as_numpy_array("open.data")[1]
+        self.assertTrue(np.isnan(actual_img))
```
gaurav274, line 76: Shouldn't the second row also have the same data?

gaurav274: https://github.com/georgia-tech-db/eva/blob/4da3e482728ad157fbd0e9e2509fda1295ea1759/eva/expression/function_expression.py#L89

jiashenC: Actually, during my testing, the second row never triggers another `Open` call. I thought it is by design.

PR #530, opened 2022-12-13
bug: fix errors due to MVIT UDF

gaurav274:


eva/udfs/udf_bootstrap_queries.py
```diff
@@ -127,13 +127,14 @@ def init_builtin_udfs(mode="debug"):
     """
     queries = [
         Fastrcnn_udf_query,
-        Mvit_udf_query,
```
pchunduri6, line 130: I think few test cases are failing due to removing this from the boostrap. Otherwise LGTM.

pchunduri6 (approved):

PR #533, opened 2022-12-14
feat: load img support

gaurav274:


eva/catalog/catalog_manager.py
```diff
@@ -388,50 +429,50 @@ def check_table_exists(self, database_name: str, table_name: str):
     def get_all_udf_entries(self):
         return self._udf_service.get_all_udfs()
 
-    def get_video_metadata_table(
+    def get_media_metainfo_table(
```
jarulraj, line 432: Maybe add a comment on the difference between `metainfo` and `metadata`.

eva/catalog/catalog_type.py
```diff
@@ -22,6 +22,7 @@ class Dimension(IntEnum):
 class TableType(IntEnum):
     STRUCTURED_DATA = auto()
     VIDEO_DATA = auto()
+    IMAGE_DATA = auto()
```
jarulraj, line 26: Do we need a very different approach to store AUDIO_DATA or PDF_DATA?

jarulraj (approved):

eva/catalog/catalog_manager.py
gaurav274: The catalog's current naming convention is ineffective, and I am working on a new pull request to improve it.

eva/catalog/catalog_type.py
gaurav274: It should be similar

PR #535, opened 2022-12-15
Similarity UDF

jiashenC:
Similarity UDF
* Compare distance between feature from open and base feature and return the calculated distance.

### DIFF
--- a/eva/executor/orderby_executor.py
+++ b/eva/executor/orderby_executor.py
@@ -15,12 +15,12 @@
 from typing import Iterator
 
 from eva.executor.abstract_executor import AbstractExecutor
+from eva.executor.executor_utils import ExecutorError
 from eva.expression.function_expression import FunctionExpression
 from eva.expression.tuple_value_expression import TupleValueExpression
 from eva.models.storage.batch import Batch
 from eva.parser.types import ParserOrderBySortType
 from eva.plan_nodes.orderby_plan import OrderByPlan
-from eva.executor.executor_utils import ExecutorError
 
 
 class OrderByExecutor(AbstractExecutor):
--- a/eva/executor/orderby_executor.py
+++ b/eva/executor/orderby_executor.py
@@ -12,6 +12,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import pandas as pd
+
 from typing import Iterator
 
 from eva.executor.abstract_executor import AbstractExecutor
@@ -68,11 +70,15 @@ class OrderByExecutor(AbstractExecutor):
 
         # Column can be a functional expression, so if it
         # is not in columns, it needs to be re-evaluated.
+        merge_batch_list = [aggregated_batch]
         for col in self._columns:
             if col.col_alias not in aggregated_batch.frames:
                 batch = col.evaluate(aggregated_batch)
-                aggregated_batch_list.append(batch)
-        aggregated_batch = Batch.concat(aggregated_batch_list, copy=False)
+                merge_batch_list.append(batch)
+        if len(merge_batch_list) > 1:
+            merge_df_list = [batch.frames for batch in merge_batch_list]
+            aggregated_df = pd.concat(merge_df_list, axis=1, join="inner")
+            aggregated_batch = Batch(aggregated_df)
 
         # sorts the batch
         try:
@@ -80,7 +86,7 @@ class OrderByExecutor(AbstractExecutor):
                 by=self.extract_column_names(),
                 sort_type=self.extract_sort_types(),
             )
-        except KeyError as e:
+        except KeyError:
             # raise ExecutorError(str(e))
             pass
 
--- a/eva/executor/orderby_executor.py
+++ b/eva/executor/orderby_executor.py
@@ -15,6 +15,8 @@
 from typing import Iterator
 
 from eva.executor.abstract_executor import AbstractExecutor
+from eva.expression.function_expression import FunctionExpression
+from eva.expression.tuple_value_expression import TupleValueExpression
 from eva.models.storage.batch import Batch
 from eva.parser.types import ParserOrderBySortType
 from eva.plan_nodes.orderby_plan import OrderByPlan
--- a/eva/executor/orderby_executor.py
+++ b/eva/executor/orderby_executor.py
@@ -18,6 +18,7 @@ from eva.executor.abstract_executor import AbstractExecutor
 from eva.models.storage.batch import Batch
 from eva.parser.types import ParserOrderBySortType
 from eva.plan_nodes.orderby_plan import OrderByPlan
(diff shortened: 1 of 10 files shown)


eva/udfs/ndarray/similarity.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+
+        def _similarity(row: pd.Series) -> float:
+            open_input, base_input, feature_extractor_name = (
+                row.iloc[0],
+                row.iloc[1],
+                row.iloc[2],
+            )
+
+            udf_metadata = CatalogManager().get_udf_by_name(feature_extractor_name)
```
gaurav274, line 47: We should avoid calling the catalog in the UDF. Rather provide it the function expression.

eva/expression/function_expression.py
```diff
@@ -68,6 +68,13 @@ def name(self):
     def output(self):
         return self._output
 
+    @property
+    def col_alias(self):
```
gaurav274, line 72: Are we merging all the column names?

eva/executor/orderby_executor.py
```diff
@@ -65,14 +67,26 @@ def exec(self) -> Iterator[Batch]:
...
         aggregated_batch = Batch.concat(aggregated_batch_list, copy=False)
 
+        # Column can be a functional expression, so if it
+        # is not in columns, it needs to be re-evaluated.
+        merge_batch_list = [aggregated_batch]
+        for col in self._columns:
+            if col.col_alias not in aggregated_batch.frames:
+                batch = col.evaluate(aggregated_batch)
+                merge_batch_list.append(batch)
+        if len(merge_batch_list) > 1:
```
gaurav274, line 77: We have a merge function in batch. Can we reuse it here?

jiashenC: Fix this.

eva/expression/function_expression.py
jiashenC: Yes. Do you have other suggestions? Not sure what is a good way to create the alias for multi-columns together.

gaurav274: This is tricky. We allow FunctionExpression to produce multiple columns, so we cannot assign a single column to it. Ideally, we should check in the binder if the UDF composition is wrong. For now, you can iterate over the list and check for each column

jiashenC: You mean iterate and check inside the orderby executor?

gaurav274: Yes

gaurav274: Yes

eva/udfs/ndarray/similarity.py
```diff
@@ -0,0 +1,53 @@
...
+        and 2. feature vector from base table.
+        """
+
+        def _similarity(row: pd.Series) -> float:
+            open_feat_np, base_feat_np = (
+                row.iloc[0],
+                row.iloc[1],
+            )
+
+            # Transform to 2D.
```
gaurav274, line 44: Why do we need to do this transformation?

jiashenC: Faiss only accepts a 2D array, so I am doing this transformation.

eva/executor/orderby_executor.py
```diff
@@ -65,14 +73,30 @@ def exec(self) -> Iterator[Batch]:
             aggregated_batch_list.append(batch)
         aggregated_batch = Batch.concat(aggregated_batch_list, copy=False)
 
+        # Column can be a functional expression, so if it
+        # is not in columns, it needs to be re-evaluated.
+        merge_batch_list = [aggregated_batch]
+        for col in self._columns:
```
gaurav274, line 79: Can we use the `extract_column_names ` function? Looks like duplicate code

eva/executor/orderby_executor.py
```diff
@@ -65,14 +73,30 @@ def exec(self) -> Iterator[Batch]:
...
+
         # sorts the batch
         try:
             aggregated_batch.sort_orderby(
                 by=self.extract_column_names(),
                 sort_type=self.extract_sort_types(),
             )
         except KeyError:
-            # pass for now
+            # raise ExecutorError(str(e))
```
gaurav274, line 99: Uncomment

eva/udfs/ndarray/similarity.py
gaurav274: Okay, we should think about enforcing it in the binder. If it requires a lot of effort, we can skip it for now but add a comment.

test/integration_tests/test_similarity.py
```diff
@@ -0,0 +1,152 @@
...
+import pandas as pd
+
+from eva.catalog.catalog_manager import CatalogManager
+from eva.configuration.configuration_manager import ConfigurationManager
+from eva.models.storage.batch import Batch
+from eva.server.command_handler import execute_query_fetch_all
+from eva.storage.storage_engine import StorageEngine
+
+
+class SimilarityTests(unittest.TestCase):
```
gaurav274, line 29: We should consider adding a test case using the `FeatureExtractor` UDF.

gaurav274 (approved):
Minor Comments 
LGTM

eva/executor/orderby_executor.py
jiashenC: You mean we want to raise Error here? If so, this test case (https://github.com/georgia-tech-db/eva/blob/90ed6f2ea1d2e7f8b2016c7f9bfc696746338980/test/integration_tests/test_select_executor.py#L61) will fail.

eva/executor/orderby_executor.py
jiashenC: Simplify this.

eva/udfs/ndarray/similarity.py
jiashenC: Ok. In this case, if the shape is not 2D (e.g., the feature vector output from the feature extractor), who will take the responsibility to convert it to 2D? The similarity UDF or something else?

gaurav274: Is it standard practice to convert the feature vector to 2D for similarity?  I was thinking of reporting it as an error to the user, and they will fix it. But if you think it is standard practise, I'm fine with it.

eva/executor/orderby_executor.py
gaurav274: I thought the new code you added will also fix it.

jiashenC: You are right.

test/integration_tests/test_similarity.py
jiashenC: Fix this.

eva/udfs/ndarray/similarity.py
jiashenC: Ok, I will leave a comment for this and we can think about improving this later.

For your approach, if we instead of asking users to take care the transformation, they will probably need another UDF just for reshaping?

PR #536, opened 2022-12-16
feat: lateral function call to linear data flow

gaurav274:
- A simpler and easier-to-understand pipeline
- Reduce need for special case handling in the execution engine
- Simpler implementation for supporting reuse and ray due to the lack of branching.

image


eva/optimizer/rules/rules.py
```diff
@@ -317,6 +319,99 @@ def apply(self, before: LogicalFilter, context: OptimizerContext):
...
+    def check(self, before: LogicalJoin, context: OptimizerContext):
+        if before.join_type == JoinType.LATERAL_JOIN:
+            if before.join_predicate is None and not before.join_project:
+                return True
+        return False
+
+    def apply(self, before: LogicalJoin, context: OptimizerContext):
+        #     LogicalJoin(Lateral)              LogicalApplyAndMerge
+        #     /           \                 ->       |
+        #    A        LogicalFunctionScan            A
```
xzdandy, line 348: Can we directly create the new operator (LogicalApplyAndMerge --- A) from the statement? I can see the current implementation requires few code modification. Does this transformation have any other benefits?

eva/optimizer/rules/rules.py
```diff
@@ -317,6 +319,99 @@ def apply(self, before: LogicalFilter, context: OptimizerContext):
...
+    beneficial, for instance, in order to prevent decoding additional frames beyond
+    those that satisfy the predicate.
+    Eg:
+
+    Filter(id < 10 and func.label = 'car')           Filter(func.label = 'car')
+            |                                                   |
+        ApplyAndMerge(func)                  ->          ApplyAndMerge(func)
+            |                                                   |
+            A                                            Filter(id < 10)
+                                                                |
```
xzdandy, line 373: Does push `label='car'` into the `ApplyAndMerge` help? So the join can work on few rows and the output batch is smaller. We can defer this optimization to later PRs. Just want to confirm whether we should have a rule for that.

eva/optimizer/rules/rules.py
gaurav274: I considered doing it. However, I feel that will require adding a lot more core. Via rules, it is easier to understand the changes, and we can fallback to lateral join whenever needed. What do you think?

eva/optimizer/rules/rules.py
gaurav274: It might not make much of a difference as we aren't doing join anymore. It is more like a merge. However, we can profile and add it later.

eva/optimizer/rules/rules.py
xzdandy: I can not think of any issues right now. We need to make sure that this rule applies first, because I assume all later rewrite rules (such as UDF reuse) will base on the new linear design.

PR #540, opened 2022-12-17
Bump v0.1.3+dev

gaurav274:


.circleci/config.yml
```diff
@@ -35,11 +35,11 @@ jobs:
 
       - restore_cache:
           keys:
-            - v12-model_cache-{{ checksum "setup.py" }}
+            - v13-model_cache-{{ checksum "setup.py" }}
```
jarulraj, line 38: We should do another checksum (maybe of some other file) instead of relying on increasing the version number.

script/test/package.sh
```diff
@@ -4,21 +4,23 @@
 
 function is_server_up () {
     # check if server started
-    grep "serving" eva.txt
+    netstat -an | grep 0.0.0.0:5432
```
jarulraj, line 7: Does this assume that `netstat` is installed?

jarulraj: We could use the same approach in 00-start-eva-server.ipynb

setup.py
```diff
@@ -45,7 +45,7 @@ def read(path, encoding="utf-8"):
     "sqlalchemy-utils>=0.36.6",
     "antlr4-python3-runtime==4.8",
     "lark>=1.0.0",
-    "pyyaml>=5.1,<5.2",
```
jarulraj, line 48: Any idea on why this constraint was there earlier?

jarulraj (approved):

setup.py
gaurav274: No

Aryan-Rajoria (approved):

Issue #542, opened 2022-12-21
Accessing a Table after restarting eva_server leads to error


Aryan-Rajoria:
Running the command
`SELECT * FROM v2`
results in error
```
@status: ResponseStatus.FAIL
@batch: 
 None
@error: '/root/.eva/0.1.3+dev/eva_datasets/b9093a36e5741249f73b2dfd00a7c8e6'
```

Steps to reporduce:
1) Create a table and add some images to it
2) restart `eva_server`
3) will get the above error

gaurav274:
Fixed with #543

PR #543, opened 2022-12-25
fix: server relaunch does not load metadata tables

gaurav274:
(1) metadata tables were not being loaded upon server restart, so we implemented reflection to load them; 
(2) the show tables command was not functioning correctly; 
(3) metadata tables have been designated as system tables and can no longer be accessed or modified by user queries.


gaurav274:
Fixes #542

PR #544, opened 2022-12-26
refactor: catalog improved file and function names

gaurav274:


jarulraj (approved):
We should add a text document to list the Catalog objects with their purpose.

Any reason for this refactoring?

gaurav274:
The original names were really confusing. We were using dataset/metadata/table interchangeably.

eva/catalog/catalog_manager.py
```diff
@@ -81,143 +85,265 @@ def _shutdown_catalog(self):
...
 
-    def create_multimedia_table(self, name: str, format_type: FileFormatType):
-        if format_type is FileFormatType.VIDEO:
-            return self._create_video_table(name)
-        elif format_type is FileFormatType.IMAGE:
-            return self._create_image_table(name)
-        else:
-            raise CatalogError(f"Format Type {format_type} is not supported")
+    def get_table_catalog_entry(
+        self, database_name: str, table_name: str
```
jiashenC, line 131: Can we drop `database_name`for this function?

eva/catalog/models/table_catalog.py
```diff
@@ -20,16 +20,16 @@
 from eva.catalog.models.base_model import BaseModel
 
 
-class DataFrameMetadata(BaseModel):
-    __tablename__ = "df_metadata"
+class TableCatalog(BaseModel):
```
jiashenC, line 23: Just a curious comment about model in general. Why do we add a catalog suffix for all models? 

To me, catalog includes everything (udf, index, etc). Would this naming give a wrong impression that we maintain different catalogs for different models? (Again, just curious, don’t need a fix).

eva/catalog/catalog_manager.py
```diff
@@ -322,191 +384,90 @@ def udf_io(
...
-        Returns:
-            The persisted UdfMetadata object with the id field populated.
-        """
-
-        metadata = self._udf_service.create_udf(name, impl_file_path, type)
-        for udf_io in udf_io_list:
-            udf_io.udf_id = metadata.id
-        self._udf_io_service.add_udf_io(udf_io_list)
-        return metadata
+    def create_multimedia_table_catalog_entry(
```
jiashenC, line 395: Can you elaborate more diff between `create_multimedia_table_catalog_entry` and `create_media_metadata_table_catalog_entry `?

eva/catalog/catalog_manager.py
```diff
@@ -322,191 +384,90 @@ def udf_io(
...
             logger.exception(err)
             raise CatalogError(err)
 
         return obj
 
-    def create_media_metainfo_table(
-        self, input_table: DataFrameMetadata
-    ) -> DataFrameMetadata:
-        """Create a media metainfo table.
+    def create_media_metadata_table_catalog_entry(
```
jiashenC, line 445: This input parameter is the output from `create_multimedia_table_catalog_entry` function?

eva/catalog/models/column_catalog.py
```diff
@@ -23,19 +23,19 @@
...
+class ColumnCatalog(BaseModel):
+    __tablename__ = "column_catalog"
 
     _name = Column("name", String(100))
     _type = Column("type", Enum(ColumnType), default=Enum)
     _is_nullable = Column("is_nullable", Boolean, default=False)
     _array_type = Column("array_type", Enum(NdArrayType), nullable=True)
     _array_dimensions = Column("array_dimensions", String(100))
-    _metadata_id = Column("metadata_id", Integer, ForeignKey("df_metadata._row_id"))
+    _table_id = Column("table_id", Integer, ForeignKey("table_catalog._row_id"))
```
jiashenC, line 34: Can we have a more specific naming here for both `table_id` and `table`? Because this model will be a table itself, I think it will be confusing for newcomers.

eva/catalog/catalog_manager.py
```diff
@@ -81,143 +85,265 @@ def _shutdown_catalog(self):
         logger.info("Shutting catalog")
         drop_db()
 
-    def create_metadata(
+    "Table catalog services"
+
+    def insert_table_catalog_entry(
```
jiashenC, line 90: Diff meaning between `create` and `insert`?

eva/catalog/models/table_catalog.py
gaurav274: I referred Postgres. They also call each model as different catalog. https://www.postgresql.org/docs/current/catalog-pg-class.html.
What do you think? Maybe I'm wrong.

eva/catalog/models/column_catalog.py
gaurav274: Sorry, I cannot understand what you mean.

eva/catalog/catalog_manager.py
gaurav274: I gave it a thought. Maybe we need it in the future if we support database-level distinction. How about I make it an optional argument?

eva/catalog/catalog_manager.py
gaurav274: I feel since we are adding an entry into the catalog table, naming it `insert` is better than `create`.

eva/catalog/catalog_manager.py
gaurav274: Yes. For every media table, we maintain a structured metadata table with a list of filenames.

gaurav274: I should add a readme with these details.

gaurav274: Please feel free to suggest better function names. Thanks!

eva/catalog/models/column_catalog.py
gaurav274: Oh, do you mean to name it `_table_catalog_id`?

eva/catalog/catalog_manager.py
jiashenC: Sure. Sounds good to me.

eva/catalog/models/column_catalog.py
jiashenC: Yes. I think it is a bit confusing which table this field refers to.

eva/catalog/catalog_manager.py
jiashenC: What about `create_multimedia_table_catalog_entry` and `create_multimedia_metadata_table_catalog_entry`?

Also, should the function name be `create_` or `insert_` or `create_and_insert_`? Just want to keep consistent with `insert_table_catalog_entry` name.

eva/catalog/catalog_manager.py
jiashenC: Sure. Just want to make sure we follow the same convention for all catalog APIs.

eva/catalog/models/table_catalog.py
jiashenC: I see. It seems like they use something like `pg_{...}` as their naming convention? I didn't really see they use `catalog` as suffix.

eva/catalog/models/column_catalog.py
gaurav274: Added doc string for clarity

PR #545, opened 2022-12-30
feat: remove sqlalchemy dep from execution

gaurav274:
1. Separation of system logic from sqlalchemy. 
     a) Avoid accidentally modifying the database outside the catalog. 
     b) Useful for multi-threading as sqlalchemy objects cannot be shared across threads.


jiashenC:
Since you mention we want to avoid modification to database outside catalog, is this code (https://github.com/georgia-tech-db/eva/blob/3ca6d435c9b9eba7f33c8d300580b3fad9539c51/eva/storage/sqlite_storage_engine.py#L69) also doing modifications?

gaurav274:
> Since you mention we want to avoid modification to database outside catalog, is this code (
> 
> https://github.com/georgia-tech-db/eva/blob/3ca6d435c9b9eba7f33c8d300580b3fad9539c51/eva/storage/sqlite_storage_engine.py#L69
> 
> ) also doing modifications?

Yup. It is unlikely that we could remove the dependency between the storage engine and the catalog, as they both rely on the same underlying database. My objective with this pull request is to prevent the sharing of sqlalchemy objects across threads when enabling parallelism. Assumption: the storage engine is aware of its actions and parallelism :) What do you think?

jiashenC:
> > Since you mention we want to avoid modification to database outside catalog, is this code (
> > https://github.com/georgia-tech-db/eva/blob/3ca6d435c9b9eba7f33c8d300580b3fad9539c51/eva/storage/sqlite_storage_engine.py#L69
> > 
> > ) also doing modifications?
> 
> Yup. It is unlikely that we could remove the dependency between the storage engine and the catalog, as they both rely on the same underlying database. My objective with this pull request is to prevent the sharing of sqlalchemy objects across threads when enabling parallelism. Assumption: the storage engine is aware of its actions and parallelism :) What do you think?

I see. Will this function still be thread-safe in multi-threaded scenario?

gaurav274:
> 

Yes, sqlalchemy is designed to be thread-safe. The issue we were facing was that we were improperly passing and sharing sqlalchemy objects between threads. During the binding and optimization process, the master thread would populate the catalog-related attributes in the plan nodes. However, when we spawned new threads or processes during execution, we would pass them the same copy of catalog objects, which caused problems with sqlalchemy. The solution is to ensure that each thread maintains its own separate copy of the metadata object. As long as we do this, we should be able to use sqlalchemy safely in a multithreaded environment.

jiashenC (approved):
LGTM

PR #550, opened 2023-01-02
feat: support key-value cache for function expressions

gaurav274:
Support added for caching when the UDF is in the `lateral join`


eva/catalog/services/udf_history_col_service.py
```diff
@@ -0,0 +1,62 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import List
+
+from eva.catalog.models.udf_history_column import UdfHistoryColumn
+from eva.catalog.services.base_service import BaseService
+from eva.utils.logging_manager import logger
+
+
+class UdfHistoryColumnService(BaseService):
```
jiashenC, line 22: Can you elaborate when is this needed?

eva/utils/kv_cache.py
```diff
@@ -0,0 +1,51 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import pickle
+import shutil
+from typing import Any
+
+from diskcache import FanoutCache
```
jiashenC, line 19: Just curious about the choice of this library. Are we trying to optimize writes to disk here?

gaurav274: I was looking for a lightweight caching library, ideally with a low setup cost. `FanoutCache` optimizes for writes.

eva/expression/function_expression.py
```diff
@@ -157,6 +173,65 @@ def _gpu_enabled_function(self):
...
+
+        cols = [obj.name for obj in self.output_objs]
+
+        # 1. check cache
+        # We are required to iterate over the batch row by row and check the cache.
+        # This can hurt performance, as we have to stitch together columns to generate
+        # row tuples. Is there an alternative approach we can take?
+
+        results = np.full([len(batch), len(cols)], None)
+        keys = batch
```
jiashenC, line 201: I am bit confused on the scope of keys here since it will be a 2D structure. How do you keep track of rows and columns in the key?

eva/catalog/catalog_utils.py
```diff
@@ -43,6 +50,15 @@ def get_video_table_column_definitions() -> List[ColumnDefinition]:
     return columns
 
 
+def get_table_primary_columns(table_catalog_obj: TableCatalogEntry):
+    if table_catalog_obj.table_type == TableType.VIDEO_DATA:
+        return get_video_table_column_definitions()[:2]
```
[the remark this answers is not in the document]
gaurav274: That uniquely identifies the row. Should I change the name?

eva/catalog/catalog_utils.py
```diff
@@ -43,6 +50,15 @@ def get_video_table_column_definitions() -> List[ColumnDefinition]:
...
+    elif table_catalog_obj.table_type == TableType.IMAGE_DATA:
+        return get_image_table_column_definitions()[:1]
```
xzdandy, line 57: There is a `get_image_table_primary_columns` below doing the same thing. Maybe remove it?

eva/catalog/catalog_utils.py
```diff
@@ -83,3 +101,49 @@ def xform_column_definitions_to_catalog_entries(
...
+    for expr in func_expr.find_all(FunctionExpression):
+        udf_depends.append(expr.udf_obj.row_id)
+    for expr in func_expr.find_all(TupleValueExpression):
+        col_depends.append(expr.col_object.row_id)
+    cache_name = func_expr.signature()
+
+    # add salt to the cache_name so that we generate unique name
+    path = str(get_str_hash(cache_name + uuid.uuid4().hex))
+    cache_dir = ConfigurationManager().get_value("storage", "cache_dir")
+    cache_path = str(Path(cache_dir) / Path(f"{path}_{func_expr.name}"))
```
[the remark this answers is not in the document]
gaurav274: The `path` includes the args (part of the signature). I appended`func_name` for readability. Are you suggesting args for the same reason? We can do it. was concerned the file_name can get arbitrarily long.

eva/catalog/models/udf_catalog.py
```diff
@@ -46,6 +49,13 @@ class UdfCatalog(BaseModel):
         cascade="all, delete, delete-orphan",
     )
 
+    _dep_caches = relationship(
+        "UdfCacheCatalog",
+        secondary=depend_udf_and_udf_cache,
+        back_populates="_udf_depends",
+        cascade="all, delete",
```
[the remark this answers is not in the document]
gaurav274: Yup, that makes sense. Will do it. Ideally, we should do it on the child table side.

eva/executor/create_index_executor.py
```diff
@@ -138,8 +138,9 @@ def _create_faiss_index(self):
         except Exception as e:
             # Roll back in reverse order.
             # Delete on-disk index.
-            if os.path.exists(self._get_index_save_path()):
-                os.remove(self._get_index_save_path())
+            index_path = Path(self._get_index_save_path())
+            if index_path.exists():
```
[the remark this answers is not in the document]
gaurav274: No benefit. We are slowly moving towards using `pathlib` everywhere.

eva/expression/function_expression.py
gaurav274: I do not understand your concern. what do you mean by rows and columns in the key? 
For clarity, here keys refer to the key in the cache table. 
In some cases, it is directly the arguments of the UDF otherwise it is a logically equivalent version of the keys.

eva/optimizer/optimizer_utils.py
```diff
@@ -180,6 +187,73 @@ def extract_pushdown_predicate_for_alias(
     )
 
 
+def optimize_cache_key(expr: FunctionExpression):
```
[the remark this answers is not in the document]
gaurav274: `For instance, frame data can be replaced with frame id.` I leverage the primary keys of the `VIDEO` table to replace `data` attribute. The idea is to simplify the key we use to hash in the key-value store.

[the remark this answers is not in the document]
gaurav274: Right now, I'm doing it for just the data column.

eva/expression/tuple_value_expression.py
```diff
@@ -89,10 +93,12 @@ def signature(self):
...
         ), f"Unsupported type of self.col_object {type(self.col_object)}, expected ColumnCatalogEntry or UdfIOCatalogEntry"
 
+        col_name = self.col_object.name
+        row_id = self.col_object.row_id
         if isinstance(self.col_object, ColumnCatalogEntry):
-            return f"{self.col_object.table_name}.{self.col_object.name}"
+            return f"{self.col_object.table_name}.{col_name}[{row_id}]"
         elif isinstance(self.col_object, UdfIOCatalogEntry):
-            return f"{self.col_object.udf_name}.{self.col_object.name}"
+            return f"{self.col_object.udf_name}.{col_name}[{row_id}]"
```
xzdandy, line 101: For my education, what is the purpose of `row_id`? Does name already point to a unique entry?

eva/optimizer/rules/rules.py
```diff
@@ -214,6 +216,39 @@ def apply(self, before: LogicalProject, context: OptimizerContext):
...
+        if expr.has_cache() or expr.name not in CACHEABLE_UDFS:
+            return False
+        # we do not support caching function expression instances with multiple arguments or nested function expressions
+        if len(expr.children) > 1 or not isinstance(
+            expr.children[0], TupleValueExpression
+        ):
+            return False
+        return True
+
+    def apply(self, before: LogicalApplyAndMerge, context: OptimizerContext):
```
xzdandy, line 241: I assume we eventually will have a rule that moves function expression from projection and filter into LogicalApplyAndMerge.

eva/optimizer/rules/rules_manager.py
```diff
@@ -77,6 +78,7 @@ class RulesManager:
     def __init__(self):
         self._logical_rules = [
             LogicalInnerJoinCommutativity(),
+            CacheFunctionExpressionInApply(),
```
xzdandy, line 81: Make sure I understand correctly. Since the cache cost is 0, for now this works like a rewrite rule.

eva/expression/tuple_value_expression.py
gaurav274: It might not in case you drop the table and add a new one with the same name. I wanted to keep name for readability.

eva/optimizer/rules/rules_manager.py
gaurav274: Yes, but I feel we might have some logic based on the cost in future. So wanted to keep it as a logical expression. Sounds good?

eva/catalog/catalog_utils.py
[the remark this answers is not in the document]
gaurav274: I'll do it in another PR. Needs to make changes to the storage engine to project `row_id`.

eva/expression/function_expression.py
```diff
@@ -157,6 +173,65 @@ def _gpu_enabled_function(self):
...
+        if self._cache.key:
+            keys = [child.evaluate(batch, **kwargs) for child in self._cache.key]
```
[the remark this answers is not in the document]
gaurav274: Fixed. There was a bug in the code that caused confusion.

test/integration_tests/test_reuse.py
```diff
@@ -0,0 +1,176 @@
...
+        # Stop and restart server
+        os.system("nohup eva_server --stop")
+        os.system("nohup eva_server --start &")
+
+        select_query = """SELECT id, label FROM DETRAC JOIN
+            LATERAL YoloV5(data) AS Obj(label, bbox, conf) WHERE id < 15;"""
+
+        reuse_batch = execute_query_fetch_all(select_query)
+        self._verify_reuse_correctness(select_query, reuse_batch)
+
```
jarulraj, line 132: Do we want to also check the time speedup with reuse?

jarulraj (approved):

eva/catalog/catalog_utils.py
```diff
@@ -83,3 +101,49 @@ def xform_column_definitions_to_catalog_entries(
...
+    args = tuple([arg.signature() for arg in func_expr.children])
+    entry = UdfCacheCatalogEntry(
+        name=func_expr.signature(),
+        udf_id=func_expr.udf_obj.row_id,
+        cache_path=cache_path,
+        args=args,
+        udf_depends=udf_depends,
+        col_depends=col_depends,
```
xzdandy, line 135: What is the motivation to store all this information instead of just func_expr's signature?

test/integration_tests/test_reuse.py
gaurav274: Might lead to a flaky test.

eva/catalog/catalog_utils.py
gaurav274: `udf_id`: to know which udf we are building cache on. Can be derived from the signature but it will add unnecessary bugs. 
`cache_path`: to know the path
`args`: this can be removed 
`udf_depends` and `col_depends`: required to know when to dorp cache. If any of the dependant col or udf is dropped, we removed the cache.

eva/expression/function_expression.py
```diff
@@ -157,6 +173,65 @@ def _gpu_enabled_function(self):
...
+            for key, value in zip(
+                missing_keys.iterrows(), cache_miss_results.iterrows()
+            ):
+                self._cache.store.set(key[1].to_numpy(), value[1].to_numpy())
+
+            # 4. merge the cache results
+            results[cache_miss] = cache_miss_results.to_numpy()
+
+        # 5. return the correct batch
+        cols = [obj.name for obj in self.output_objs]
```
xzdandy: Duplicate code at line 193.

eva/expression/function_expression.py
```diff
@@ -157,6 +173,65 @@ def _gpu_enabled_function(self):
...
+
+        # 1. check cache
+        # We are required to iterate over the batch row by row and check the cache.
+        # This can hurt performance, as we have to stitch together columns to generate
+        # row tuples. Is there an alternative approach we can take?
+
+        results = np.full([len(batch), len(cols)], None)
+        keys = batch
+        if self._cache.key:
```
xzdandy, line 204: When I first read code, I was thinking this is the key of the cache store. Maybe add a comment. This is to optimize the key for the function expression (data -> id).

eva/catalog/catalog_utils.py
xzdandy: Got.

xzdandy (approved):

PR #516, opened 2022-12-06
feat: Add toxic meme detection via UDF

leungyukshing:
**Toxic Meme Detection**

1. Integrated OCR text extractor and text meme classification model to identify harmful memes in a video.
2. Ways to create UDF to use our function

- Run the following command:
`CREATE UDF IF NOT EXISTS HarmfulMemeDetector INPUT (data NDARRAY (3,ANYDIM,ANYDIM)) OUTPUT (labels TEXT(10)) TYPE Classification IMPL 'eva/udfs/harmful_meme_detector.py';`

To test this function, please follow these steps:
1. LOAD FILE "data/meme/meme.mp4" INTO MEMEVideoTable; (meme.mp4 is available [here](https://drive.google.com/file/d/1iSuuMEQDkleCnksSiFpGEfzMrxBHW-11/view?usp=sharing))
2. SELECT id, HarmfulMemeDetector(data).labels FROM MEMEVideoTable WHERE id = 5; (test a single frame)
3. SELECT id, HarmfulMemeDetector(data).labels FROM MEMEVideoTable WHERE id < 50; (test a clip)

PR #551, opened 2023-01-02
feat: add support for aggregates and toxicity classification

jarulraj:
1. Based on https://github.com/georgia-tech-db/eva/pull/519
2. Brings back pip package testing in CI
3. Reduces verbosity of YOLO model
4. Update links in readme to point to the `stable` version in read-the-docs
5. Add support for toxicity detection based on https://github.com/georgia-tech-db/eva/pull/516
6. Add support for querying based on video timestamps (based on https://github.com/georgia-tech-db/eva/pull/520)


eva/parser/lark_visitor/_functions.py
```diff
@@ -125,22 +136,24 @@ def get_aggregate_function_type(self, agg_func_name):
...
+
         for child in tree.children:
             if isinstance(child, Tree):
                 if child.data == "function_arg":
                     agg_func_arg = self.visit(child)
+                elif child.data == "aggregate_function_name":
+                    agg_func_name = self.visit(child).value
+            elif isinstance(child, Token):
+                token = child.value
+                # Support for COUNT(*)
```
gaurav274, line 151: I don't understand this logic. Are we hardcoding `*` to `id` in the parser? If yes, I guess even though it is hacky, it saves us from handling this corner case in the binder. We could change it to `IDENTIFIER_COLUMN`, which is supposed to be a unique row id in all the tables.

eva/models/storage/batch.py
```diff
@@ -378,7 +380,14 @@ def aggregate(self, method: str) -> None:
         Arguments:
             method: string with one of the five above options
         """
-        self._frames = self._frames.agg([method])
+        # Aggregate ndarray
+        if isinstance(self._frames.iat[0, 0], np.ndarray):
```
gaurav274, line 384: Is it aggregating each row of the array? If yes, I suspect that will break the execution logic.

test/integration_tests/test_select_executor.py
```diff
@@ -247,6 +247,21 @@ def test_select_and_limit(self):
...
         self.assertEqual(actual_batch, expected_batch[0])
 
+    def test_select_and_aggregate(self):
+        simple_aggregate_query = "SELECT COUNT(*), AVG(id) FROM MyVideo;"
+        actual_batch = execute_query_fetch_all(simple_aggregate_query)
+
+        self.assertEqual(actual_batch.frames.iat[0, 0], 10)
+        self.assertEqual(actual_batch.frames.iat[0, 1], 4.5)
+
+        complex_aggregate_query = """SELECT SUM(id), COUNT(label)
```
gaurav274, line 257: We should add a test case with aggregate on ndarray column.

eva/parser/lark_visitor/_functions.py
jarulraj: Yes. This query currently works. I was also worried if "id" will be always present. How should I change it to `IDENTIFIER_COLUMN`?

eva/models/storage/batch.py
jarulraj: Yes, how will it break the execution logic?

The NDARRAY case is for object detection array etc. The normal case is the one that existed earlier -- `self._frames = self._frames.agg([method])`

test/integration_tests/test_select_executor.py
jarulraj: A weird part is that when the query operates on an ndarray column, it does not reduce it to a single row. It actually keeps as many rows around as the number of the rows in the input column. Does that even make sense?

jarulraj: Should we support the `NDARRAY` case?

eva/parser/lark_visitor/_functions.py
jarulraj: The unit tests do not create tables using `IDENTIFIER_COLUMN` -- so the test case fails.

test/integration_tests/test_select_executor.py
jarulraj: I have reverted back the `NDARRAY` case as it does not make sense.

eva/models/storage/batch.py
jarulraj: I have reverted back the `NDARRAY` case as it does not make sense.

jarulraj:
I just added a `ToxicityClassifier` UDF -- that runs on top of OCR labels.

eva/parser/lark_visitor/_functions.py
gaurav274: Yeah, I verified we don't support projecting `IDENTIFIER_COLUMN`, which causes the binder to fail.
`id` won't work for images or other tables. Ideally, the binder should take care of it. 
An `if` condition  [here](https://github.com/georgia-tech-db/eva/blob/04368ef72a04973b5ad3bfdf98b34e957440539e/eva/binder/statement_binder.py#L103) should fix it.

gaurav274: Reference: https://github.com/georgia-tech-db/eva/blob/04368ef72a04973b5ad3bfdf98b34e957440539e/eva/binder/binder_utils.py#L77

eva/models/storage/batch.py
gaurav274: Aggregate on `ndarray` and primitive column will result in different row counts. 

We have a different set of aggregate operators to apply row-wise aggregates like`Array_Count`

test/integration_tests/test_select_executor.py
gaurav274: Do we raise an error if the query tries to aggregate on an array column? We should add a test case to verify it. Thanks!

eva/expression/aggregation_expression.py
```diff
@@ -55,9 +55,14 @@ def evaluate(self, *args, **kwargs):
         elif self.etype == ExpressionType.AGGREGATION_MAX:
```
gaurav274, line 55: We miss an `else` condition to raise an error. Right now, we silently ignore it and return the origin batch.

eva/readers/opencv_reader.py
```diff
@@ -59,7 +59,11 @@ def _read(self) -> Iterator[Dict]:
                 _, frame = video.read()
                 frame_id = begin
                 while frame is not None and frame_id <= end:
-                    yield {"id": frame_id, "data": frame}
+                    yield {
+                        "id": frame_id,
+                        "data": frame,
+                        "seconds": frame_id // video.get(cv2.CAP_PROP_FPS),
```
gaurav274, line 65: Are we planning to add a `seconds` column? If yes, we need to register it in the catalog, or the rest of the system won't be aware of it. Add it https://github.com/georgia-tech-db/eva/blob/04368ef72a04973b5ad3bfdf98b34e957440539e/eva/catalog/catalog_utils.py#L27

Btw we should not hardcode these strings in the reader, but rather get them from the catalog object. Something to do in the future. I'll take note of it.

eva/udfs/ndarray/timestamp.py
```diff
@@ -0,0 +1,49 @@
+# coding=utf-8
```
gaurav274, line 1: What is this for?

eva/udfs/toxicity_classifier.py
```diff
@@ -0,0 +1,79 @@
...
+        # reconstruct dimension of the input
+        outcome = pd.DataFrame()
+        dataframe_size = text_dataframe.size
+
+        for i in range(0, dataframe_size):
+            text = text_dataframe.iat[i, 0]
+            single_result = self.model.predict(text)
+            toxicity_score = single_result["toxicity"][0]
+            if toxicity_score >= self.threshold:
+                outcome = outcome.append({"labels": "toxic"}, ignore_index=True)
```
gaurav274, line 75: I changed it to use `list` for the append operation. `DataFrame ` throws a lot of warnings. You can refer to other udfs.

eva/udfs/ndarray/timestamp.py
```diff
@@ -0,0 +1,49 @@
...
+            second           timestamp
+        0   int           0   string
+        1   int           1   string
+        """
+
+        # Sanity check
+        if len(inp.columns) != 1:
+            raise ValueError("input must only contain one column (seconds)")
+
+        seconds = pd.DataFrame(inp[inp.columns[0]])
```
gaurav274, line 42: Isn't it `no-op`?

test/catalog/test_catalog_manager.py
```diff
@@ -76,6 +76,7 @@ def test_create_multimedia_table_catalog_entry(self, mock):
             ColumnDefinition(
                 "data", ColumnType.NDARRAY, NdArrayType.UINT8, [None, None, None]
```
gaurav274, line 77: `array_dimension` should be a tuple.

gaurav274:
> I just added a `ToxicityClassifier` UDF -- that runs on top of OCR labels.

Thanks! This will be a fun example to showcase 💯

eva/readers/opencv_reader.py
jarulraj: Yes, this is done.

eva/udfs/toxicity_classifier.py
jarulraj: Did you push your change?

eva/udfs/ndarray/timestamp.py
jarulraj: This is a timestamp UDF.

jarulraj: `SELECT id, seconds, Timestamp(seconds)
                          FROM MyVideo
                          WHERE Timestamp(seconds) <= "00:00:01";`

PR #553, opened 2023-01-05
feat: Fix legacy issue in index

jiashenC:
* Remove secondary index maintained by ourselves. Instead, we now use `IDMap` provided by the FAISS library, which is also faster in terms of performance.
* Add support of creating index on UDF expression directly. Along with that, add udf signature as part of the index catalog, so the expression can be matched again when index is used later.

### DIFF
--- a/eva/binder/statement_binder.py
+++ b/eva/binder/statement_binder.py
@@ -80,6 +80,8 @@ class StatementBinder:
     @bind.register(CreateIndexStatement)
     def _bind_create_index_statement(self, node: CreateIndexStatement):
         self.bind(node.table_ref)
+        if node.udf_func:
+            self.bind(node.udf_func)
 
         # TODO: create index currently only supports single numpy column.
         assert len(node.col_list) == 1, "Index cannot be created on more than 1 column"
@@ -89,9 +91,13 @@ class StatementBinder:
 
         col_def = node.col_list[0]
         table_ref_obj = node.table_ref.table.table_obj
+
         col = [col for col in table_ref_obj.columns if col.name == col_def.name][0]
         assert col.type == ColumnType.NDARRAY, "Index input needs to be numpy array"
-        assert col.array_type == NdArrayType.FLOAT32, "Index input needs to be float32"
+
+        # Relax type checking for raw input table.
+        if not node.udf_func:
+            assert col.array_type == NdArrayType.FLOAT32, "Index input needs to be float32"
 
     @bind.register(SelectStatement)
     def _bind_select_statement(self, node: SelectStatement):
--- a/eva/binder/statement_binder.py
+++ b/eva/binder/statement_binder.py
@@ -23,7 +23,7 @@ from eva.binder.binder_utils import (
 )
 from eva.binder.statement_binder_context import StatementBinderContext
 from eva.catalog.catalog_manager import CatalogManager
-from eva.catalog.catalog_type import ColumnType, NdArrayType, TableType
+from eva.catalog.catalog_type import NdArrayType, TableType, IndexType
 from eva.expression.abstract_expression import AbstractExpression
 from eva.expression.function_expression import FunctionExpression
 from eva.expression.tuple_value_expression import TupleValueExpression
@@ -89,17 +89,27 @@ class StatementBinder:
         # TODO: create index currently only works on TableInfo, but will extend later.
         assert node.table_ref.is_table_atom(), "Index can only be created on Tableinfo"
 
-        col_def = node.col_list[0]
-        table_ref_obj = node.table_ref.table.table_obj
-
-        col = [col for col in table_ref_obj.columns if col.name == col_def.name][0]
-        assert col.type == ColumnType.NDARRAY, "Index input needs to be numpy array"
-
-        # Relax type checking for raw input table.
-        if not node.udf_func:
-            assert (
-                col.array_type == NdArrayType.FLOAT32
-            ), "Index input needs to be float32"
+        if IndexType.is_faiss_index_type(node.index_type):
+            if not node.udf_func:
+                # Feature table type needs to be float32 numpy array.
+                col_def = node.col_list[0]
+                table_ref_obj = node.table_ref.table.table_obj
+                col = [col for col in table_ref_obj.columns if col.name == col_def.name][0]
+                if not col.array_type == NdArrayType.FLOAT32:
+                    raise BinderError("Index input needs to be float32.")
+                if not len(col.array_dimensions) == 2:
+                    raise BinderError("Index input needs to be 2 dimensional.")
+            else:
+                # Output of the UDF should be 2 dimension and float32 type.
+                catalog_manager = CatalogManager()
+                udf_obj = catalog_manager.get_udf_catalog_entry_by_name(node.udf_func.name)
+                for output in udf_obj.outputs:
+                    if not output.array_type == NdArrayType.FLOAT32:
+                        raise BinderError("Index input needs to be float32.")
+                    if not len(output.array_dimensions) == 2:
+                        raise BinderError("Index input needs to be 2 dimensional.")
+        else:
+            raise BinderError("Index type {} is not supported.".format(node.index_type))
 
     @bind.register(SelectStatement)
     def _bind_select_statement(self, node: SelectStatement):
--- a/eva/binder/statement_binder.py
+++ b/eva/binder/statement_binder.py
@@ -97,7 +97,9 @@ class StatementBinder:
 
         # Relax type checking for raw input table.
         if not node.udf_func:
-            assert col.array_type == NdArrayType.FLOAT32, "Index input needs to be float32"
+            assert (
+                col.array_type == NdArrayType.FLOAT32
+            ), "Index input needs to be float32"
 
     @bind.register(SelectStatement)
     def _bind_select_statement(self, node: SelectStatement):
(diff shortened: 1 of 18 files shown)


eva/binder/statement_binder.py
```diff
@@ -89,9 +91,15 @@ def _bind_create_index_statement(self, node: CreateIndexStatement):
 
         col_def = node.col_list[0]
         table_ref_obj = node.table_ref.table.table_obj
+
         col = [col for col in table_ref_obj.columns if col.name == col_def.name][0]
         assert col.type == ColumnType.NDARRAY, "Index input needs to be numpy array"
-        assert col.array_type == NdArrayType.FLOAT32, "Index input needs to be float32"
+
+        # Relax type checking for raw input table.
```
jarulraj, line 98: Is this relaxing type checking for raw input table or udf case?

eva/executor/create_index_executor.py
```diff
@@ -56,40 +55,32 @@ def exec(self):
...
-            input_dim, num_rows = -1, 0
-            logical_to_row_id = dict()
+            input_dim = -1
             storage_engine = StorageEngine.factory(feat_catalog_entry)
-            for batch in storage_engine.read(feat_catalog_entry, 1):
-                # Pandas wraps numpy array as an object inside a numpy
-                # array. Use zero index to get the actual numpy array.
-                feat = batch.column_as_numpy_array(feat_col_name)[0]
+            for input_batch in storage_engine.read(feat_catalog_entry, 1):
+                if udf_func:
```
jarulraj, line 67: We should add some comments on what is happening -- like SimilaritySearch(f).x -> f

eva/executor/create_index_executor.py
```diff
@@ -173,43 +148,11 @@ def _get_index_save_path(self):
...
-            col_catalog_entries,
-            identifier_column="logical_id",
-            table_type=TableType.STRUCTURED_DATA,
-        )
-        return table_catalog_entry
-
     def _create_index(self, index_type: IndexType, input_dim: int):
         if index_type == IndexType.HNSW:
-            return faiss.IndexHNSWFlat(input_dim, 32)
+            # HSNW is the actual index. Faiss also provides
```
jarulraj, line 153: Can we add a FAISS documentation url for reference?

eva/parser/lark_visitor/_create_statements.py
```diff
@@ -292,11 +293,24 @@ def create_index(self, tree):
...
+                elif child.data == "index_elem":
+                    index_elem = self.visit(child)
+
+        # Parse either a single UDF function call or column list.
+        col_list, udf_func = None, None
+        if not isinstance(index_elem, list):
+            udf_func = index_elem
+
+            # Traverse to the tuple value expression.
+            while not isinstance(index_elem, TupleValueExpression):
```
jarulraj, line 305: Why do we need this traversal? This should happen automatically in -- https://github.com/georgia-tech-db/eva/blob/f88db49bb3d6f9adbac6a2ae57fc82047a9df467/eva/parser/lark_visitor/__init__.py#L50

test/integration_tests/test_create_index_executor.py
```diff
@@ -156,8 +163,42 @@ def test_should_cleanup_when_exception(self, faiss_mock):
         with self.assertRaises(ExecutorError):
             execute_query_fetch_all(query)
 
-        # Check secondary index is dropped.
-        secondary_index_tb_name = "secondary_index_{}_{}".format(
-            "HNSW", "testCreateIndexName"
+        # Check faulty index is not persisted on the disk
+        self.assertFalse(os.path.exists(self._index_save_path()))
+
+    def test_should_create_index_with_udf(self):
```
jarulraj, line 169: Do we need another test without specifying `USING HNSW`?

jarulraj (approved):

eva/binder/statement_binder.py
jiashenC: For the UDF case. If the table already have feature vectors, their type needs to be float32. Otherwise, the type can be relaxed since it will go through the feature extractor, who outputs a float32 feature vector.

eva/parser/lark_visitor/_create_statements.py
jiashenC: My goal is to get the column definition for the query. For query that involves UDF expression: e.g., `CREATE INDEX ON table (FeatureExtractor(col));`, the column definition becomes the children tree of `index_elem`, so I need to do a traversal here. Is there a better way to do that in the current parser?

eva/binder/statement_binder.py
jarulraj: The comment seems wrong then?

eva/parser/lark_visitor/_create_statements.py
jarulraj: I see. You will need to just add a similar function below for `index_elem`: 

```
               elif child.data == "index_type":
                    index_type = self.visit(child)
                elif child.data == "index_elem":
                    index_elem = self.visit(child)

        # Parse either a single UDF function call or column list.
        col_list, udf_func = None, None
        if not isinstance(index_elem, list):
            udf_func = index_elem

            # Traverse to the tuple value expression.
            while not isinstance(index_elem, TupleValueExpression):
                index_elem = index_elem.children[0]
            index_elem = [index_elem]

        col_list = [
            ColumnDefinition(tv_expr.col_name, None, None, None)
            for tv_expr in index_elem
        ]

def index_elem(self, tree):
        uid_list = []
        function_call = None

        for child in tree.children:
            if isinstance(child, Tree):
                if child.data == "uid_list":
                    uid_list = self.visit(child)
                elif child.data == "function_call":
                    function_call = self.visit(child)

        return None
`

**Comment by xzdandy** [comment]
I'd like to get some context on adding support of creating index on UDF expression directly. For this case `CREATE INDEX ON table (FeatureExtractor(col));`, what is the expected index we want to create? Do we treat `FeatureExtractor(col)` as a table and create index on the feature extracted and store the mapping between the index and `row_id`?

**Comment by jiashenC** [comment]
> I'd like to get some context on adding support of creating index on UDF expression directly. For this case `CREATE INDEX ON table (FeatureExtractor(col));`, what is the expected index we want to create? Do we treat `FeatureExtractor(col)` as a table and create index on the feature extracted and store the mapping between the index and `row_id`?

Yes, you are right.

File: eva/binder/statement_binder.py (hunk 1 above)
Comment (xzdandy) [reply eva/binder/statement_binder.py:98]: In general, the index can be created on the types that are not float32 vectors. More accurately, it should be given the index method we are using, which is `HNSW`, we need either the column or the output of `node.udf_func` to be NDARRAY FLOAT32. For clarity, I think it will good to have this logic in the bind. 
```
If index_type == "HNSW":
   if node.udf_func is not None:
        # Check the output type is NDARRAY FLOAT32
   else:
        # Check the column type is NDARRAY FLOAT32
else:
   raise BinderError("Unsupported index method")
```
PS: I think we don't need the col type to be NDARRAY if there is `node.udf_func`.

File: eva/executor/create_index_executor.py [hunk 6]
```diff
@@ -56,40 +55,32 @@ def exec(self):
...
-            secondary_index = Batch(
-                pd.DataFrame(data={"logical_id": logical_id, "row_id": row_id})
-            )
-            storage_engine.write(secondary_index_catalog_entry, secondary_index)
+                # Row ID for mapping back to the row.
+                row_id = input_batch.column_as_numpy_array(IDENTIFIER_COLUMN)[0]
+                index.add_with_ids(feat, np.array([row_id]))
 
             # Persist index.
             faiss.write_index(index, self._get_index_save_path())
```
xzdandy, line 86: This are all `faiss`-specific implementation. I think it will be better to check the `index_type` at the begining of the executor instead of in the `_create_index`.

eva/parser/eva.lark
```diff
@@ -41,6 +41,9 @@ udf_impl: string_literal
 
 index_type: USING HNSW
 
+index_elem: ("(" uid_list ")"
```
xzdandy, line 44: I think we only support single column right now.

xzdandy:
> > I'd like to get some context on adding support of creating index on UDF expression directly. For this case `CREATE INDEX ON table (FeatureExtractor(col));`, what is the expected index we want to create? Do we treat `FeatureExtractor(col)` as a table and create index on the feature extracted and store the mapping between the index and `row_id`?
> 
> Yes, you are right.

Thanks. Overall LGTM.

eva/parser/eva.lark
jiashenC: Yes. I had this discussion with @gaurav274. We agree to keep this in the parser but do a check in the statement binder.

eva/binder/statement_binder.py
jiashenC: Got it. FAISS itself also has other index types besides `HNSW`. The FLOAT32 type checking applies to those other index types as well, so I think the `if index_type == "HNSW":` is probably not needed?

xzdandy: I think that the query `CREATE INDEX ON table (ColorDetector(col));` should be legit. It's okay we are not supporting it now with FAISS. But the binder should raise `unsupported index` instead of `Index input needs to be numpy array FLOAT32` for clarity.

eva/executor/create_index_executor.py
jiashenC: Sure. Refactor the code.

test/integration_tests/test_create_index_executor.py
jiashenC: I think this will lead to some ambiguities. As @xzdandy suggested, I will throw an error for this case in the binder.

gaurav274 (approved):

xzdandy (approved):

PR #554, opened 2023-01-06
fix: 03-emotion-analysis tutorial notebook

Aryan-Rajoria:
- Fixing google colab link
- Adding new link for Emotion detection model `.t7` file (the old dropbox link seems to be removed).


jarulraj:
Good catch. We can merge it in.

jarulraj:
Resolved in: https://github.com/georgia-tech-db/eva/pull/551

PR #559, opened 2023-01-09
feat: optimizer orderby limit rule

jiashenC:
* Add a faiss index scan executor.
* Add a rule to automatically convert orderby + limit on similarity UDF to existing faiss index scan.
* Disable rule whenever there is a predicate on table.


eva/optimizer/rules/rules.py
```diff
@@ -413,6 +418,114 @@ def apply(self, before: LogicalFilter, context: OptimizerContext):
...
+        pattern = Pattern(OperatorType.LOGICALLIMIT)
+        orderby_pattern = Pattern(OperatorType.LOGICALORDERBY)
+        orderby_pattern.append_child(Pattern(OperatorType.DUMMY))
+        pattern.append_child(orderby_pattern)
+        super().__init__(
+            RuleType.COMBINE_SIMILARITY_ORDERBY_AND_LIMIT_TO_FAISS_INDEX_SCAN, pattern
+        )
+
+        # Entries populate after rule egibility validation.
+        self._index_catalog_entry = None
```
gaurav274, line 441: We should not do this as a rule is supposed to be stateless. You can move the check code to the `apply.` We can easily support cases where the rule doesn't return any plan. I made the required changes in my other PR. I will merge it so that you can leverage it.

eva/executor/faiss_index_scan_executor.py
```diff
@@ -0,0 +1,90 @@
...
+        index = faiss.read_index(index_catalog_entry.save_file_path)
+
+        # Aggregate batches from sequential scan.
+        seq_scan_batch = Batch()
+        for batch in self.children[0].exec():
+            seq_scan_batch = Batch.concat([seq_scan_batch, batch])
+
+        # Get the query feature vector. Create a dummy
+        # batch to retreat a single file path.
+        dummy_batch = Batch(
```
gaurav274, line 52: Did not understand this

eva/executor/faiss_index_scan_executor.py
```diff
@@ -0,0 +1,90 @@
...
+                row_id_alias = alias
+        index_scan_batch = Batch(
+            pd.DataFrame(
+                {
+                    "similarity.distance": distance_list,
+                    "{}._row_id".format(row_id_alias): row_id_list,
+                }
+            )
+        )
+        res_df = seq_scan_batch.frames.merge(
```
gaurav274, line 85: Rather than bring the entire table in memory, can we leverage the fact that the limit clause is typically small, and we can do a nested join?

eva/optimizer/rules/rules.py
```diff
@@ -413,6 +418,114 @@ def apply(self, before: LogicalFilter, context: OptimizerContext):
         return root_node
 
 
+class CombineSimilarityOrderByAndLimitToFaissIndexScan(Rule):
+    """
+
+    Limit(10)
+        |
+    OrderBy(func)        ->        IndexScan(10)
```
gaurav274, line 426: We append `A` to the IndexScan right?

eva/plan_nodes/faiss_index_scan_plan.py
```diff
@@ -0,0 +1,47 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from eva.expression.function_expression import FunctionExpression
+from eva.plan_nodes.abstract_plan import AbstractPlan
+from eva.plan_nodes.types import PlanOprType
+
+
+class FaissIndexScanPlan(AbstractPlan):
+    def __init__(self, index_name: str, query_num: int, query_expr: FunctionExpression):
```
gaurav274, line 21: Please add the doc string to the class. Also, consider renaming `query_name` and `query_expr`.  They aren't intuitive to me. Thanks!

eva/executor/faiss_index_scan_executor.py
jiashenC: I just push this down, so that constant value expression can give back results (https://github.com/georgia-tech-db/eva/blob/c2c2bdcb7d46848764e592a462d452953c3209fd/eva/expression/constant_value_expression.py#L36). 

What would be a better way to do this?

eva/optimizer/rules/rules.py
jiashenC: I see. I can move some code to `apply`. Then, when are we supposed to use check?

eva/executor/faiss_index_scan_executor.py
jiashenC: Got it. Let me give a try.

eva/optimizer/rules/rules.py
gaurav274: To cater invalidations based on simple logic. It isn't that important in our setting since we don't have a lot of rules.

eva/executor/faiss_index_scan_executor.py
gaurav274: Can you explain with an example? I'm not getting the logic.

jiashenC: I am just passing a dummy batch so that `len(batch)` is not 0. Otherwise, the call `Open("some file path")` will not get the path.

eva/executor/faiss_index_scan_executor.py
```diff
@@ -0,0 +1,90 @@
...
+        dummy_batch = Batch(
+            frames=pd.DataFrame(
+                {"0": [0]},
+            )
+        )
+        search_batch = self.query_expr.evaluate(dummy_batch)
+
+        # Scan index.
+        search_batch.drop_column_alias()
+        search_feat = search_batch.column_as_numpy_array("features")[0]
```
gaurav274, line 61: Curious, why do we have a hardcoded string "feature" here?

jiashenC: You are right. It won't work if the index is built on a feature vector table. I will need to fix this.

In general, what is a good way to index a column. In this case, the column name can be different, then we can only use index. But then, there is no guarantee which column appears at which index.

eva/executor/faiss_index_scan_executor.py
gaurav274: I see

eva/executor/faiss_index_scan_executor.py
gaurav274: We can get the column name from the udf object.

jiashenC: Fix it.

eva/executor/faiss_index_scan_executor.py
```diff
@@ -0,0 +1,93 @@
...
+            row_id_list.append(row_id)
+
+        # Load projected columns from disk and join with search results.
+        row_id_col_name = None
+        res_row_list = [None for _ in range(self.limit_count.value)]
+        for batch in self.children[0].exec():
+            column_list = batch.columns
+            if not row_id_col_name:
+                row_id_alias = get_row_id_column_alias(column_list)
+                row_id_col_name = "{}._row_id".format(row_id_alias)
```
gaurav274, line 81: replace `_row_id` with IDENTIFIER_COLUMN

eva/executor/faiss_index_scan_executor.py
```diff
@@ -0,0 +1,93 @@
...
+
+            # Nested join.
+            for _, row in batch.frames.iterrows():
```
gaurav274, line 84: Maybe Join here is more performant. But we can come back to it later.

eva/optimizer/operators.py
```diff
@@ -1207,3 +1209,50 @@ def __hash__(self) -> int:
                 self._merge_type,
             )
         )
+
+
+class LogicalFaissIndexScan(Operator):
```
gaurav274, line 1214: Also change the member names similar to physical node

gaurav274 (approved):

eva/executor/faiss_index_scan_executor.py
```diff
@@ -0,0 +1,93 @@
...
+                for idx, rid in enumerate(row_id_list):
+                    if rid == row[row_id_col_name]:
+                        res_row = dict()
+                        for col_name in column_list:
+                            res_row[col_name] = row[col_name]
+                        res_row["similarity.distance"] = distance_list[idx]
```
gaurav274, line 90: There is no way to project distance, right? For that, I guess we need to use select list alias.

gaurav274 (approved):

PR #565, opened 2023-01-17
feat: new expression utils and bug fixes

gaurav274:


eva/expression/expression_utils.py
```diff
@@ -22,31 +22,48 @@
...
+        expression_list.extend(to_conjunction_list(expression_tree.children[0]))
+        expression_list.extend(to_conjunction_list(expression_tree.children[1]))
     else:
         expression_list.append(expression_tree)
 
     return expression_list
 
 
-def conjuction_list_to_expression_tree(
+def and_(
```
jiashenC, line 53: Any reason we change to this name? I think the original name is more intuitive.

jiashenC (approved):
LGTM

eva/expression/expression_utils.py
gaurav274: Sure, I'll revert back

PR #566, opened 2023-01-20
feat: object tracking

gaurav274:
Supported query: 
```
SELECT id, T.iids, T.bboxes, T.scores, T.labels
FROM MyVideo JOIN LATERAL EXTRACT_OBJECT(data, YoloV5, NorFairTracker)
    AS T(iids, labels, bboxes, scores)
WHERE id < 30;
```


jarulraj:
'DataFrame' object has no attribute 'append'

eva/executor/extract_object_executor.py
```diff
@@ -0,0 +1,44 @@
...
+            objects = self.detector.evaluate(batch)
+
+            # run tracker
+            intermediate_batch = Batch.merge_column_wise([batch, objects])
+            results = self.tracker.evaluate(intermediate_batch)
+
+            # merge batch and tracker outputs and yield
+            outcomes = Batch.merge_column_wise([batch, results])
+            if self.node.do_unnest:
+                outcomes.unnest(results.columns)
```
xzdandy, line 43: What is the difference between doing unnest and not for `EXTRACT_OBJECT`?

eva/optimizer/rules/rules.py
```diff
@@ -404,6 +406,37 @@ def apply(self, before: LogicalFilter, context: OptimizerContext):
...
+class XformExtractObjectToLinearFlow(Rule):
+    """If the inner node of a lateral join is a Extract_Object function-valued
+    expression, we eliminate the join node and make the inner node the parent of the
+    outer node. This produces a linear data flow path."""
+
+    #     LogicalJoin(Lateral)               LogicalExtractObject
+    #     /           \                 ->        |
+    #    A        LogicalExtractObject            A
+
+    def __init__(self):
```
xzdandy, line 418: This is a special case. Ideally, I think it will be better to we can base UDFs on `LogicalApplyAndMerge`. So later, if we have a rule for ray, or for the old reuse, or any function-oriented optimization. We do not need to consider this as a special case and handle it accordingly.

eva/utils/math_utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+    Args:
+        bboxes (numpy.ndarray): Array of shape `(n, 4)` or of shape `(4,)` where
+            each row contains `(xmin, ymin, width, height)`.
+
+    Returns:
+        numpy.ndarray: Centroid (x, y) coordinates of shape `(n, 2)` or `(2,)`.
+
+    """
+
```
xzdandy, line 31: I have a feeling this function is not very efficient.

eva/executor/extract_object_executor.py
gaurav274: Same as object detection, one object per row or all objects of a frame per row

eva/optimizer/rules/rules.py
gaurav274: I see. Reduces redundancy. Good idea!

eva/utils/math_utils.py
gaurav274: I borrowed it from another library. I choose it as it handles corner cases. Any suggestions?

eva/executor/extract_object_executor.py
```diff
@@ -0,0 +1,44 @@
...
+        for batch in child_executor.exec():
+            # run detector
+            objects = self.detector.evaluate(batch)
+
+            # run tracker
+            intermediate_batch = Batch.merge_column_wise([batch, objects])
+            results = self.tracker.evaluate(intermediate_batch)
+
+            # merge batch and tracker outputs and yield
+            outcomes = Batch.merge_column_wise([batch, results])
```
jiashenC, line 41: How does the tracking outputs look like here?

eva/utils/math_utils.py
xzdandy: Maybe run on a long video and check how many time it costs. If it is all right, we can leave it as it is. We can think this as  the overhead of tracker UDF (not EVA).

gaurav274: I see. We are benchmarking `eva` on larger datasets for correctness and performance, so will take care of it with that exercise.

eva/optimizer/rules/rules.py
gaurav274: Thanks for this idea! The logic is quite clean now.

pchunduri6 (commented):
One-off invocations of the tracker would generate different object-ids, correct? Let's say I run the object tracker on frames 1 to 30, then 70 to 100 in a continuous scene. Would the same objects have different object ids? The follow-up question would be, can we merge them in a subsequent call? Seems like a good use-case for supporting re-identification.

pchunduri6 (commented):
Can we use the `SEGMENT` construct to track sequences within a segment? Since tracking is done on a sequence of consecutive frames, calling an object tracker on arbitrary frames in the videos seems unintuitive.

eva/binder/binder_utils.py
```diff
@@ -134,3 +138,74 @@ def resolve_alias_table_value_expression(node: FunctionExpression):
...
+    columns = get_video_table_column_definitions()
+    tracker.append_child(
+        TupleValueExpression(
+            col_name=columns[1].name, table_alias=video_data.table_alias
+        )
+    )
+    tracker.append_child(video_data.copy())
+    binder_context.bind(tracker)
+    # append the bound output of detector
+    for obj in detector.output_objs:
```
pchunduri6, line 190: Would the object detection be run on every frame? norfair supports the period operator, but I'm curious how we can combine that with our operators (e.g., SAMPLE) to run detections only once every few frames and associate tracks.

gaurav274: Nice observation! Let me think about it.

gaurav274:
> Can we use the `SEGMENT` construct to track sequences within a segment? Since tracking is done on a sequence of consecutive frames, calling an object tracker on arbitrary frames in the videos seems unintuitive.

I don't understand what you mean by `arbitrary frames`. We are always calling the tracker on consecutive frames and the tracker ensures it provides a unique id to any new object.

gaurav274:
> One-off invocations of the tracker would generate different object-ids, correct? Let's say I run the object tracker on frames 1 to 30, then 70 to 100 in a continuous scene. Would the same objects have different object ids? The follow-up question would be, can we merge them in a subsequent call? Seems like a good use-case for supporting re-identification.

Yes, this can happen. However, if we add support for a tracker that also takes into about `reid` features, we should be able to achieve it. This can be an extension of the current implementation.

eva/binder/binder_utils.py
xzdandy: Does the current implementation already work with SAMPLE? The frames passed to the object detector is after sampling.

xzdandy:
> > Can we use the `SEGMENT` construct to track sequences within a segment? Since tracking is done on a sequence of consecutive frames, calling an object tracker on arbitrary frames in the videos seems unintuitive.
> 
> I don't understand what you mean by `arbitrary frames`. We are always calling the tracker on consecutive frames and the tracker ensures it provides a unique id to any new object. Reg `SEGMENT`: I was thinking of using `GROUP BY` on object ids returned by the tracker to construct actual tracks and maybe support some geometric predicates on it.

I think ``

eva/binder/binder_utils.py
gaurav274: It will work, however incorrectly. Tracker won't know that we are skipping frames.

gaurav274: I was sending `frame_id` to the `tracker` for such use cases. Fixed it!

gaurav274:
@pchunduri6 Let me know your thoughts.

xzdandy:
> > One-off invocations of the tracker would generate different object-ids, correct? Let's say I run the object tracker on frames 1 to 30, then 70 to 100 in a continuous scene. Would the same objects have different object ids? The follow-up question would be, can we merge them in a subsequent call? Seems like a good use-case for supporting re-identification.
> 
> Yes, this can happen. However, if we add support for a tracker that also takes into about `reid` features, we should be able to achieve it. This can be an extension of the current implementation.
> 
> I'm deliberately sending the `frame` data to the tracker for extracting `reid` features if required.

This leads to an interesting problem. Predicate pushdown may not work in this case. For example, when we have `WHERE id < 30 or id > 70`. Filtering the frame first will give us different results.

gaurav274:
> > > One-off invocations of the tracker would generate different object-ids, correct? Let's say I run the object tracker on frames 1 to 30, then 70 to 100 in a continuous scene. Would the same objects have different object ids? The follow-up question would be, can we merge them in a subsequent call? Seems like a good use-case for supporting re-identification.
> > 
> > 
> > Yes, this can happen. However, if we add support for a tracker that also takes into about `reid` features, we should be able to achieve it. This can be an extension of the current implementation.
> > I'm deliberately sending the `frame` data to the tracker for extracting `reid` features if required.
> 
> This leads to an interesting problem. Predicate pushdown may not work in this case. For example, when we have `WHERE id < 30 or id > 70`. Filtering the frame first will give us different results.

Yup, that is true. But isn't the user deliberately only wanting to track the objects from 30-70? We can add it as a warning in the documentation. I haven't worked on documentation yet. Was thinking of doing it in another PR.

xzdandy:
> > > > One-off invocations of the tracker would generate different object-ids, correct? Let's say I run the object tracker on frames 1 to 30, then 70 to 100 in a continuous scene. Would the same objects have different object ids? The follow-up question would be, can we merge them in a subsequent call? Seems like a good use-case for supporting re-identification.
> > > 
> > > 
> > > Yes, this can happen. However, if we add support for a tracker that also takes into about `reid` features, we should be able to achieve it. This can be an extension of the current implementation.
> > > I'm deliberately sending the `frame` data to the tracker for extracting `reid` features if required.
> > 
> > 
> > This leads to an interesting problem. Predicate pushdown may not work in this case. For example, when we have `WHERE id < 30 or id > 70`. Filtering the frame first will give us different results.
> 
> Yup, that is true. But isn't the user deliberately only wanting to track the objects from 30-70? We can add it as a warning in the documentation. I haven't worked on documentation yet. Was thinking of doing it in another PR.

This is not a common use case, but just for example. User are interested in vehicles before 6PM or after 7PM. So the query will have `WHERE timestamp < 6PM or timestamp > 7PM`. Now consider there is a vehicle circling around through the whole time. Without predicate pushdown, the vehicle will have a unique id associated with it (assuming no errors). With predicate push down, the vehicle can have two ids because the location of the vehicle in the last frame for `timestamp < 6PM` and the first frame for `timestamp > 7PM` can be quite different. 
There is also a risk that random vehicles in the first frame for `timestamp > 7pm` will be matched to random vehicles in the last frame for `timestamp < 6PM`, which will not happen if we don't do predicate push down.

gaurav274:
> > > > > One-off invocations of the tracker would generate different object-ids, correct? Let's say I run the object tracker on frames 1 to 30, then 70 to 100 in a continuous scene. Would the same objects have different object ids? The follow-up question would be, can we merge them in a subsequent call? Seems like a good use-case for supporting re-identification.
> > > > 
> > > > 
> > > > Yes, this can happen. However, if we add support for a tracker that also takes into about `reid` features, we should be able to achieve it. This can be an extension of the current implementation.
> > > > I'm deliberately sending the `frame` data to the tracker for extracting `reid` features if required.
> > > 
> > > 
> > > This leads to an interesting problem. Predicate pushdown may not work in this case. For example, when we have `WHERE id < 30 or id > 70`. Filtering the frame first will give us different results.
> > 
> > 
> > Yup, that is true. But isn't the user deliberately only wanting to track the objects from 30-70? We can add it as a warning in the documentation. I haven't worked on documentation yet. Was thinking of doing it in another PR.
> 
> This is not a common use case, but just for example. User are interested in vehicles before 6PM or after 7PM. So the query will have `WHERE timestamp < 6PM or timestamp > 7PM`. Now consider there is a vehicle circling around through the whole time. Without predicate pushdown, the vehicle will have a unique id associated with it (assuming no errors). With predicate push down, the vehicle can have two ids because the location of the vehicle in the last frame for `timestamp < 6PM` and the first frame for `timestamp > 7PM` can be quite different. There is also a risk that random vehicles in the first frame for `timestamp > 7pm` will be matched to random vehicles in the last frame for `timestamp < 6PM`, which will not happen if we don't do predicate push down.

1. Yes, that is a problem. I can't think of a solution right now. Not pushing down the predicate will be super expensive. 
2. I believe trackers should be able to handle it. Most trackers consume frame id to compute flow. But depends on the tracker. 
What do you suggest?

pchunduri6:
> > Can we use the `SEGMENT` construct to track sequences within a segment? Since tracking is done on a sequence of consecutive frames, calling an object tracker on arbitrary frames in the videos seems unintuitive.
> 
> I don't understand what you mean by `arbitrary frames`. We are always calling the tracker on consecutive frames and the tracker ensures it provides a unique id to any new object. Reg `SEGMENT`: I was thinking of using `GROUP BY` on object ids returned by the tracker to construct actual tracks and maybe support some geometric predicates on it.

The `GROUP BY` idea looks good to me.

By arbitrary frames, I meant the upstream query could potentially return any frames based on the predicate right. For example, consider the return ids `10, 120, 250, and 300`. What would track mean when run on such frames? It's not an implementation or design issue. From a usability standpoint, I was just thinking if there's a way to limit the use of the tracker to specific types of input sequences.

xzdandy:
> > > > > > One-off invocations of the tracker would generate different object-ids, correct? Let's say I run the object tracker on frames 1 to 30, then 70 to 100 in a continuous scene. Would the same objects have different object ids? The follow-up question would be, can we merge them in a subsequent call? Seems like a good use-case for supporting re-identification.
> > > > > 
> > > > > 
> > > > > Yes, this can happen. However, if we add support for a tracker that also takes into about `reid` features, we should be able to achieve it. This can be an extension of the current implementation.
> > > > > I'm deliberately sending the `frame` data to the tracker for extracting `reid` features if required.
> > > > 
> > > > 
> > > > This leads to an interesting problem. Predicate pushdown may not work in this case. For example, when we have `WHERE id < 30 or id > 70`. Filtering the frame first will give us different results.
> > > 
> > > 
> > > Yup, that is true. But isn't the user deliberately only wanting to track the objects from 30-70? We can add it as a warning in the documentation. I haven't worked on documentation yet. Was thinking of doing it in another PR.
> > 
> > 
> > This is not a common use case, but just for example. User are interested in vehicles before 6PM or after 7PM. So the query will have `WHERE timestamp < 6PM or timestamp > 7PM`. Now consider there is a vehicle circling around through the whole time. Without predicate pushdown, the vehicle will have a unique id associated with it (assuming no errors). With predicate push down, the vehicle can have two ids because the location of the vehicle in the last frame for `timestamp < 6PM` and the first frame for `timestamp > 7PM` can be quite different. There is also a risk that random vehicles in the first frame for `timestamp > 7pm` will be matched to random vehicles in the last frame for `timestamp < 6PM`, which will not happen if we don't do predicate push down.
> 
> 1. Yes, that is a problem. I can't think of a solution right now. Not pushing down the predicate will be super expensive.
> 2. I believe trackers should be able to handle it. Most trackers consume frame id to compute flow. But depends on the tracker.
>    What do you suggest?

I see. We can print a warning message for now. Re-id is needed I  think to achieve the same results with / without predicate push down.  Even then, there still can be errors from re-id.

test/integration_tests/test_pytorch.py
```diff
@@ -243,6 +243,27 @@ def test_should_run_ocr_on_cropped_data(self):
...
+
+        select_query = """
+            SELECT id, T.iid, T.bbox, T.score, T.label
+            FROM MyVideo JOIN LATERAL
+                UNNEST(EXTRACT_OBJECT(data, Yolo, NorFairTracker)) AS T(iid, label, bbox, score)
+            WHERE id < 30;
+            """
+        actual_batch = execute_query_fetch_all(select_query)
+        # do some more meaningful check
+        self.assertEqual(len(actual_batch), 685)
```
xzdandy, line 265: @gaurav274 Where does the number 685 come from? Does it not change with the version of yolo and tracker installed from pip?

gaurav274: It might change. I wasn't sure about the check I should add.

xzdandy: With the current Yolo. it is 625 instead of 685. I think this type test cases will be flaky. I suggest we either remove it, or use dummy detector and tracker, so the  results are deterministic.

xzdandy: Actually, there is a better idea, we can get the count from the previous query.

test/integration_tests/test_reuse.py
```diff
@@ -170,6 +170,17 @@ def test_reuse_filter_with_project(self):
...
+    def test_reuse_in_extract_object(self):
+        select_query = """
+            SELECT id, T.iids, T.bboxes, T.scores, T.labels
+            FROM DETRAC JOIN LATERAL EXTRACT_OBJECT(data, Yolo, NorFairTracker)
+                AS T(iids, labels, bboxes, scores)
+            WHERE id < 30;
+            """
+        execute_query_fetch_all(select_query)
+        reuse_batch = execute_query_fetch_all(select_query)
+        self._verify_reuse_correctness(select_query, reuse_batch)
```
xzdandy, line 182: Add a execution time assert.

xzdandy (approved):
@gaurav274 Please help checking the Notebook test case.

gaurav274:
> @gaurav274 Please help checking the Notebook test case.

Sure, looking into it.

gaurav274:
@xzdandy Hopefully, the changes should fix the build-related issue. 
Btw, do you think there is a meaningful test case for object tracking? Also, please review that I accidentally didn't remove your changes.

xzdandy:
> @xzdandy Hopefully, the changes should fix the build-related issue. Btw, do you think there is a meaningful test case for object tracking? Also, please review that I accidentally didn't remove your changes.

That is a good point. We should have a test case for EVAtracker abstract class and builtin nor_fair tracker.

PR #580, opened 2023-02-01
feat: enable loading videos from s3

suryatejreddy:
This PR adds support for loading video datasets from S3. 

### Sample Queries
- `LOAD VIDEO 's3://bucket/dummy.avi' INTO MyVideo;` 
- `LOAD VIDEO 's3://bucket/eva_videos/*.mp4 INTO MyVideos;`

### Design Decisions
- The video files are downloaded to a directory `s3_download_dir` inside the config and by default points to inside the `.eva` folder. 
- This would add redundancy but will be removed in a future PR that removes data duplicity by adding symlinks.

### Testing
- Added test cases for single and multi-video loading


eva/executor/load_multimedia_executor.py
```diff
@@ -41,14 +43,27 @@ def exec(self):
...
+
+            # If it is a s3 path, download the file to local
+            if search_path.as_posix().startswith("s3:/"):
+                s3_dir = Path(
+                    ConfigurationManager().get_value("storage", "s3_download_dir")
+                )
+                dst_path = s3_dir / self.node.table_info.table_name
+                dst_path.mkdir(parents=True, exist_ok=True)
+                download_from_s3(self.node.file_path, dst_path)
+                search_path = dst_path / "*"
```
gaurav274, line 56: Why are we downloading all the files and not just the once that satisfy the regex?

gaurav274, line 56: If we fire multiple load commands on the same table, will it end up loading the same videos multiple times (because we download them in the same directory)? How about we not do this and rather iterate over the downloaded files?

test/util.py
```diff
@@ -337,6 +344,31 @@ def create_dummy_batches(num_frames=NUM_FRAMES, filters=[], batch_size=10, start
         yield Batch(pd.DataFrame(data))
 
 
+def create_dummy_batches_s3(
```
gaurav274, line 347: Maybe make `dir_name` as a function param and reuse an existing function.

test/integration_tests/test_s3_load_executor.py
```diff
@@ -0,0 +1,87 @@
...
+        select_query = """SELECT * FROM MyVideo;"""
+
+        actual_batch = execute_query_fetch_all(select_query)
+        actual_batch.sort()
+        expected_batch = list(create_dummy_batches_s3())[0]
+        self.assertEqual(actual_batch, expected_batch)
+        execute_query_fetch_all("DROP TABLE IF EXISTS MyVideo;")
+
+    def test_s3_multiple_file_load_executor(self):
+        s3_client = boto3.client("s3")
```
gaurav274, line 73: Add a test case with multiple load commands on the same table. Consider a combination of local and s3 paths (consider multiple s3 paths).

gaurav274 (changes requested):

gaurav274:
Add changes to the https://github.com/georgia-tech-db/eva/blob/master/CHANGELOG.md

gaurav274 (approved):

PR #581, opened 2023-02-02
fix: debugging issue in yolov5

Aryan-Rajoria:
Should fix the `models.yolo` not found error found when running debugger on eva.
remove `yolov5-pip` package and instead uses `torch.hub` for yolov5 model.

### DIFF
--- a/eva/udfs/yolo_object_detector.py
+++ b/eva/udfs/yolo_object_detector.py
@@ -44,7 +44,7 @@ class YoloV5(PytorchAbstractClassifierUDF):
 
     def setup(self, threshold=0.85):
         self.threshold = threshold
-        self.model = torch.hub.load('ultralytics/yolov5', 'yolov5s', verbose=False)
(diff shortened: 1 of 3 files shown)


gaurav274 (approved):

eva/udfs/yolo_object_detector.py
```diff
@@ -51,7 +51,7 @@ def name(self) -> str:
 
     def setup(self, threshold=0.85):
         self.threshold = threshold
-        self.model = yolov5.load("yolov5s.pt", verbose=False)
```
gaurav274, line 54: Remove the import yolov5 statement

gaurav274 (changes requested):

gaurav274:
Add changes to the https://github.com/georgia-tech-db/eva/blob/master/CHANGELOG.md

eva/udfs/yolo_object_detector.py
Aryan-Rajoria: done

ashmitaraju (approved):

Issue #563, opened 2023-01-15
Loading non-existent image


jarulraj:
When an image file does not exist and we attempt to LOAD it, there is an error in the subsequent `SELECT` query:

```
file_name = "dog.png"
cursor.execute('LOAD IMAGE "' + file_name + '" INTO MyImages;')
response = cursor.fetch_all()
print(response)

cursor.execute("""SELECT AnimalDetectorPlus(data)
                FROM MyImages""")
response = cursor.fetch_all()
print(response)
```

```
@error: Failed to read image file /root/.eva/0.1.3/eva_datasets/c711ed68a19df9d363d5d3a5f6296af3/63570321
```

PR #583, opened 2023-02-03
feat: Add symlinks to data files instead of copying

suryatejreddy:
This PR adds a major change (with little code). 
### Current Behavior
We copy the dataset file that the user loads into the `eva_datasets` folder. 

### Change
This PR modifies this behavior to add a symlink from `eva_datasets` to the file on the user's filesystem. In case the user later deletes the file, an appropriate error is thrown. 

### Testing 
- Added test cases for individual and multiple file loads.
- Added test case for testing correct exception thrown if the file is removed from user's system.


gaurav274 (commented):
Can you verify this also fixes https://github.com/georgia-tech-db/eva/issues/563? Thanks!

gaurav274:
Okay, so it does say 0 files loaded. We can discuss, what the behavior should be if some files are missing while loading and selecting. Btw, in this case if no file was loaded, why does it say missing file? Shouldn't it just execute with empty output?

suryatejreddy:
> Okay, so it does say 0 files loaded. We can discuss, what the behavior should be if some files are missing while loading and selecting. Btw, in this case if no file was loaded, why does it say missing file? Shouldn't it just execute with empty output?

I think it says missing file during select because the original dataset file doesn't exist. That is the check that I have implemented as of now. I think if files are missing while loading we should raise an immediate error there.

gaurav274 (approved):

gaurav274:
Can we merge this PR, or is there anything remaining?

PR #584, opened 2023-02-09
Rename Array_Count to ArrayCount

LordDarkula:
- `ArrayCount` matches Python naming convention
- This change should not be polluting the UDF registration PR


kaushikravichandran:
Why did just renaming stuff reduce coverage by 0.2%?

kaushikravichandran (approved):
Why did just renaming stuff reduce coverage by 0.2%?
Looks good otherwise.

PR #585, opened 2023-02-09
feat: UDF Decorators

IshSiva:
Adding the feature to create UDFs using Decorators in EVA. 
The decorators can be used for enforcing sanity checks on the input arguments. 

Once it is fully integrated the create udf command can be like
"""
CREATE UDF IF NOT EXISTS YoloObjectDetector
IMPL eva/udfs/yolo_object_detector.py;
"""

- [x] Create separate classes for input types
- [x] Create the format of decorator functions
- [x] Create a sample udf using the decorators and type checking
- [x] Write unit tests for the decorators.
- [x] Update the parser to suit the new udf format


eva/udfs/new_udf_impl/decorators_yolo.py
```diff
@@ -0,0 +1,75 @@
...
+class YoloDecorators(PytorchAbstractClassifierUDF):
+    
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+    
+    @property
+    def name(self) -> str:
+        return "yolo"
+    
+    property
```
LordDarkula, line 32: forgot @

test/udfs/test_udf_service.py
```diff
@@ -0,0 +1,81 @@
+# coding=utf-8
+# Copyright 2018-2022 EVA
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
LordDarkula, line 4: I would like to see some tests

eva/udfs/new_udf_impl/base_udf_fns.py
```diff
@@ -0,0 +1,59 @@
+
+import numpy as np
+import pandas as pd
+import torch
+from facenet_pytorch import MTCNN
```
LordDarkula, line 5: the fundamental problem with this implementation
is that it does not work for user code
the point of a decorator is that the user can decorate their own functions
and then run a SQL command
to ingest the UDF
if you have an if statement that checks the model type in the base_udf_fns.py file
this is not possible

eva/udfs/new_udf_impl/base_udf_fns.py
```diff
@@ -0,0 +1,59 @@
...
+import yolov5
+from eva.models.catalog.frame_info import FrameInfo
+from eva.models.catalog.properties import ColorSpace
+from eva.udfs.abstract.pytorch_abstract_udf import PytorchAbstractClassifierUDF
```
LordDarkula, line 9: I dislike the name of this file. Filenames should not be plural or contain abbreviations

LordDarkula (changes requested):

eva/udfs/new_udf_impl/base_udf_fns.py
IshSiva: This is just a sample file and not the final one. A new file with proper naming conventions and coding practices will be created for the final integration into the eva system

IshSiva:
I have not yet implemented the actual udfs, but wanted to get feedback for the framework for the input types that are being created.

suryatejreddy:
@IshSiva could you add a description in the PR for how the CREATE command would change with decorators?

suryatejreddy (changes requested):
Just quickly skimmed through it, at a broad level:
- You need to find a good place to puts the decorators code, it won't be in `udfs/contrib/decorators/ish` as it is now. Proabably `udfs/setup` would be a good location. 
- I think enforcing the type checking should happen in a different PR. This PR should just replicate the current status of the system using decorators. 
- Could you add a description in the PR and/or documentation on how we would expect users to use the NumpyArray and Pandas datatypes that we have defined. 

Will do a more detailed review later tonight. 

Edit: Added third point above.

eva/parser/create_udf_statement.py
```diff
@@ -71,7 +71,12 @@ def __str__(self) -> str:
                 output_str += str(expr) + ", "
             output_str = output_str.rstrip(", ")
 
-        return f"CREATE UDF {self._name} INPUT ({input_str}) OUTPUT ({output_str}) TYPE {self._udf_type} IMPL {self._impl_path.name}"
+        if len(input_str) > 0:
+            return f"CREATE UDF {self._name} INPUT ({input_str}) OUTPUT ({output_str}) TYPE {self._udf_type} IMPL {self._impl_path.name}"
+        elif len(self._udf_type) > 0:
+            return f"CREATE UDF {self._name} TYPE {self._udf_type} IMPL {self._impl_path.name}"
+        else:
```
suryatejreddy, line 78: nit: Can probably do this in a cleaner way, by adding "INPUT" and "OUTPUT" to the strings defined above. Similarly for type.

eva/udfs/contrib/decorators/ish/io_descriptors/data_types/type_numpy.py
```diff
@@ -0,0 +1,84 @@
...
+                )
+            elif self.dtype == "float32":
+                return isinstance(input_object, np.ndarray) and (
+                    input_object.dtype == np.float32
+                )
+
+        elif not self.dtype:
+            return isinstance(input_object, np.ndarray)
+
+    def convert_data_type(self, input_object: any):
```
suryatejreddy, line 47: When would we need to use this conversion?

eva/udfs/contrib/decorators/ish/io_descriptors/data_types/type_tensor.py
```diff
@@ -0,0 +1,96 @@
...
+                )
+            elif self.dtype == "float32":
+                return isinstance(input_object, torch.Tensor) and (
+                    input_object.dtype == torch.float32
+                )
+
+        elif not self.dtype:
+            return isinstance(input_object, torch.Tensor)
+
+    def convert_data_type(self, input_object: any):
```
suryatejreddy, line 48: Same question as numpy file

eva/udfs/contrib/decorators/ish/io_descriptors/eva_arguments.py
```diff
@@ -0,0 +1,126 @@
...
+class EvaArgument(object):
+    """
+    Base class for the data types that are used inside Eva. This class is inherited by the NumPyArray,
+    PyTorchTensor and PandasDataFrame classes.
+    The functions are implemented in the child classes.
+
+    """
+
+    @abstractmethod
+    def __init__(self, shape=None, dtype=None, columns=None) -> None:
```
suryatejreddy, line 27: This is unclear to me. Shouldn't each column be it's own EvaArgument? And we also need to have a column name for each of these? Further, we need to be able to pass multiple inputs to a UDF.

eva/udfs/contrib/decorators/ish/udf_decorators.py
```diff
@@ -0,0 +1,121 @@
...
+# limitations under the License.
+
+
+from eva.udfs.contrib.decorators.ish.io_descriptors.eva_arguments import EvaArgument
+from eva.utils.errors import TypeException
+
+
+# decorator for the setup function. It will be used to set the cache, batching and
+# udf_type parameters in the catalog
+def setup(use_cache: bool, udf_type: str, batch: bool):
```
suryatejreddy, line 23: Remove the debug print statements. We probably need to move type checking to another PR as discussed in a previous comment.

eva/udfs/contrib/decorators/ish/udf_decorators.py
```diff
@@ -0,0 +1,121 @@
...
+        tags["cache"] = use_cache
+        tags["udf_type"] = udf_type
+        tags["batching"] = batch
+        wrapper.tags = tags
+        return wrapper
+
+    return inner_fn
+
+
+def forward(input_signature: EvaArgument, output_signature: EvaArgument):
```
suryatejreddy, line 45: Probably need to rename `input_signature` to `inputs` and `output_signature` to `outputs`.

eva/utils/errors.py
```diff
@@ -16,6 +16,12 @@ class CatalogError(Exception):
     pass
 
 
+class TypeException(Exception):
```
suryatejreddy, line 19: SHould rename to something more specific, `UDFInputOutputTypeException`?

test/integration_tests/test_pytorch.py
```diff
@@ -116,6 +116,18 @@ def test_should_run_pytorch_and_asl(self):
         for idx in res.index:
             self.assertTrue("computer" in res["aslactionrecognition.labels"][idx])
 
+    @pytest.mark.torchtest
```
suryatejreddy, line 119: Good test!

test/udfs/test_udf_decorators.py
```diff
@@ -0,0 +1,145 @@
...
+)
+def forward_fn_pytorch_input_output_constraint(self_obj, torch_tensor):
+    ans = torch.sum(torch_tensor)
+    ans = ans.to(torch.int32)
+    ans = torch.unsqueeze(ans, 0)
+    ans = torch.unsqueeze(ans, 0)
+    return ans
+
+
+class UdfDecoratorTest(unittest.TestCase):
```
suryatejreddy, line 75: I believe these tests will move to a different PR as all of them do type checking.

suryatejreddy (changes requested):
Thanks @IshSiva. Decorators looks great overall. Left some more comments.

eva/udfs/contrib/decorators/ish/io_descriptors/eva_arguments.py
suryatejreddy: This should probably be the same as https://github.com/georgia-tech-db/eva/blob/076704705c35245a6c83a626dba599342c59ff64/eva/catalog/models/udf_io_catalog.py#L106

eva/udfs/contrib/decorators/ish/io_descriptors/data_types/type_pandas_dataframe.py
```diff
@@ -0,0 +1,41 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import pandas as pd
+
+from eva.udfs.contrib.decorators.ish.io_descriptors.eva_arguments import EvaArgument
+
+
+class PandasDataframe(EvaArgument):
```
suryatejreddy, line 20: We probably need to discuss how to support pandas dataframe as input with multiple columns. In that case it won't be as straightforward as a tensor or numpy array and cannot extend the EVAArgument class.

eva/udfs/contrib/decorators/ish/io_descriptors/data_types/type_numpy.py
IshSiva: In some Pytorch functions a difference in the data type produces completely different results for the same set of inputs. so it is better to convert to a specific data type if explicitly mentioned by the user.

eva/udfs/contrib/decorators/ish/io_descriptors/data_types/type_tensor.py
IshSiva: same as above

IshSiva:
The changes will be split into multiple PRs. New PR is #601

PR #588, opened 2023-02-16
feat: Add UDF Cost Catalog

kaushikravichandran:
This PR adds a UDF cost catalog table that can be used to store the function expression execution time as a cost. In the future, more complex functions can be used to cost UDFs.

PR #595, opened 2023-02-24
doc: Adding Catalog manager documentation

aryan-rajoria:
Documentation for catalog manager

### DIFF
--- a/docs/source/reference/catalog.rst
+++ b/docs/source/reference/catalog.rst
@@ -1,0 +1,31 @@
+EVA Catalog Manager Reference
+=============================
+
+Catalog Manager
+---------------
+
+Explanation for developers on how to use the eva catalog_manager.
+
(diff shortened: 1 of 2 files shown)


kaushikravichandran (commented):
The PNG seems too rigid. Is there a way to visualize and generate it using the sqlalchemy objects in the code? Something like this: `https://github.com/sqlalchemy/sqlalchemy/wiki/SchemaDisplay`?

kaushikravichandran (changes requested):
Could you add information about the `udf_cost_catalog` as well?
Ref: https://github.com/georgia-tech-db/eva/pull/588

gaurav274:
Add script and we should merge this. Thanks!

gaurav274:
> Could you add information about the `udf_cost_catalog` as well? Ref: #588

@kaushikravichandran You can add that as part of your PR. Thanks!

gaurav274 (approved):

PR #599, opened 2023-03-03
feat: EVA x HuggingFace

suryatejreddy:
This PR adds support for HuggingFace via [Pipelines](https://huggingface.co/docs/transformers/main_classes/pipelines).

### Background
The Pipeline API is an easy way to use a HuggingFace model for inference. In EVA, we require users to mention the task they want to perform for simplicity. A HuggingFace task is different from a model(pytorch). There are a large number of models on HuggingFace hub that can be used for a particular task. The user can specify the model or a default model will be used.

### Parser
We differentiate HF UDFs using the Type attribute of a UDF. It should be defined as "HuggingFace". The metadata attributes of the CREATE statement are used to specify/define the HF UDF. We expect "task" to be passed as metadata in the CREATE statement. Users can additionally specify "model" and other attributes specific to the model similarly. 

### Create UDF Executor
Currently, we do not support all tasks defined by HF. The supported tasks are defined in `eva/third_party/huggingface/create.py`. 

All HF UDFs have the same implementation file `udfs/abstract/hf_abstract_udf.py`. The inputs and outputs for the UDF are extracted using the following method:
- Identify the type of input for the UDF. This could be text, image, audio, or video. We store a 1-1 mapping between tasks dn input types. 
- Instantiate a pipeline using the attributes defined by the user (task, model, etc) and pass in a dummy input of correct type to return an output. We use this to identify the names and types of the outputs returned by the model. 
We generate IOCatalogEntry objects for these inputs and outputs to store along with the UDF. 

### Binder
The actual function for HF UDF is bound during the runtime of the function evaluation. We fetch the input to the UDF (generated in the above step). 

The data format for an HF UDF depends on the input type. We define different input types in `eva/third_party/huggingface/model.py`. Each implementation transforms the inputs into a format that can be passed to an HF model. 

### Final Syntax

```
CREATE UDF {udf_name}
TYPE HuggingFace
'task' 'text-classification'
```

```
CREATE UDF {udf_name}
TYPE HuggingFace
'task' 'object-detection'
'model' 'facebook / detr-resnet-50'
```

### Testing
Adds testcases


gaurav274:
@suryatejreddy  Is it ready to review?

eva/binder/statement_binder.py
```diff
@@ -243,40 +244,44 @@ def _bind_func_expr(self, node: FunctionExpression):
...
-        except Exception as e:
-            err_msg = (
-                f"{str(e)}. Please verify that the UDF class name in the"
-                "implementation file matches the UDF name."
-            )
-            logger.error(err_msg)
-            raise BinderError(err_msg)
+        if udf_obj.type == "HuggingFace":
+            node.function = bind_hf_func_from_udf(udf_obj)
+            node.projection_columns = ["*"]
```
gaurav274, line 249: What is this?

eva/utils/huggingface_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    "object-detection": CustomObjectDetection,
+    "text-classification": CustomTextClassification,
+}
+
+
+def bind_hf_func_from_udf(udf_obj: UdfCatalogEntry):
+    """
+    Generate and return pipeline function from UdfCatalogEntry
+    """
+    default_args = {
```
gaurav274, line 84: Do you think we should default to this? Isn't it better to complain if they don't specify any task?

eva/utils/huggingface_utils.py
```diff
@@ -0,0 +1,92 @@
...
+
+def image_data_preprocesser(images):
+    frames_list = images.values.tolist()
+    frames = np.array(frames_list)
+    frames = np.vstack(frames)
+    images = [Image.fromarray(row) for row in frames]
+    return images
+
+
+def data_postprocesser(model_output):
```
gaurav274, line 36: Same as above

eva/models/storage/batch.py
```diff
@@ -247,6 +247,10 @@ def project(self, cols: None) -> Batch:
         We do a copy for now.
         """
         cols = cols or []
+        # The following is a workaround to support dynamic HuggingFace UDF's
+        # There is no concrete method of knowing the columns of the output beforehand
+        if cols == ["*"]:
```
gaurav274, line 252: HuggingFace models cannot be composed in this case.

gaurav274: @xzdandy Any ideas?

gaurav274: Can't we get output based on the task?

docs/source/reference/hf.rst
```diff
@@ -0,0 +1,35 @@
+HuggingFace Models in EVA
+======================
+
+This section provides an overview of how you can use out-of-thebox HuggingFace models in EVA.
```
gaurav274, line 4: "out-of-the box"

docs/source/reference/hf.rst
```diff
@@ -0,0 +1,35 @@
...
+- Text Classification
+- Summarization
+- Text2Text Generation
+- Text Generation
+- Image Classification
+- Image Segmentation
+- Image-to-Text
+- Object Detection
+- Depth Estimation
+- Video Classification
```
gaurav274, line 35: We need to expand this. Maybe in the next PR.

eva/third_party/huggingface/binder.py
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from eva.catalog.catalog_type import NdArrayType
+from eva.catalog.models.udf_catalog import UdfCatalogEntry
+from eva.third_party.huggingface.model import ImageHFModel, TextHFModel
+
+
+def assign_hf_udf(udf_obj: UdfCatalogEntry):
+    # NOTE: Currently supports one input
```
gaurav274, line 21: Please better explain this comment. Thanks!

gaurav274: Add doc string

eva/third_party/huggingface/create.py
```diff
@@ -0,0 +1,208 @@
...
+    return pipe(image=image_input, text=text_input)
+
+
+def infer_output_type(**pipeline_args):
+    """
+    Infer the output type of a pipeline based on the task type.
+    Note: The below code is flaky and depends on the version of transformers.
+    It also hard-codes the type assignment to certain tasks.
+    """
+    if "task" not in pipeline_args:
```
gaurav274, line 104: Convert all raised Exceptions to asserts.

eva/third_party/huggingface/create.py
```diff
@@ -0,0 +1,208 @@
...
+    task = pipeline_args["task"]
+    if task in UNSUPPORTED_TASKS:
+        raise Exception(f"HuggingFace task:{task} not supported in EVA currently")
+
+    task_input = SUPPORTED_TASKS.get(task, {}).get("type", None)
+    if task_input is None:
+        raise Exception("Task Type Could Not Be Inferred")
+
+    pipe = pipeline(**pipeline_args)
+    if task_input == "text":
```
gaurav274, line 115: A lot of hard-coded strings. Will making them enums help? What is HF version specific here?

eva/udfs/abstract/hf_abstract_udf.py
```diff
@@ -0,0 +1,72 @@
...
+        return "GenericHuggingfaceModel"
+
+    def setup(
+        self, udf_obj: UdfCatalogEntry, device: int = -1, *args, **kwargs
+    ) -> None:
+        pipeline_args = {entry.key: entry.value for entry in udf_obj.metadata}
+        self.hf_udf_obj = pipeline(**pipeline_args, device=device)
+
+    @property
+    def input_format(self) -> str:
```
gaurav274, line 42: What is this?

docs/source/reference/hf.rst
suryatejreddy: Yeah the other models require more than one input so might require some more work. Will do that next.

gaurav274: I meant the documentation.

eva/third_party/huggingface/binder.py
```diff
@@ -0,0 +1,38 @@
...
+def assign_hf_udf(udf_obj: UdfCatalogEntry):
+    """
+    Assigns the correct HF Model to the UDF. The model assigned depends on
+    the type of input the UDF expects. This is done so that we can
+    process the input correctly before passing it to the HF model.
+    """
+    inputs = udf_obj.args
+
+    # NOTE: Currently, we only support models that require a single input.
+    input_type = inputs[0].array_type
```
gaurav274, line 29: Do we need an assert here for safety?

docs/source/reference/hf.rst
suryatejreddy: What else do you think we can add? More examples?

eva/third_party/huggingface/create.py
```diff
@@ -0,0 +1,203 @@
...
+from eva.catalog.catalog_type import ColumnType, NdArrayType
+from eva.catalog.models.udf_io_catalog import UdfIOCatalogEntry
+from eva.catalog.models.udf_metadata_catalog import UdfMetadataCatalogEntry
+from eva.third_party.huggingface.model import HFInputTypes
+
+"""
+We currently support the following tasks from HuggingFace.
+"""
+SUPPORTED_TASKS = {
+    "audio-classification": HFInputTypes.AUDIO,
```
gaurav274, line 32: Clean!

eva/third_party/huggingface/create.py
```diff
@@ -0,0 +1,203 @@
...
+    "image-classification": HFInputTypes.IMAGE,
+    "image-segmentation": HFInputTypes.IMAGE,
+    "image-to-text": HFInputTypes.IMAGE,
+    "object-detection": HFInputTypes.IMAGE,
+    "depth-estimation": HFInputTypes.IMAGE,
+    "video-classification": HFInputTypes.VIDEO,
+}
+
+
+def run_pipe_through_text(pipe: Pipeline):
```
gaurav274, line 48: How about we have different functions to create the required data and decouple the `pipe(data)` call? I feel we will be able to clean it further.

eva/third_party/huggingface/create.py
```diff
@@ -0,0 +1,203 @@
...
+    task = pipeline_args["task"]
+    assert task in SUPPORTED_TASKS, f"Task {task} not supported in EVA currently"
+
+    # Construct the pipeline
+    pipe = pipeline(**pipeline_args)
+
+    # Run the pipeline through a dummy input to get a sample output
+    model_input = SUPPORTED_TASKS[task]
+    if model_input == HFInputTypes.TEXT:
+        output = run_pipe_through_text(pipe)
```
gaurav274, line 116: Call `pipe(data)` here

eva/udfs/abstract/hf_abstract_udf.py
```diff
@@ -0,0 +1,79 @@
...
+        result_df = pd.DataFrame(result_list)
+        return result_df
+
+    def forward(self, inputs, *args, **kwargs) -> None:
+        hf_input = self.input_formatter(inputs)
+        hf_output = self.hf_udf_obj(hf_input, *args, **kwargs)
+        eva_output = self.output_formatter(hf_output)
+        return eva_output
+
+    def to_device(self, device: str) -> GPUCompatible:
```
gaurav274, line 78: Did you verify we are able to run it on GPUs?

eva/third_party/huggingface/create.py
suryatejreddy: I initially did it that way, but the reason for not going ahead was for video. I would need a temp file for video and this was the cleanest way of dealing with videos.

docs/source/reference/hf.rst
gaurav274: Yes

eva/udfs/abstract/hf_abstract_udf.py
suryatejreddy: Will cover in next PR with GPU redesign.

docs/source/reference/hf.rst
```diff
@@ -0,0 +1,34 @@
...
+Creating UDF from HuggingFace
+-----
+EVA supports UDFS similar to `Pipelines <https://huggingface.co/docs/transformers/main_classes/pipelines>`_  in HuggingFace. 
+
+.. code-block:: sql
+
+    CREATE UDF IF NOT EXISTS HFObjectDetector
+    TYPE  HuggingFace
+    'task' 'object-detection'
+    'model' 'facebook / detr-resnet-50'
```
xzdandy, line 16: Is this another way to create UDF? Internally it will be translated to how EVA normally creates UDF.

suryatejreddy: Yeah that's the goal of the design, users can use HF UDF's without writing code and EVA treats it similar to any other UDF.

gaurav274 (approved):

PR #601, opened 2023-03-10
feat: Decorators for UDF - Part 1

IshSiva:
Adding the feature to create UDFs using Decorators in EVA.
The decorators can be used for enforcing sanity checks on the input arguments.

Once it is fully integrated the create udf command can be like
"""
CREATE UDF IF NOT EXISTS YoloObjectDetector
IMPL eva/udfs/yolo_object_detector.py;
"""

- [x] Create separate classes for input types
- [x] Create decorators for the setup and forward functions
- [x] Create a custom exception class for UDF IO Errors
- [x] Update parser to suit the new syntax
- [x] Unit tests for decorators
- [x] Inputs for UDF decorators should be lists.


eva/io_descriptors/eva_arguments.py
```diff
@@ -0,0 +1,140 @@
...
+    def is_shape_defined(self):
+        """returns True if the size has been specified. False otherwise"""
+        return True
+
+    def is_dtype_defined(self):
+        """returns True if the dtype has been specified. False otherwise"""
+        return True
+
+
+class IOColumnArgument(object):
```
suryatejreddy, line 117:
```suggestion
class IOColumnArgument(IOArgument):
```

eva/io_descriptors/eva_arguments.py
```diff
@@ -0,0 +1,140 @@
...
+# limitations under the License.
+from abc import abstractmethod
+from typing import Tuple
+
+from eva.catalog.catalog_type import ColumnType, NdArrayType
+
+
+class IOArgument(object):
+    """
+    Base class for the data types that are used inside Eva. This class is inherited by the NumPyArray,
```
suryatejreddy, line 23: The docstring is misleading IMO. This class is purely for interfacing with UDF Input/Outputs. For example Pandas Dataframe option would extend `IOArgument` but is not necessarily a data type that we use inside EVA.

eva/io_descriptors/eva_arguments.py
```diff
@@ -0,0 +1,140 @@
...
+        """returns True if the size has been specified. False otherwise"""
+        return True
+
+    def is_dtype_defined(self):
+        """returns True if the dtype has been specified. False otherwise"""
+        return True
+
+
+class IOColumnArgument(object):
+    """ """
```
suryatejreddy, line 118: Add a docstring here explaining what this would be sued for, essentially for defining types pertaining to an individual column/field.

eva/io_descriptors/eva_arguments.py
```diff
@@ -0,0 +1,140 @@
+# coding=utf-8
```
suryatejreddy, line 1: nit: Can skip `eva_` from filenames.

eva/io_descriptors/eva_data_types.py
```diff
@@ -0,0 +1,144 @@
+# coding=utf-8
```
suryatejreddy, line 1: Same comment as above regarding `eva_` in the filename.

eva/io_descriptors/eva_data_types.py
```diff
@@ -0,0 +1,144 @@
...
+    ) -> None:
+        super().__init__(
+            name=name,
+            type=ColumnType.NDARRAY,
+            is_nullable=is_nullable,
+            array_type=type,
+            array_dimensions=dimensions,
+        )
+
+    def check_type(self, input_object) -> bool:
```
suryatejreddy, line 43:
```suggestion
    def is_valid_type(self, input_object) -> bool:
```
nit: easier naming

eva/io_descriptors/eva_data_types.py
```diff
@@ -0,0 +1,144 @@
...
+                )
+            elif self.dtype == "float32":
+                return isinstance(input_object, np.ndarray) and (
+                    input_object.dtype == np.float32
+                )
+
+        elif not self.dtype:
+            return isinstance(input_object, np.ndarray)
+
+    def check_shape(self, input_object) -> bool:
```
suryatejreddy, line 61:
```suggestion
    def is_valid_shape(self, input_object) -> bool:
```

eva/io_descriptors/eva_arguments.py
```diff
@@ -0,0 +1,140 @@
...
+    The functions are implemented in the child classes.
+
+    """
+
+    @abstractmethod
+    def __init__(self) -> None:
+        pass
+
+    @abstractmethod
+    def check_type(self, input_object: any) -> bool:
```
suryatejreddy, line 34: Docstring descritpion seems incomplete.

eva/io_descriptors/eva_arguments.py
```diff
@@ -0,0 +1,140 @@
...
+            input_object (any): the object whose type should be checked. The required type is given in the constructor.
+
+        Returns:
+            bool: True if the type of input_object matches the required type. False otherwise.
+
+        """
+        pass
+
+    @abstractmethod
+    def check_shape(self, input_object: any, required_shape: tuple = None) -> bool:
```
suryatejreddy, line 47: Same as above: docstring description is incomplete.

eva/io_descriptors/eva_data_types.py
```diff
@@ -0,0 +1,144 @@
...
+            if input_object.shape != self.shape:
+                return False
+
+        return True
+
+    def arg_name(self):
+        return "NumpyArray"
+
+
+class PyTorchTensor(IOColumnArgument):
```
suryatejreddy, line 72: Same suggestions as above re: naming functions. Also needs to change in Abstract Class.

eva/io_descriptors/eva_data_types.py
```diff
@@ -0,0 +1,144 @@
...
+class PandasDataframe(IOArgument):
+    """EVA data type for Pandas Dataframe"""
+
+    def __init__(self, columns, shape, enforce_columns, enforce_shape) -> None:
+        super().__init__()
+        self.columns = columns
+
+        self.type = ColumnType.NDARRAY
+        self.array_type = NdArrayType.ANYTYPE
+        self.is_nullable = False
```
suryatejreddy, line 128: `type`, `array_type`, and `is_nullable` do not apply to a dataframe IMO. We probably need a function that converts the entire dataframe to a list of NumpyArguments assigning the correct class.

eva/parser/create_udf_statement.py
```diff
@@ -71,7 +71,12 @@ def __str__(self) -> str:
                 output_str += str(expr) + ", "
             output_str = output_str.rstrip(", ")
 
-        return f"CREATE UDF {self._name} INPUT ({input_str}) OUTPUT ({output_str}) TYPE {self._udf_type} IMPL {self._impl_path.name}"
+        if len(input_str) > 0:
```
suryatejreddy, line 74: nit: Can probably do this in a cleaner way, by adding "INPUT" and "OUTPUT" to the strings defined above. Similarly for type.

eva/udfs/decorators/udf_decorators.py
```diff
@@ -0,0 +1,66 @@
...
+
+from eva.io_descriptors.eva_arguments import IOArgument
+
+
+# decorator for the setup function. It will be used to set the cache, batching and
+# udf_type parameters in the catalog
+def setup(use_cache: bool, udf_type: str, batch: bool):
+    def inner_fn(arg_fn):
+        def wrapper(*args, **kwargs):
+            # TODO set the batch and caching parameters. update in catalog
```
suryatejreddy, line 27: This will happen inside the create udf executor. All we need from here is the tags to be updated.

eva/udfs/decorators/udf_decorators.py
```diff
@@ -0,0 +1,66 @@
...
+        output_signature (EvaArgument): _description_
+    """
+
+    def inner_fn(arg_fn):
+        def wrapper(*args):
+            frames = args[1]
+
+            # first argument is self and second is frames.
+            output = arg_fn(args[0], frames)
+            return output
```
suryatejreddy, line 58:
```suggestion
            return arg_fn(*args)
```

eva/udfs/decorators/utils.py
```diff
@@ -0,0 +1,64 @@
...
+            result_list.append(
+                UdfIOCatalogEntry(
+                    io.name,
+                    io.type,
+                    io.is_nullable,
+                    array_type=io.array_type,
+                    array_dimensions=io.array_dimensions,
+                    is_input=is_input,
+                )
+            )
```
suryatejreddy, line 63: This could be cleaned up by adding a function `get_catalog_entries` in the abstract class. Each implementation would return it's own catalog entry. The Pandas dataframe would then iterate over it's columns for this.

suryatejreddy (changes requested):
Thanks @IshSiva ! Left some comments mostly around the pandas dataframe input output. Apart from these we need also need some test cases:
- Verify that input and output are updated correctly in the catalog when defining a UDF with decorators.
- Improving coverage (right now, the coverage has dropped). You can check the files that have changed.

suryatejreddy:
We also need an update to the documentation describing decorators and how to use them.

gaurav274 (approved):

eva/udfs/decorators/io_descriptors/data_types.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, columns, column_types=[], column_shapes=[]) -> None:
+        super().__init__()
+        self.columns = columns
+        self.column_types = column_types
+        self.column_shapes = column_shapes
+
+    def generate_catalog_entries(self, is_input=False) -> List[Type[UdfIOCatalogEntry]]:
+        catalog_entries = []
+
+        for column_name, column_type, column_shape in zip(
```
gaurav274, line 75: We need to gracefully handle cases where column count, number of types and shapes don't match

eva/udfs/decorators/udf_decorators.py
```diff
@@ -0,0 +1,66 @@
...
+
+from eva.udfs.decorators.io_descriptors.abstract_types import IOArgument
+
+
+def setup(use_cache: bool, udf_type: str, batch: bool):
+    """decorator for the setup function. It will be used to set the cache, batching and
+    udf_type parameters in the catalog
+
+    Args:
+        use_cache (bool): True if the udf should be cached
```
gaurav274, line 27: Change it to `cachable: True if the udf outcomes can be cached.`

eva/udfs/decorators/udf_decorators.py
```diff
@@ -0,0 +1,66 @@
...
+        udf_type (str): Type of the udf
+        batch (bool): True if the udf should be batched
```
gaurav274, line 29: `batchable: True if the udf supports batched inputs.`

test/util.py
```diff
@@ -484,3 +487,47 @@ def _extract_feature(row: pd.Series):
...
+class DummyObjectDetectorDecorators(AbstractClassifierUDF):
+    @decorators.setup(cachable=True, udf_type="object_detection", batchable=True)
+    def setup(self, *args, **kwargs):
+        pass
+
+    @property
+    def name(self) -> str:
+        return "DummyObjectDetectorDecorators"
+
+    @property
```
gaurav274, line 501: We don't need this anymore, right?

eva/udfs/decorators/yolo_object_detection_decorators.py
```diff
@@ -0,0 +1,207 @@
...
+    def name(self) -> str:
+        return "yolo"
+
+    @setup(cachable=True, udf_type="object_detection", batchable=True)
+    def setup(self, threshold=0.85):
+        self.threshold = threshold
+        self.model = torch.hub.load("ultralytics/yolov5", "yolov5s", verbose=False)
+
+    @property
+    def input_format(self) -> FrameInfo:
```
gaurav274, line 54: We don't need this anymore, right?

gaurav274: If it fails because it is an abstract method, you can get rid of it from the base class itself.

gaurav274 (commented):
Can we get rid of `input_format`?

suryatejreddy:
> Can we get rid of `input_format`?

Done!

PR #607, opened 2023-03-15
feat: Decorators - part 2

IshSiva:
This PR has the code to perform type checking and shape checking in the decorators. It also has the documentation for using decorators in the EVA UDFs.


docs/source/reference/udf_decorators.rst
```diff
@@ -0,0 +1,153 @@
...
+
+1. Create a new file under `udfs/` folder and give it a descriptive name. eg: `yolo_object_detection.py`. 
+
+.. note::
+
+    UDFs packaged along with EVA are located inside the `udfs <https://github.com/georgia-tech-db/eva/tree/master/eva/udfs>`_ folder.
+
+2. Create a Python class that inherits from `PytorchClassifierAbstractUDF`.
+
+* The `PytorchClassifierAbstractUDF` is a parent class that defines and implements standard methods for model inference.
```
gaurav274, line 21: Why are we introducing `PytorchClassifierAbstractUDF` upfront? How about we break it into using `AbstractUDF` etc.? And it is not a classification job so we should also change it in the system. 
We had a bad design earlier, let us carefully fix most of it now.

eva/udfs/decorators/decorators.py
```diff
@@ -64,3 +103,22 @@ def wrapper(*args):
         return wrapper
 
     return inner_fn
+
+
+# #if data type has been specified
```
gaurav274, line 108: Remove this. Thanks!

docs/source/reference/udf_decorators.rst
```diff
@@ -0,0 +1,153 @@
+User-Defined Functions
```
gaurav274, line 1: How about creating a new PR with only the documentation?

gaurav274:
Create a new PR with only the documentation. Thanks!

gaurav274:
We need more descriptive documentation. 
1. Talk about supported data types 
2. Talk about creating Pytorch Based UDF
3. Talk about generic UDFs like `easyocr`

docs/source/reference/udf_decorators.rst
IshSiva: Do we want to do this in this PR or a different one?

eva/udfs/decorators/decorators.py
IshSiva: done

eva/udfs/decorators/decorators.py
xzdandy, line 1: How many overhead will this bring? Are we doing the input and output check on every forward call?

IshSiva: Yes we will be checking it on every forward call. 
I don't think there will be a lot of overhead as we are using the library functions and just doing basic shape and type checking.

xzdandy: If we can have some numbers, that will be great. Do you know why we are not considering a more static type checking?

jarulraj: Instead of checking on every `forward` call, we could optimistically assume that once the first `forward` works, the remaining `forward` calls will be on similar data shape etc. Just a boolean variable for the first check (self.check) might be sufficient for this.

test/udfs/decorators/test_decorators.py
```diff
@@ -41,20 +50,234 @@ def setup_func():
...
+            ],
+        )
+        def forward_func_shape_matched(self, input_obj):
+            ans = torch.sum(input_obj, axis=1)
+            ans = torch.unsqueeze(ans, 1)
+            return ans
+
+        input_object_1 = np.ones((4, 3))
+
+        with self.assertRaises(UDFIODefinitionError):
```
jarulraj, line 246: Good test cases.

jarulraj (approved):

eva/udfs/decorators/decorators.py
gaurav274: We can have it as input from the user (`check_inputs`). And default to always doing it.

IshSiva: added this parameter

eva/udfs/decorators/decorators.py
```diff
@@ -54,8 +55,47 @@ def forward(input_signatures: List[IOArgument], output_signatures: List[IOArgume
...
+                    if len(args) > 0:
+                        args_lst.append(args[0])
+
+                    for i, input_signature in enumerate(input_signatures):
+                        try:
+                            args_lst.append(
+                                # the first object in the forward function is self so we start from the second object
+                                input_signature.validate_object(args[i + 1], True)
+                            )
+                        except UDFIODefinitionError as e:
```
suryatejreddy, line 72: I think it's better to use the assert style syntax that we have adopted in other parts of the code recently. @gaurav274 what do you think?

eva/udfs/decorators/decorators.py
```diff
@@ -54,8 +55,47 @@ def forward(input_signatures: List[IOArgument], output_signatures: List[IOArgume
 
     def inner_fn(arg_fn):
         def wrapper(*args):
+
+            # checking the user constraints
+            if check_input:
+                if len(input_signatures) > 0:
```
suryatejreddy, line 61: nit: Isn't `input_signatures` a list by default? I think we can cleanup the code by just looping over the input_signatures.

eva/udfs/decorators/decorators.py
```diff
@@ -54,8 +55,47 @@ def forward(input_signatures: List[IOArgument], output_signatures: List[IOArgume
...
+
+                    if len(args_lst) > 0:
+                        args = tuple(args_lst)
+
             # calling the forward function defined by the user inside the udf implementation
-            return arg_fn(*args)
+            output = arg_fn(*args)
+
+            if check_output:
+                if len(output_signatures) == 1:
```
suryatejreddy, line 82: nit: Same as input_signatures, can cleanup.

eva/udfs/decorators/io_descriptors/abstract_types.py
```diff
@@ -41,6 +42,47 @@ def generate_catalog_entries(
...
+        pass
+
+    @abstractmethod
+    def check_array_and_convert_type(self, input_obj, data_type):
+        pass
+
+    def is_array_type_defined(self):
+        if self.array_type is None:
+            return False
+        return True
```
suryatejreddy, line 56: nit:
```suggestion
        return self.array_type is not None
```

eva/udfs/decorators/io_descriptors/abstract_types.py
```diff
@@ -41,6 +42,47 @@ def generate_catalog_entries(
...
+
+    def is_shape_defined(self):
+        if self.array_dimensions is None:
+            return False
+        return True
```
suryatejreddy, line 61:
```suggestion
        return self.array_dimensions is not None:
```

eva/udfs/decorators/io_descriptors/abstract_types.py
```diff
@@ -41,6 +42,47 @@ def generate_catalog_entries(
...
+    def validate_object(self, object, input_flag=True):
+        if self.is_array_type_defined:
+            try:
+                object = self.check_array_and_convert_type(object)
+            except Exception as e:
+                if input_flag:
+                    msg = "Data type mismatch of Input parameter. "
+                else:
+                    msg = "Data type mismatch of Output parameter."
+                raise UDFIODefinitionError(msg)
```
suryatejreddy, line 72: Moving to assert syntax will mostly cleanup code by avoiding try and except everywhere.

suryatejreddy: The assert statement can have the appropriate message.

eva/udfs/decorators/io_descriptors/abstract_types.py
```diff
@@ -41,6 +42,47 @@ def generate_catalog_entries(
...
+            try:
+                object = self.check_array_and_convert_shape(object)
+            except Exception as e:
+                if input_flag:
+                    msg = "Shape mismatch of Input parameter. "
+                else:
+                    msg = "Shape mismatch of Output parameter."
+                raise UDFIODefinitionError(msg)
+
+        return object
```
suryatejreddy, line 84: Same as above.

eva/udfs/decorators/io_descriptors/data_types.py
```diff
@@ -41,6 +44,46 @@ def __init__(
             array_dimensions=dimensions,
         )
 
+    def check_array_and_convert_shape(self, input_object):
```
suryatejreddy, line 47: Do we want to be converting data on our own? Or should that be an additional flag? For example user specifies that all data should be converted to this the given IOSignature format. We assert if it isn't possible

eva/udfs/decorators/io_descriptors/data_types.py
```diff
@@ -60,6 +103,43 @@ def __init__(
...
+
+        if not isinstance(input_object, torch.Tensor):
+            if isinstance(input_object, list):
+                input_object = torch.Tensor(input_object)
+            elif isinstance(input_object, np.ndarray):
+                input_object = torch.from_numpy(input_object)
+            else:
+                raise UDFIODefinitionError(
+                    "Unknown data type. Only input types List and Tensor can be converted to Numpy array"
+                )
```
suryatejreddy, line 126: Same as above.

suryatejreddy (changes requested):
Added some comments, mostly looks good to me. Thanks @IshSiva !

eva/udfs/decorators/decorators.py
IshSiva: since converting the shape or data type might throw an exception I am using the try-except

eva/udfs/decorators/decorators.py
IshSiva: removed this.

eva/udfs/decorators/io_descriptors/abstract_types.py
IshSiva: same response as before, the shape conversions might throw an error if its not possible, hence used try-except

eva/udfs/decorators/io_descriptors/data_types.py
IshSiva: I feel that it is redundant to add another flag given that we have check_input and check_output flags.

gaurav274:
@suryatejreddy Please help review this.

jarulraj:
@IshSiva mentioned that we could close this PR and revert the `Decorators` in the documentation. 

@gaurav274 @xzdandy Does that sound good?

xzdandy:
> @IshSiva mentioned that we could close this PR and revert the `Decorators` in the documentation.
> 
> @gaurav274 @xzdandy Does that sound good?

Sounds good, We will redesign in #1017 .

PR #618, opened 2023-03-20
feat: Predicate UDF Reordering Optimization

gaurav274:
1. Reorders the conjuncts based on the cost of the UDFs 
2. Bug fix: orderby executor did not safely handle corner case if the batch is empty.
3. Bug fix: TupleValue Expression did not respect the "mask" used by short-circuiting leading to wrong results.


eva/constants.py
```diff
@@ -17,3 +17,4 @@
 NO_GPU = -1
 UNDEFINED_GROUP_ID = -1
 IFRAMES = "IFRAMES"
+DEFAULT_FUNCTION_EXPRESSION_COST = 100
```
jarulraj, line 20: In a future PR, we should look at why we have two `constants.py` files.

eva/optimizer/rules/rules.py
```diff
@@ -499,6 +503,64 @@ def apply(self, before: LogicalJoin, context: OptimizerContext):
...
+
+    def apply(self, before: LogicalFilter, context: OptimizerContext):
+        # Decompose the expression tree into a list of conjuncts
+        conjuncts = to_conjunction_list(before.predicate)
+
+        # Segregate the conjuncts into simple and function expressions
+        contains_func_exprs = []
+        simple_exprs = []
+        for conjunct in conjuncts:
+            if list(conjunct.find_all(FunctionExpression)):
```
jarulraj, line 535: Do we have a test case where there are no `FunctionExpression` in `conjunct` -- edge case?

test/benchmark_tests/test_benchmark_pytorch.py
```diff
@@ -80,7 +80,7 @@ def test_should_run_pytorch_and_resnet50(benchmark, setup_pytorch_tests):
     # non-trivial test case for Resnet50
     res = actual_batch.frames
     assert res["featureextractor.features"][0].shape == (1, 2048)
-    assert res["featureextractor.features"][0][0][0] > 0.3
+    # assert res["featureextractor.features"][0][0][0] > 0.3
```
jarulraj, line 83: why was this commented out?

test/integration_tests/test_optimizer_rules.py
```diff
@@ -123,3 +125,96 @@ def test_should_pushdown_without_pushdown_join_rule(self):
...
         self.assertEqual(result_without_pushdown_join_rule, result_with_rule)
         self.assertEqual(query_plan, query_plan_without_pushdown_join_rule)
+
+    @patch("eva.catalog.catalog_manager.CatalogManager.get_udf_cost_catalog_entry")
+    def test_should_reorder_predicate(self, mock):
+        def side_effect_func(name):
+            if name == "DummyMultiObjectDetector":
+                return MagicMock(cost=10)
+            else:
+                return MagicMock(cost=5)
```
jarulraj, line 135: Should we have a test case for a scenario where there is no cost in the catalog? (defaults to `DEFAULT_FUNCTION_EXPRESSION_COST`)

jarulraj (approved):

test/benchmark_tests/test_benchmark_pytorch.py
gaurav274: I changed the feature extractor to use a newer model.

test/integration_tests/test_optimizer_rules.py
gaurav274: Done!

eva/constants.py
gaurav274: Sure

eva/optimizer/rules/rules.py
```diff
@@ -499,6 +503,64 @@ def apply(self, before: LogicalJoin, context: OptimizerContext):
         yield new_join
 
 
+class ReorderPredicates(Rule):
```
jiashenC, line 506: Should we give it a more specific name like `CostBasedReorder...` to differentiate with `SelectivityBasedReorder`?

eva/optimizer/rules/rules.py
```diff
@@ -499,6 +503,64 @@ def apply(self, before: LogicalJoin, context: OptimizerContext):
...
+        return len(list(before.predicate.find_all(FunctionExpression))) > 0
+
+    def apply(self, before: LogicalFilter, context: OptimizerContext):
+        # Decompose the expression tree into a list of conjuncts
+        conjuncts = to_conjunction_list(before.predicate)
+
+        # Segregate the conjuncts into simple and function expressions
+        contains_func_exprs = []
+        simple_exprs = []
+        for conjunct in conjuncts:
```
jiashenC, line 534: Just make sure I understand this correctly, if  we have something like `OR(AND(a,b), c)`, this reorder rule won't be able to flip any order inside `OR` right?

PR #620, opened 2023-03-21
feat: Add support for AudioStorageEngine

tushar-97:
- [X] Add AVTableRef which tells the storage engine how to read audio/video files 
- [ ] Add test files

### DIFF
--- a/eva/binder/binder_utils.py
+++ b/eva/binder/binder_utils.py
@@ -20,6 +20,8 @@ from typing import TYPE_CHECKING, List
 from eva.catalog.catalog_type import TableType
 from eva.catalog.catalog_utils import is_video_table
 from eva.catalog.sql_config import IDENTIFIER_COLUMN
+from eva.expression.function_expression import FunctionExpression
+from eva.parser.select_statement import SelectStatement
 
 if TYPE_CHECKING:
     from eva.binder.statement_binder_context import StatementBinderContext
@@ -27,7 +29,7 @@ if TYPE_CHECKING:
 from eva.catalog.catalog_manager import CatalogManager
 from eva.catalog.models.table_catalog import TableCatalogEntry
 from eva.expression.tuple_value_expression import TupleValueExpression
-from eva.parser.table_ref import TableInfo, TableRef
+from eva.parser.table_ref import AVTableRef, ImageTableRef, TableInfo, TableRef
 from eva.utils.logging_manager import logger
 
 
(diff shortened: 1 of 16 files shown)


eva/binder/binder_utils.py
```diff
@@ -107,3 +109,60 @@ def check_table_object_is_video(table_ref: TableRef) -> None:
...
+    if node.from_table.is_table_atom():
+        has_audio_tuple = False
+        has_video_tuple = False
+        if node.target_list:
+            for expr in node.target_list:
+                if isinstance(expr, FunctionExpression):
+                    for child in expr.children:
+                        if isinstance(child, TupleValueExpression):
+                            has_audio_tuple = (
+                                has_audio_tuple or child.col_object.name == "audio"
```
gaurav274, line 137: We should not hard-code strings. Rather get them from the table.

eva/binder/statement_binder.py
```diff
@@ -131,6 +133,8 @@ def _bind_select_statement(self, node: SelectStatement):
                 node.target_list = extend_star(self._binder_context)
             for expr in node.target_list:
                 self.bind(expr)
+        node.from_table = get_table_ref_subclass(node.from_table)
+        set_audio_video_tuple_flags(node)
```
gaurav274, line 137: Move this logic to the end of this function. There might be column dependencies in `order_by` and `group_by` clause.

eva/readers/decord_reader.py
```diff
@@ -121,6 +132,45 @@ def _read(self) -> Iterator[Dict]:
                             "seconds": math.floor(
                                 video.get_frame_timestamp(frame_id)[0]
                             ),
+                            "audio": np.empty(0),
                         }
                     else:
                         break
+
+    def get_av_frames(self) -> Iterator[Dict]:
+        decord = _lazy_import_decord()
```
gaurav274, line 141: @suryatejreddy Please help review this logic.

eva/readers/decord_reader.py
```diff
@@ -121,6 +132,45 @@ def _read(self) -> Iterator[Dict]:
...
+        if self._sampling_type is not None:
+            err_msg = "Sampling type not supported with audio based queries"
+            logger.exception(err_msg)
+            raise Exception(err_msg)
+        if self._sampling_rate != 1:
+            err_msg = "Sampling rate not supported with audio based queries"
+            logger.exception(err_msg)
```
gaurav274, line 148: Heads up: We have moved to a newer design for error handling, where we don't raise exceptions but rather use asserts. Have a look at the master codebase for references. Thanks!

eva/readers/decord_reader.py
gaurav274: Thanks!

eva/parser/table_ref.py
```diff
@@ -252,3 +252,104 @@ def __eq__(self, other):
 
     def __hash__(self) -> int:
         return hash((self._ref_handle, self.alias, self.sample_freq, self.sample_type))
+
+
+class AVTableRef(TableRef):
```
gaurav274, line 257: While you are working on this rewrite, also think if we can improve `TableRef` class. For example, can we remove unnecessary complexity? Thanks!

eva/binder/binder_utils.py
```diff
@@ -107,3 +109,60 @@ def check_table_object_is_video(table_ref: TableRef) -> None:
     if not is_video_table(table_ref.table.table_obj):
         err_msg = "GROUP BY only supported for video tables"
         raise BinderError(err_msg)
+
+
+def get_table_ref_subclass(table_ref: TableRef) -> TableRef:
+    if table_ref.is_table_atom():
```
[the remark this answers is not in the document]
tushar-97: I tried but couldn't get it to work.Since we are changing the same node object which is being passed as an argument, it will work only if I assign the child object to something. Which is why I do it in the bind_select_stmnt method.

eva/binder/binder_utils.py
```diff
@@ -107,3 +109,60 @@ def check_table_object_is_video(table_ref: TableRef) -> None:
...
+            return ImageTableRef(table_ref)
+    elif table_ref.is_join():
+        table_ref.join_node.left = get_table_ref_subclass(table_ref.join_node.left)
+        table_ref.join_node.right = get_table_ref_subclass(table_ref.join_node.right)
+
+    return table_ref
+
+
+def set_audio_video_tuple_flags(node: SelectStatement):
+    if node.from_table.is_table_atom():
```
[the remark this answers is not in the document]
tushar-97: Done

test/util.py
```diff
@@ -285,6 +285,7 @@ def create_dummy_batches(
                     np.ones((2, 2, 3)) * float(i + 1) * 25, dtype=np.uint8
                 ),
                 "myvideo.seconds": 0.0,
+                "myvideo.audio": np.empty(0),
```
tushar-97, line 288: Ever since I added this, a few test cases have been failing. Spent a lot of time, couldn't figure out whats wrong. If you figure out why, do let me know

jarulraj:
@tushar-97 Why don't we by default use `AVReader` for all video files -- what's the need for `read_audio` and `read_video` flags? Do you have a SQL query in mind to test this -- like `LOAD VIDEO... with audio` or `LOAD AUDIO`?

jarulraj:
Closing this PR. See #623

PR #621, opened 2023-03-23
feat: Ray integration

jiashenC:
* Auto-detect availability of GPUs.
* Enable CircleCI auto-testing for Ray execution.
  * Only test CPU Ray on CircleCI.
  * Disable coverage report where it causes long hang for multi-process execution. 
* Address integration test failures for Ray. Here are a list of causes that I address in this PR.
  * Bug in Ray optimizer that does not correctly yield, which often results infinite loop in optimization step.
  * A tree-like execution plan was not previously supported for Ray, mainly because the queue population does not support that yet. For a quick merge now, I do a post-processing to disable Exchange plan for all execution plans that have branch.

### DIFF
--- a/eva/executor/execution_context.py
+++ b/eva/executor/execution_context.py
@@ -50,7 +50,7 @@ class Context:
     def _populate_gpu_from_env(self) -> List:
         # Populate GPU IDs from env variable.
         gpu_conf = map(
-            lambda x: x.strip(), os.environ.get("GPU_DEVICES", "").strip().split(",")
+            lambda x: x.strip(), os.environ.get("CUDA_VISIBLE_DEVICES", "").strip().split(",")
         )
         gpu_conf = list(filter(lambda x: x, gpu_conf))
         gpu_conf = [int(gpu_id) for gpu_id in gpu_conf]
--- a/eva/executor/execution_context.py
+++ b/eva/executor/execution_context.py
@@ -44,7 +44,6 @@ class Context:
     def _populate_gpu_from_config(self) -> List:
         # Populate GPU IDs from yaml config file.
         gpu_conf = self._config_manager.get_value("executor", "gpu_ids")
-        gpu_conf = gpu_conf if gpu_conf else []
         availble_gpus = [i for i in range(get_gpu_count())]
         return list(set(availble_gpus) & set(gpu_conf))
 
@@ -56,7 +55,7 @@ class Context:
         gpu_conf = list(filter(lambda x: x, gpu_conf))
         gpu_conf = [int(gpu_id) for gpu_id in gpu_conf]
         availble_gpus = [i for i in range(get_gpu_count())]
-        return list(set(gpu_conf) & set(availble_gpus))
+        return list(set(availble_gpus) & set(gpu_conf))
 
     def _populate_gpu_ids(self) -> List:
         # Populate available GPU IDs from Torch library. Then, select subset of GPUs
--- a/eva/executor/execution_context.py
+++ b/eva/executor/execution_context.py
@@ -19,7 +19,7 @@ from typing import List, Set
 
 from eva.configuration.configuration_manager import ConfigurationManager
 from eva.constants import NO_GPU
-from eva.utils.generic_utils import is_gpu_available
+from eva.utils.generic_utils import is_gpu_available, get_gpu_count
 
 
 class Context:
@@ -42,30 +42,26 @@ class Context:
     def gpus(self):
         return self._gpus
 
-    def _possible_addresses(self) -> Set:
-        host = socket.gethostname()
-        result_address = {host}
-        true_host, aliases, address = socket.gethostbyaddr(host)
-        result_address.add(true_host)
-        result_address.update(set(aliases).union(set(address)))
-        return result_address
-
     def _populate_gpu_from_config(self) -> List:
-        gpu_conf = self._config_manager.get_value("executor", "gpus")
-        gpu_conf = gpu_conf if gpu_conf else {}
-        this_address = self._possible_addresses()
-        intersection_addresses = this_address.intersection(gpu_conf.keys())
-        if len(intersection_addresses) != 0:
-            return [str(gpu) for gpu in gpu_conf.get(intersection_addresses.pop())]
-        return []
+        # Populate GPU IDs from yaml config file.
+        gpu_conf = self._config_manager.get_value("executor", "gpu_ids")
+        gpu_conf = gpu_conf if gpu_conf else []
+        availble_gpus = [i for i in range(get_gpu_count())]
+        return list(set(availble_gpus) & set(gpu_conf))
 
     def _populate_gpu_from_env(self) -> List:
-        gpus = map(
+        # Populate GPU IDs from env variable.
+        gpu_conf = map(
             lambda x: x.strip(), os.environ.get("GPU_DEVICES", "").strip().split(",")
         )
-        return list(filter(lambda x: x, gpus))
+        gpu_conf = list(filter(lambda x: x, gpu_conf))
+        gpu_conf = [int(gpu_id) for gpu_id in gpu_conf]
+        availble_gpus = [i for i in range(get_gpu_count())]
+        return list(set(gpu_conf) & set(availble_gpus))
 
     def _populate_gpu_ids(self) -> List:
+        # Populate available GPU IDs from Torch library. Then, select subset of GPUs
+        # from available GPUs based on user configuration.
         if not is_gpu_available():
             return []
         gpus = self._populate_gpu_from_config()
--- a/eva/executor/execution_context.py
+++ b/eva/executor/execution_context.py
@@ -14,12 +14,11 @@
 # limitations under the License.
 import os
 import random
-import socket
-from typing import List, Set
+from typing import List
 
 from eva.configuration.configuration_manager import ConfigurationManager
 from eva.constants import NO_GPU
-from eva.utils.generic_utils import is_gpu_available, get_gpu_count
+from eva.utils.generic_utils import get_gpu_count, is_gpu_available
 
 
 class Context:
--- a/eva/executor/execution_context.py
+++ b/eva/executor/execution_context.py
@@ -50,7 +50,8 @@ class Context:
     def _populate_gpu_from_env(self) -> List:
         # Populate GPU IDs from env variable.
         gpu_conf = map(
-            lambda x: x.strip(), os.environ.get("CUDA_VISIBLE_DEVICES", "").strip().split(",")
+            lambda x: x.strip(),
+            os.environ.get("CUDA_VISIBLE_DEVICES", "").strip().split(","),
         )
         gpu_conf = list(filter(lambda x: x, gpu_conf))
         gpu_conf = [int(gpu_id) for gpu_id in gpu_conf]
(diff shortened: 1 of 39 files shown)


gaurav274 (commented):
Can you enable `circleci` to run with `ray`? It is fine if it does not pass. We can do for `python 3.10` version only.

jiashenC:
> Can you enable `circleci` to run with `ray`? It is fine if it does not pass. We can do for `python 3.10` version only.

@gaurav274 I enable the circleci. It does not have GPU resource, so we need to change that as well.

xzdandy:
I think lateral join becomes linear now. So ray should be able to support function scan without worrying about tree-like execution plan.

eva/experimental/ray/optimizer/rules/rules.py
```diff
@@ -104,7 +104,8 @@ def apply(self, before: LogicalGet, context: OptimizerContext):
         if before.target_list is None or not any(
             [isinstance(expr, FunctionExpression) for expr in before.target_list]
         ):
-            return scan
+            yield scan
```
xzdandy, line 107: If there are no UDFs, we probably don't need ray. How about move this logic to the `check` function?

eva/experimental/ray/executor/exchange_executor.py
```diff
@@ -60,25 +62,35 @@ def __init__(self, node: ExchangePlan):
...
+        # Traverse the query execution tree in a DFS manner. This
+        # function will return the parent executor of ExchangeExecutor,
+        # which will later be inserted a queue reader.
+
+        # Start with checking if current executor has any child.
+        if len(curr_exec.children) > 0:
+            # Check if child executor is ExchangeExecutor. If so, return.
+            # Otherwise, traverse further down.
+            if isinstance(curr_exec.children[0], ExchangeExecutor):
+                yield curr_exec
```
xzdandy, line 75: I feel this does not work with tree structure. Consider the following:
```mermaid
graph TD;
    R[EXCHANGE] --> A;
    A[JOIN]-->B[EXCHANGE];
    A-->C[EXCHANGE];
    B --> D[SOME UDF1];
    C --> E[SOME UDF2];
```
We never get to the right subtree.

jiashenC: I see. I will need to cover this case then. I mainly attempted to fix the following case based on our test cases. 
```mermaid
graph TD;
E1[Exchange]-->P
P[Project]-->N
N[NestedLoopJoin]-->SL
N-->SR
SL[SeqScan]-->E2
E2[Exchange]-->S1[Storage]
SR[SeqScan]-->E3
E3[Exchange]-->S2[Storage]
```

I think we can revisit the Exchange optimizer now to decide when it is actually needed and we can modify the executor based on that.

xzdandy: Got this will work. I think what we want to parallel is, for example, the UDF in the seqscan's projection. Parallelizing the Storage is not really needed. It was a "lazy" implementation, so we don't need to support partition in the Storage. But it is not  optimal.

eva/experimental/ray/executor/ray_stage.py
```diff
@@ -15,28 +15,32 @@
...
-    ray.get(tasks)
-    for q in output_queue:
-        q.put(StageCompleteSignal)
+    try:
+        ray.get(tasks)
+        for q in output_queue:
+            q.put(StageCompleteSignal)
+    except RayError as e:
+        for q in output_queue:
+            q.put(ExecutorError(e.cause))
```
xzdandy, line 36: Do need to handle `ExecutorError` in  `QueueReaderExecutor`? For example, for `StageCompleteSignal`, it needs to be put back into the input queue, so that other readers can also be notified.

eva/executor/execution_context.py
```diff
@@ -42,30 +41,25 @@ def __init__(self):
...
+        # Populate GPU IDs from yaml config file.
+        gpu_conf = self._config_manager.get_value("executor", "gpu_ids")
+        availble_gpus = [i for i in range(get_gpu_count())]
+        return list(set(availble_gpus) & set(gpu_conf))
 
     def _populate_gpu_from_env(self) -> List:
-        gpus = map(
+        # Populate GPU IDs from env variable.
+        gpu_conf = map(
             lambda x: x.strip(), os.environ.get("GPU_DEVICES", "").strip().split(",")
```
gaurav274, line 53: we should switch to the  standard env name `CUDA_VISIBLE_DEVICES`

eva/experimental/ray/executor/ray_stage.py
jiashenC: You are right. It needs to be populated to other queues as well.

eva/experimental/ray/executor/exchange_executor.py
jiashenC: I think `Storage` will still benefit from `Exchange(dop=1)` for async IOs. Otherwise, the disk read needs to happen in a blocking fashion.

My main point here is that we probably have two cases to cover for tree-like execution.
1. There exists multiple children and children are not `Exchange`. I think my current change can handle this case.
2. As you pointed out, a tree can have multiple children, which are `Exchange`. Then, we need add some mechanism for `ExchangeExecutor` to handle this case.

Maybe we have some other cases as well?

eva/experimental/ray/optimizer/rules/rules.py
jiashenC: I think we have to keep the logic here because it also needs to handle the sequential `LogicalGet` to `SeqScan` case. https://github.com/georgia-tech-db/eva/blob/1aac5cf09c99e992038cd8b4091332e233a48b2d/eva/optimizer/rules/rules_manager.py#L104

eva/experimental/ray/executor/exchange_executor.py
xzdandy: BTW, join does not work with ray with `parallelism > 1` as for now, because two ray actors will not have the complete information. Maybe tree structure is not a priority to worry about now.

eva/experimental/ray/optimizer/rules/rules.py
xzdandy: oh I see. I thought SequentialLogicalGetToSeqScan and DistributedLogicalGetToSeqScan are two different rules.

eva/experimental/ray/executor/ray_stage.py
```diff
@@ -15,28 +15,35 @@
...
```
xzdandy, line 36: Consider the following set up
```mermaid
graph TD;
      A[Projection with parallelism=1]--> B[UDF with parallelism = 2];
      B --> C[Get with parallelism=1];
```
, and there is ExecutorError raised at the bottom Get operator. 
If I understand correctly, what will happen is that one ExecutorError will be put into the queue between UDF and Get. And one Actor will get the ExecutorError and due to the fact that it is not root, the [code](https://github.com/georgia-tech-db/eva/blob/9aa9fac520a264dfc3e26583a59d816b8bf53c8d/eva/experimental/ray/executor/exchange_executor.py#L97) will not come into play. So it will try to process on the ExecutorError object, which is not expected. Furthermore, the second actor will not be affected, because it does not get the ExecutorError, which has already been read by the first actor. I think you want to modify the [QueueReaderExecutor](https://github.com/georgia-tech-db/eva/blob/9aa9fac520a264dfc3e26583a59d816b8bf53c8d/eva/experimental/ray/executor/exchange_executor.py#L44) instead.

eva/optimizer/plan_generator.py
```diff
@@ -86,8 +92,54 @@ def optimize(self, logical_plan: Operator):
...
+                        children += return_child
+                    else:
+                        children += [return_child]
+
+                plan.clear_children()
+                for child in children:
+                    plan.append_child(child)
+
+                if isinstance(plan, ExchangePlan):
+                    if is_top:
```
gaurav274, line 123: I think this code piece can be simplified. How about you assert on `is_top` and `len(plan.children) == 1` and always return `plan.children`.  This is also simplify the above check on `List`.

eva/experimental/ray/optimizer/rules/rules.py
```diff
@@ -105,7 +109,13 @@ def apply(self, before: LogicalGet, context: OptimizerContext):
...
         ):
-            return scan
-        upper = ExchangePlan(parallelism=2)
-        upper.append_child(scan)
-        yield upper
+            yield scan
+        else:
+            upper = ExchangePlan(
+                parallelism=2,
+                ray_conf={"num_gpus": 1}
```
gaurav274, line 116: @kaushikravichandran This can be moved outside to a heuristic in a future PR.

gaurav274 (commented):
LGTM. We can wait for Andy's review before merging.

gaurav274:
Please add a test case where the `exchange` removal code is triggered. Thanks!

gaurav274:
@kaushikravichandran Let us wait for a few builds and then promote `ray` from experimental to default feature. Thanks!

test/integration_tests/test_error_handling_with_ray.py
```diff
@@ -0,0 +1,73 @@
...
+        Path(img_path).unlink()
+
+    def tearDown(self):
+        shutdown_ray()
+
+        # Drop table.
+        drop_table_query = "DROP TABLE testRayErrorHandling;"
+        execute_query_fetch_all(drop_table_query)
+
+    def test_ray_error_populate_to_all_stages(self):
```
xzdandy, line 58: Thanks for the testcase!

test/optimizer/rules/test_rules.py
```diff
@@ -207,15 +225,16 @@ def test_supported_rules(self):
             LogicalJoinToPhysicalHashJoin(),
             LogicalCreateMaterializedViewToPhysical(),
             LogicalFilterToPhysical(),
-            LogicalProjectToPhysical(),
+            DistributedLogicalProjectToPhysical()
+            if ray_enabled
+            else SequentialLogicalProjectToPhysical(),
```
xzdandy, line 230: Could we add a comment that indicates right now `DistributedLogicalProjectToPhysical()` and `SequentialLogicalProjectToPhysical()` are the same?

test/optimizer/rules/test_rules.py
```diff
@@ -195,7 +211,9 @@ def test_supported_rules(self):
             LogicalInsertToPhysical(),
             LogicalDeleteToPhysical(),
             LogicalLoadToPhysical(),
-            LogicalGetToSeqScan(),
+            DistributedLogicalGetToSeqScan()
+            if ray_enabled
+            else SequentialLogicalGetToSeqScan(),
```
xzdandy, line 216: Could we add a comment that indicates `DistributedLogicalGetToSeqScan()` and `SequentialLogicalGetToSeqScan()` are same with a brief justification?

eva/optimizer/plan_generator.py
xzdandy: Is this a temporary work around? I feel We should be able to support branch and createMat. But I agree it is not priority.

xzdandy (approved):
Overall LGTM. Please check minor comments.

jiashenC:
@xzdandy @gaurav274 Fix all comments. Should be ready.

PR #623, opened 2023-03-28
feat: Adding support for AudioStorageEngine

jarulraj:
- [X] Branch based on #620


jarulraj:
@dungnmaster Can you please review this PR? It looks like we don't really need to hit `ffmpeg` `probe` \ cc @tushar-97 @gaurav274

eva/readers/decord_reader.py
```diff
@@ -56,10 +62,23 @@ def __init__(
...
+        self._read_video = read_video
         super().__init__(*args, **kwargs)
 
     def _read(self) -> Iterator[Dict]:
         decord = _lazy_import_decord()
+        if self._read_audio:
+            try:
+                if "audio" in [
+                    stream["codec_type"]
+                    for stream in ffmpeg.probe(self.file_url)["streams"]
```
dungnmaster, line 75: I think we can remove the ffmpeg dependency by catching 'DECORDError' which is thrown by AVReader when a file with only video stream is read

eva/readers/decord_reader.py
```diff
@@ -71,38 +92,54 @@ def _read(self) -> Iterator[Dict]:
...
                 while frame_id <= end:
-                    frame = video[frame_id]
+                    if has_audio:
+                        frame_audio, frame = av_reader[frame_id]
+                        frame_audio = frame_audio.asnumpy()
+                        if self._read_video is None or not self._read_video:
+                            frame = np.empty(0)
+                    else:
+                        frame = v_reader[frame_id]
+                        frame_audio = np.empty(0)
```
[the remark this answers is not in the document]
gaurav274: Any reason for this design?

eva/readers/decord_reader.py
```diff
@@ -71,38 +92,54 @@ def _read(self) -> Iterator[Dict]:
...
+                    frame = frame.asnumpy()
+                    timestamp = v_reader.get_frame_timestamp(frame_id)[0]
```
gaurav274, line 136: I see a lot of code duplication. How about we create private functions to improve readability and maintainability?

eva/catalog/catalog_type.py
```diff
@@ -98,3 +98,20 @@ class IndexType(EVAEnum):
     @classmethod
     def is_faiss_index_type(cls, t):
         return t in [cls.HNSW]
+
+
+class ColumnName(EVAEnum):
```
gaurav274, line 103: Like this! Maybe make the name more descriptive and add similar logic for images

eva/catalog/catalog_utils.py
```diff
@@ -39,6 +39,7 @@ def get_video_table_column_definitions() -> List[ColumnDefinition]:
             "data", ColumnType.NDARRAY, NdArrayType.UINT8, (None, None, None)
         ),
         ColumnDefinition("seconds", ColumnType.FLOAT, None, []),
+        ColumnDefinition("audio", ColumnType.NDARRAY, NdArrayType.UINT8, (None)),
```
gaurav274, line 42: Can we use ColumnNames from `enum` here?

eva/binder/statement_binder.py
```diff
@@ -207,6 +211,10 @@ def _bind_tuple_expr(self, node: TupleValueExpression):
         table_alias, col_obj = self._binder_context.get_binded_column(
             node.col_name, node.table_alias
         )
+        if node.col_name == ColumnName.audio:
```
gaurav274, line 214: Do we need it for deletion? We will just drop the video in both cases, right?

eva/binder/statement_binder.py
```diff
@@ -207,6 +211,10 @@ def _bind_tuple_expr(self, node: TupleValueExpression):
...
+            self._binder_context.enable_audio_retrieval()
+        if node.col_name == ColumnName.data:
+            self._binder_context.enable_video_retrieval()
```
gaurav274, line 217: Neat Design!

eva/readers/decord_reader.py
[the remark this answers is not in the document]
tushar-97: Just adding empty arrays for completeness. For example if user does SELECT * FROM VIDEO on a video that doesn't have an audio stream, we would still need something to show for under the audio column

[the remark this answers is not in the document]
gaurav274: Sounds good

eva/binder/statement_binder.py
tushar-97: I don't follow :sweat_smile: I added this in  `_bind_tuple_expr` but from the UI it makes it look like I added this in the bind delete part. Is that what is causing the confusion?

gaurav274: I see. My bad

gaurav274: I see. My bad

eva/readers/decord_reader.py
tushar-97: Done

eva/catalog/catalog_utils.py
tushar-97: Done

eva/readers/decord_reader.py
```diff
@@ -51,16 +56,32 @@ def __init__(
...
+        av_reader = None
+        # Check for availability of audio stream
+        # If there is an audio stream => AVReader
+        # Else => VideoReader (no audio)
+        if self._read_audio:
+            try:
+                av_reader = decord.AVReader(self.file_url, mono=True, sample_rate=16000)
+            except decord._ffi.base.DECORDError as error_msg:
+                if "Can't find audio stream" in str(error_msg):
+                    print("No audio stream in video")
```
gaurav274, line 80: We should `assert` here

gaurav274: And we should rather return the error message from decord.

eva/readers/decord_reader.py
```diff
@@ -108,12 +121,34 @@ def _read(self) -> Iterator[Dict]:
                 if begin % self._sampling_rate:
                     begin += self._sampling_rate - (begin % self._sampling_rate)
                 for frame_id in range(begin, end + 1, self._sampling_rate):
-                    frame = video[frame_id].asnumpy()
+                    frame = self.__get_frame(frame_id, av_reader, v_reader)
```
gaurav274, line 124: We can yield directly from the function rather than checking for `None` again.

test/readers/test_decord_reader.py
```diff
@@ -140,3 +145,49 @@ def test_should_sample_every_k_frame(self):
...
+            batch_mem_size=100000000000,
+            file_url=self.video_with_audio_file_url,
+            read_audio=True,
+            read_video=True,
+        )
+        batches = list(video_loader.read())
+        # gave a big enough batch_mem_size so that all frames fit in one batch
+        self.assertEqual(len(batches[0]), 996)
+        # verify that the 100th frame is read correctly and hope that all the other frames were too!
+        assert np.sum(batches[0].frames.loc[100]["audio"]) == pytest.approx(2.7592432)
```
gaurav274, line 170: How about also pushing a small numpy of decoded data and comparing against it?

test/readers/test_decord_reader.py
```diff
@@ -140,3 +145,49 @@ def test_should_sample_every_k_frame(self):
...
+            batch_mem_size=100000000000,
+            file_url=self.video_with_audio_file_url,
+            read_audio=True,
+            read_video=False,
+        )
+        batches = list(video_loader.read())
+        # gave a big enough batch_mem_size so that all frames fit in one batch
+        self.assertEqual(len(batches[0]), 996)
+        # verify that the 100th frame is read correctly and hope that all the other frames were too!
+        assert np.sum(batches[0].frames.loc[100]["audio"]) == pytest.approx(2.7592432)
```
gaurav274, line 188: Same as above.

eva/catalog/catalog_type.py
gaurav274: How about we call them VideoColumnNames?

gaurav274 (commented):
Overall looks good apart from minor comments.

eva/readers/decord_reader.py
gaurav274: Btw, we should check for `AudioReader ` or `VideoReader` or `AVReader`?

eva/readers/decord_reader.py
```diff
@@ -108,12 +121,34 @@ def _read(self) -> Iterator[Dict]:
...
-                        yield {
-                            "id": frame_id,
-                            "data": frame,
-                            "seconds": round(video.get_frame_timestamp(frame_id)[0], 2),
-                        }
+                        yield frame
                     else:
                         break
+
+    def __get_frame(self, frame_id, av_reader, v_reader):
```
gaurav274, line 130: We shouldn't hit these decisions for every frame. How about we have 3 functions and use logic upfront to decide the correct callable and then just invoke it?

suryatejreddy:
Just FYI, decord can't be installed via pypi on recent macos with pyhton3.9. [Reference](https://github.com/dmlc/decord/issues/213)

eva/binder/statement_binder.py
```diff
@@ -147,6 +147,15 @@ def _bind_select_statement(self, node: SelectStatement):
             self.bind(node.union_link)
             self._binder_context = current_context
 
+        assert not (
+            self._binder_context.is_retrieve_audio()
+            and self._binder_context.is_retrieve_video()
+        ), "Cannot query over both audio and video streams"
```
xzdandy, line 153: To make sure my understanding is correct, does this mean we can not query audio and video in the same query?

gaurav274: Yes, we are going with this design for now. We plan to support both in the future if we find a use case.

xzdandy: Got it. We can explore a JOIN-like syntax for video and audio in the future. For example,  `SELECT ... FROM VIDEO, AUDIO WHERE ...`.

eva/readers/decord_reader.py
```diff
@@ -51,16 +56,29 @@ def __init__(
...
         self._sampling_type = sampling_type
+        self._read_audio = read_audio
+        self._read_video = read_video
         super().__init__(*args, **kwargs)
 
     def _read(self) -> Iterator[Dict]:
         decord = _lazy_import_decord()
-        video = decord.VideoReader(self.file_url)
-        num_frames = int(len(video))
+        if self._read_audio:
```
gaurav274, line 71: nit: we can move this logic (reader selection) to a separate function. Just want to clean the code as much as possible.

eva/readers/decord_reader.py
```diff
@@ -70,50 +88,50 @@ def _read(self) -> Iterator[Dict]:
...
+        timestamp = reader.get_frame_timestamp(frame_id)[0]
+
+        return {
+            VideoColumnName.id.name: frame_id,
+            VideoColumnName.data.name: frame_video,
+            VideoColumnName.seconds.name: round(timestamp, 2),
+        }
+
+    def __get_audio_frame(self, frame_id, reader):
+        frame_audio, _ = reader[frame_id]
```
gaurav274, line 129: So, we are going with audio per frame design. Will this work for whisper? If yes, we are good. We can worry about the sampling rate in the next PR when we add whisper.

gaurav274 (approved):
Minor comments. Fix and merge. Thanks for the effort!

eva/readers/decord_reader.py
tushar-97: Yes, I verified that it works with whisper

Issue #414, opened 2022-10-01
Frame Accurate Seeking


BrianPugh:
[OpenCV's video reader suffers from an issue](https://github.com/opencv/opencv/issues/9053) where frames after setting `CAP_PROP_POS_FRAMES` might either be invalid, or inaccurate (e.g. seeking to frame 1022 might actually seek to frame 1005). For some projects, this might not be a huge deal, but for projects that need to be synchronized with other data, this could be a huge silent bug.

While not a complete solution (I'm not too familiar with EVA's codebase), linked below is an easy-to-use, frame-accurate video reader using PyAV. PyAV has ffmpeg builtin, so it doesn't add any additional external non-python dependencies that could complicate installation. Its (mostly) a drop in for the opencv reader.

https://gist.github.com/BrianPugh/8de1a6f124516ec2a2a61caadcdf039a

Unrelated self-promotion: For managing your different readers, storage-backends, etc; consider checking out my [AutoRegistry](https://github.com/BrianPugh/autoregistry) project. Basically if your abstract base classes inherit from `Registry` instead of `ABC/ABCMeta`, child classes can be easily accessed via strings (from a config file, or other).

gaurav274:
@BrianPugh Nice catch! Thanks for raising the issue. We are aware of the incorrect frame seeking in OpenCV. Based on our prior discussions, I recall we found another library [decord](https://github.com/dmlc/decord) that provides an accurate frame seeking out of the box. However, we did not change the storage engine as we were considering moving to a `time-based` seeking instead of `frame based.` IMO, time-based seeking seems intuitive to use in queries compared to frame-based. However, we might have to support both. 

Anyways, for frame-based seeking, OpenCV has to go. `PyAV`solution looks neat and can easily replace OpenCV. I guess it boils down to performance metrics and stable support. 
 
`AutoRgistry` sounds interesting. I'll check it out and consider the possibility of using it in `eva`.

jarulraj:
@dungnmaster Is this issue fixed with our `decord` based reader?

dungnmaster:
@jarulraj This should be fixed with the decord based reader but we are currently using it only for IFRAMES based queries. Should be a small change to switch every query to decord reader.

petered:
Just a heads up - the seek of 
https://gist.github.com/BrianPugh/8de1a6f124516ec2a2a61caadcdf039a
Isn't frame-accurate for me (sometimes 1 frame off, sometimes more). 

[...]

PR #625, opened 2023-03-30
Bug: Reader fixes

gaurav274:
1. Remove OpenCV based reader and fully move to decord - closing issue #414
2. Fix decord and csv reader test cases that weren't testing correctly.
3. Fix bug with `open` UDF (https://github.com/opencv/opencv/issues/14095)
4. Remove `timestamp` UDF as we now support it using  `decord`


gaurav274:
@tushar-97 If this looks good, we can merge.

tushar-97 (approved):

PR #627, opened 2023-04-01
feat: adding as_df() in response

aryan-rajoria:
Adding `as_df()` function into response, user will now use `response.as_df()` instead of `response.batch.frames` to get dataframe returned by queries.

Updating notebooks to reflect the same.


eva/models/server/response.py
```diff
@@ -44,6 +44,9 @@ def deserialize(cls, data):
         obj = PickleSerializer.deserialize(data)
         return obj
 
+    def as_df(self):
+        return self.batch.frames
```
gaurav274, line 48: `Raise error if the batch is None`

gaurav274 (approved):

PR #628, opened 2023-04-02
test: adding support for pytest-xdist

jarulraj:
- [X] These changes create `worker-specific` config folders and databases, it is possible to run `pytest test -n 4`

```
(test_evadb) ada-01@~/eva ∴ ls ~/.eva/0.1.5+dev/                                                                
gw0  gw2  test_gw0_eva_catalog.db  test_gw2_eva_catalog.db                                                      
gw1  gw3  test_gw1_eva_catalog.db  test_gw3_eva_catalog.db                                                        
```                                                                                     

- [X] Added support for `MacOS` build


gaurav274:
> -[X] These changes create `worker-specific` config folders and databases, it is possible to run `pytest test -n 4`
> 
> ```
> (test_evadb) ada-01@~/eva ∴ ls ~/.eva/0.1.5+dev/                                                                
> gw0  gw2  test_gw0_eva_catalog.db  test_gw2_eva_catalog.db                                                      
> gw1  gw3  test_gw1_eva_catalog.db  test_gw3_eva_catalog.db                                                        
> ```
> 
> * [ ]  This test is flaky because it does not have their own setup (rely on some other test file's setup)? \ cc @gaurav274
> 
> ```
> test/catalog/services/test_index_catalog_service.py::IndexCatalogServiceTest::test_index_get_all_entries
> sqlalchemy error: no such table: index_catalog
> ```

Hopefully, my new changes should fix it. 
Btw, locally it failed on a bunch of test cases with `no such file error`.

jarulraj:
@gaurav274 I also added a `MacOS` and updated the `Windows` build. Can you look at the updated PR?

With MacOS and Python version 3.8, the FAISS library hits a segfault -- https://georgia-tech-db.slack.com/archives/D02CHL23KDM/p1681059206477049
With MacOS and Python versions 3.9 and 3.10, there is no `decord` package.

script/test/test.sh
```diff
@@ -76,23 +76,19 @@ if [[ "$OSTYPE" != "msys" ]];
 then
     if [[ "$MODE" = "TEST" || "$MODE" = "ALL" ]];
     then
-        PYTHONPATH=./ pytest --cov-report term-missing:skip-covered  --cov-config=.coveragerc --cov-context=test --cov=eva/ -s -v --log-level=WARNING -m "not benchmark"
-    elif [[ "$MODE" = "RAY" ]];
```
jiashenC, line 80: I think this causes Ray test to not run.

jarulraj: Good catch!

eva/catalog/sql_config.py
```diff
@@ -58,10 +48,20 @@ def __init__(self):
         Retrieves the database uri for connection from ConfigurationManager.
         """
         uri = ConfigurationManager().get_value("core", "catalog_database_uri")
+
         # parallelize using xdist
-        # worker_uri = prefix_worker_id(str(uri))
+        def prefix_worker_id(uri: str):
```
gaurav274, line 53: How is it different from the similar function in `utils`?

eva/configuration/configuration_manager.py
```diff
@@ -34,11 +36,25 @@ def __new__(cls):
...
+            worker_id = os.environ["PYTEST_XDIST_WORKER"]
+            path = path / str(worker_id)
+        except KeyError:
+            pass
+        return path
+
     @classmethod
     def _create_if_not_exists(cls):
         if not cls._yml_path.exists():
+            initial_eva_config_dir = Path(EVA_DEFAULT_DIR)
```
gaurav274, line 51: Please document the need for the logic. Thanks!

eva/catalog/sql_config.py
jarulraj: One of them is for a database uri. The other one is for a directory path.

eva/configuration/configuration_manager.py
jarulraj: Earlier, we directly used `EVA_DEFAULT_DIR`. Now, we prefix the worker id if needed. I will add a comment.

PR #633, opened 2023-04-03
fix: Install Decord from EVA-Fork

suryatejreddy:
The current `pypi` package of `decord` doesn't build on newer macos versions (v13). This PR installed EVA from [our own fork of record.](https://github.com/georgia-tech-db/eva-decord)


gaurav274:
@suryatejreddy Push the package to `pypi` and let us close the PR. Thanks!

gaurav274 (approved):

PR #636, opened 2023-04-05
fix: Timeout Error

aryan-rajoria:
[server.py](https://github.com/georgia-tech-db/eva/blob/0f4160692b37b6343a5e47d22e3c83aff7a15508/eva/server/server.py#L73) was set to timeout after `60.0` seconds, now changing that to `None`.

It should resolve the error where notebooks randomly fail tests

### DIFF
--- a/eva/server/server.py
+++ b/eva/server/server.py
@@ -70,7 +70,7 @@ class EvaServer:
     ):
         try:
             while True:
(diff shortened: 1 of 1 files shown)


jarulraj:
Very good catch @aryan-rajoria

PR #641, opened 2023-04-06
fix: Unnest bug 

gaurav274:
```
SELECT id, yolov5.label, yolov5.bbox, yolov5.score
FROM ObjectDetectionVideos
JOIN LATERAL UNNEST(YoloV5(data)) AS yolov5(label, bbox, score)
WHERE id < 20 AND yolov5.label = 'car';
```
The above does not work correctly in the current system. 
Reasons:
1. Since the `PushDownFilterThroughJoin` rule is applied before the `XformLateralJoinToLinearFlow`, the plan never got a linear flow and uses the old LateralJoin code which was buggy. Prioritizing `XformLateralJoinToLinearFlow` now.
2. Fixed bug in the old LateralJoin code by not resetting the batch index used to join in lateral join.
3. Remove `EmbedProjectIntoGet` rule as it was inconsistent.


xzdandy:
Is `old LateralJoin code`  still needed for any corner cases? Maybe we should print a warning message when the `old LateralJoin code` is being called to avoid accidentally  falling back to the old code.

eva/optimizer/optimizer_tasks.py
```diff
@@ -91,6 +91,8 @@ def execute(self):
                     self.optimizer_context.task_stack.push(
                         TopDownRewrite(new_expr, self.rule_set, self.optimizer_context)
                     )
+                    # The root has changed so we cannot apply more rules to the same root, hence return
+                    return
```
xzdandy, line 96: 1. The return is not at the right indention. I think we need to go through all the plan in after.
2. And in a cost-based optimization, multiple rules can be matched to the same root. I don't think we can just return and ignore the rest.

gaurav274: 1. Rewrite plans should only return 1 output, else it is ambiguous.  I'll add an assert message to take care of it.
2. Rewrites are not part of cost-based optimizations so we are fine. They are taken care of by the state machine.

xzdandy: Ah, you are right. Topdown is only rewrite rule not for logical rules. I missed that. Thanks!

xzdandy (approved):

PR #643, opened 2023-04-06
feat: Add audio sampling rate

tushar-97:
This feature allows the users to specify the audio sampling rate like so:

`SELECT UDF(audio,..) FROM VIDEOS SAMPLE AUDIORATE 22500`

If no sampling rate is given, we default to 16000

### DIFF
--- a/test/readers/test_decord_reader.py
+++ b/test/readers/test_decord_reader.py
@@ -25,7 +25,7 @@ import numpy as np
 import pytest
 
 from eva.configuration.constants import EVA_ROOT_DIR
-from eva.constants import IFRAMES
+from eva.constants import AUDIORATE, IFRAMES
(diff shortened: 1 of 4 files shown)


test/readers/test_decord_reader.py
```diff
@@ -164,32 +164,55 @@ def test_should_throw_error_for_audioless_video(self):
...
-            read_video=False,
-        )
         try:
+            video_loader = DecordReader(
+                file_url=self.video_with_audio_file_url,
+                sampling_type=IFRAMES,
+                read_audio=True,
+                read_video=False,
+            )
             list(video_loader.read())
```
xzdandy, line 174: `with pytest.raises` seems to be better approach to test raising exceptions.

PR #644, opened 2023-04-06
chore: Add optimizer timeout

kaushikravichandran:
This PR adds a timeout of 120 seconds to the EVA optimizer.


eva/optimizer/plan_generator.py
```diff
@@ -131,9 +133,12 @@ def _recursive_strip_exchange(plan: AbstractPlan, is_top: bool = False):
         else:
             return physical_plan
 
-    def build(self, logical_plan: Operator):
+    async def build(self, logical_plan: Operator):
         # apply optimizations
-        plan = self.optimize(logical_plan)
+        try:
+            plan = await wait_for(self.optimize(logical_plan), timeout=60.0)
```
xzdandy, line 139: Do we decide we want to use `asyncio` to achieve this goal? If so, I think we need to wrap the complete processing pipeline in async, instead of calling `asyncio.run`. Or we can use `loop.run_until_complete`. BTW, how do we ensure the plan we get is valid when it times out?

kaushikravichandran: We throw an error if we time out, so EVA crashes.
There are two alternatives to `asyncio`:
- Using `signal`, but this only has UNIX support
- Using a third party software, which provides timeout support using decoraters, but it needs to be installed to be used

kaushikravichandran: How would `loop.run_until_complete` be different from `asyncio.run`? Don't they both use event loops in the end?

jarulraj: In Postgres, a basic optimization algorithm is used to come up with a valid plan. After that, the optimizer keeps iteratively refining it using genetic programming -- and this process can be stopped at any time to get a valid plan. We will need a boolean variable to check if a valid plan is constructed (first stage is completed).

jarulraj: try:
    loop.run_until_complete(asyncio.wait_for(my_coroutine(), timeout=3))
except asyncio.TimeoutError:
    print("Coroutine took too long to complete!")
finally:
    loop.close()

jarulraj: asyncio.run creates a new event loop by default, while loop.run_until_complete uses the current event loop

xzdandy: I feel this is non-trivial feature and may require some re-design of the optimizer architecture. Is this priority? I think we do not have complicated rules now that will cause the optimizer to run over 60 seconds. @gaurav274 What do you think?

gaurav274: My motivation behind this feature is to throw an error if we cannot construct a valid plan for the query. Ideally, this should not occur but due to some bug in the optimizer or poor handling in the binder, we might reach such a state. Can we accomplish this use case without any changes to the optimizer?

kaushikravichandran: @xzdandy signal has only UNIX support, which is a problem.

xzdandy: You are right. Let's go for asyncio approach.

gaurav274:
@xzdandy @kaushikravichandran  is it ready to merged?

xzdandy:
From the [doc](http://google3/production/borg/contentads/loadtest/omegaperf-ags-leaf-setup.borg;l=74;ws=codereview%2F595759384;rcl=1), 

>This function always creates a new event loop and closes it at the end. It should be used as a main entry point for asyncio programs, and should ideally only be called once.

Please check if we are good to call asynctio.run multiple times [reference](https://stackoverflow.com/questions/69035759/using-asyncio-run-is-it-safe-to-run-multiple-times). 

Could we have a test case for the timeout?

kaushikravichandran:
> From the [doc](http://google3/production/borg/contentads/loadtest/omegaperf-ags-leaf-setup.borg;l=74;ws=codereview%2F595759384;rcl=1),
> 
> > This function always creates a new event loop and closes it at the end. It should be used as a main entry point for asyncio programs, and should ideally only be called once.
> 
> Please check if we are good to call asynctio.run multiple times [reference](https://stackoverflow.com/questions/69035759/using-asyncio-run-is-it-safe-to-run-multiple-times).
> 
> Could we have a test case for the timeout?

Yes, we are good calling asyncio.run(). It cannot be called when another asyncio event loop is running in the same thread, which is the case. Using `loop.run_until_complete` is not correct

> From the [doc](http://google3/production/borg/contentads/loadtest/omegaperf-ags-leaf-setup.borg;l=74;ws=codereview%2F595759384;rcl=1),
> 
> > This function always creates a new event loop and closes it at the end. It should be used as a main entry point for asyncio programs, and should ideally only be called once.
> 
> Please check if we are good to call asynctio.run multiple times [reference](https://stackoverflow.com/questions/69035759/using-asyncio-run-is-it-safe-to-run-multiple-times).
> 
> Could we have a test case for the timeout?

Yes, we are good to use it. Looks like `loop.run_until_complete` is blocking, which defeats our purpose. Moreover, `loop` is deprecated with `python3.8`, so `asyncio.run` is our only choice.

jarulraj:
Looks good to me!

jarulraj (approved):

PR #653, opened 2023-04-15
test: Use concurrent accesses instead of multiple databases under xdist

sameer-s:
@jarulraj These are currently the only stable changes that we have for this milestone. On my system, makes all tests pass when run with xdist.

To run the tests, I use:

`pytest --cov-report term-missing:skip-covered  --cov-config=.coveragerc --cov-context=test --cov=eva/ -s -v --log-level=WARNING -m "not benchmark" --dist loadscope -n auto`


jarulraj:
Thanks for the changes. We will borrow ideas from this PR.

PR #655, opened 2023-04-17
feat: LLM-based remote UDFs -- support for ChatGPT

IshSiva:
- [x] Support OpenAI based ChatCompletion UDF

```
CREATE UDF {udf_name}
IMPL 'eva/udfs/openai_chat_completion_udf.py'
'model' 'gpt-3.5-turbo'

```


eva/udfs/gpt_udf.py
```diff
@@ -0,0 +1,78 @@
...
+                ],
+                column_shapes=[(None,), (None,)],
+            )
+        ],
+    )
+    def forward(self, text_df):
+
+        result = []
+        queries = text_df[text_df.columns[0]]
+        cache = {}
```
gaurav274, line 62: Why are we `caching` here? This should rather use system-wide caching framwork.

eva/udfs/gpt_udf.py
```diff
@@ -0,0 +1,78 @@
...
+        for query in queries:
+            if query in cache:
+                result.append(cache[query])
+
+            else:
+                response = openai.ChatCompletion.create(
+                    model="gpt-3.5-turbo", messages=[{"role": "user", "content": query}]
+                )
+                for choice in response.choices:
+                    result.append(choice.message.content)
```
gaurav274, line 73: Do we have multiple responses for a single query? If yes then the returned output will have a mismatched number of rows.

eva/udfs/gpt_udf.py
```diff
@@ -0,0 +1,78 @@
...
+                    cache[query] = choice.message.content
+
+        df = pd.DataFrame({"input_query": queries, "responses": result})
```
gaurav274, line 76: Why are we returning back `input_query`?

eva/udfs/gpt_udf.py
gaurav274: Do we need to concat all the choices?

eva/udfs/gpt_udf.py
```diff
@@ -0,0 +1,78 @@
...
+        result = []
+        queries = text_df[text_df.columns[0]]
+        cache = {}
+
+        for query in queries:
+            if query in cache:
+                result.append(cache[query])
+
+            else:
+                response = openai.ChatCompletion.create(
```
gaurav274, line 69: Can't we send all the queries at once?

eva/udfs/gpt_udf.py
IshSiva: I can remove this as this is not required. This was done when I was testing something on my local machine.

eva/udfs/gpt_udf.py
IshSiva: No, its just one response for a single query.

eva/udfs/gpt_udf.py
```diff
@@ -0,0 +1,83 @@
...
+        prompts = text_df[text_df.columns[0]]
+        queries = text_df[text_df.columns[1]]
+
+        # chatgpt api currently supports answers to a single prompt only
+        # so this udf is designed such that
+
+        results = []
+
+        for prompt, query in zip(prompts, queries):
+            if prompt != "None":
```
suryatejreddy, line 72: Do we necessarily need a prompt? I couldn't find it in the OpenAI docs (https://platform.openai.com/docs/guides/chat/chat-vs-completions). Or is this because it makes downstream tasks easy?

test/data/queries.csv
```diff
@@ -0,0 +1,2 @@
+id,prompt,query
+0,summarize,"There are times when the night sky glows with bands of color. The bands may begin as cloud shapes and then spread into a great arc across the entire sky. They may fall in folds like a curtain drawn across the heavens. The lights usually grow brighter, then suddenly dim. During this time the sky glows with pale yellow, pink, green, violet, blue, and red. These lights are called the Aurora Borealis. Some people call them the Northern Lights. Scientists have been watching them for hundreds of years. They are not quite sure what causes them"
```
suryatejreddy, line 2: Do we have to git track this? I think we can make it similar to create_dummy_csv.

test/integration_tests/test_chatgpt_udf.py
```diff
@@ -0,0 +1,73 @@
...
+        execute_query_fetch_all(csv_query)
+
+    def tearDown(self) -> None:
+        execute_query_fetch_all("DROP TABLE IF EXISTS MyTextCSV;")
+
+    @ray_skip_marker
+    @patch("eva.udfs.gpt_udf.openai.ChatCompletion.create")
+    def test_gpt_udf(self, mock_req):
+        # set dummy api key
+        ConfigurationManager().update_value("third_party", "openai_api_key", "my_key")
```
suryatejreddy, line 51: Probably need to add a test case when no api key is set. Coverage will ask for it.

eva/udfs/gpt_udf.py
IshSiva: This is to just specify the task. For example, if we want to summarize we can just give "summarize" in the prompt field and the remaining text in the query field. So the content field in the api call is formed as "summarize: <query text>".

test/integration_tests/test_chatgpt_udf.py
IshSiva: added.

test/data/queries.csv
IshSiva: done.

eva/udfs/gpt_udf.py
IshSiva: removed this.

test/data/queries.csv
IshSiva: removed this.

gaurav274 (approved):

PR #659, opened 2023-04-19
feat: Add support for audio based hf models

tushar-97:
- [x] Changes to base HF classes to support audio based models
- [x] Refactored a few classes to get the correct model based on task and not input type
- [x] Made changes to allow decimal values for metadata values that can be passed to the pipeline API
- [x] Added test cases for getting summary of a video based on its transcript

### DIFF
--- a/eva/parser/lark_visitor/_functions.py
+++ b/eva/parser/lark_visitor/_functions.py
@@ -17,6 +17,7 @@ from lark import Token, Tree
 
 from eva.expression.abstract_expression import ExpressionType
 from eva.expression.aggregation_expression import AggregationExpression
+from eva.expression.constant_value_expression import ConstantValueExpression
 from eva.expression.function_expression import FunctionExpression
 from eva.expression.tuple_value_expression import TupleValueExpression
 from eva.parser.create_udf_statement import CreateUDFStatement
@@ -101,7 +102,11 @@ class Functions:
                 elif child.data == "udf_metadata":
                     # Each UDF metadata is a key value pair
                     key_value_pair = self.visit(child)
-                    metadata.append((key_value_pair[0].value, key_value_pair[1].value)),
+                    # value can be an integer or string
+                    value = key_value_pair[1]
+                    if isinstance(value, ConstantValueExpression):
+                        value = value.value
+                    metadata.append((key_value_pair[0].value, value)),
 
         return CreateUDFStatement(
             udf_name,
--- a/eva/third_party/huggingface/binder.py
+++ b/eva/third_party/huggingface/binder.py
@@ -12,28 +12,28 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from eva.catalog.catalog_type import NdArrayType
 from eva.catalog.models.udf_catalog import UdfCatalogEntry
-from eva.third_party.huggingface.model import ImageHFModel, TextHFModel
+from eva.third_party.huggingface.create import MODEL_FOR_TASK
 
 
 def assign_hf_udf(udf_obj: UdfCatalogEntry):
     """
     Assigns the correct HF Model to the UDF. The model assigned depends on
-    the type of input the UDF expects. This is done so that we can
+    the task type for the UDF. This is done so that we can
     process the input correctly before passing it to the HF model.
     """
     inputs = udf_obj.args
 
     # NOTE: Currently, we only support models that require a single input.
     assert len(inputs) == 1, "Only single input models are supported."
-    input_type = inputs[0].array_type
 
-    # By default, we assume that the input is an image.
-    model_class = ImageHFModel
+    task = None
+    for metadata in udf_obj.metadata:
+        if metadata.key == "task":
+            task = metadata.value
+            break
 
-    # If the input is a string, assign a TextHFModel.
-    if input_type == NdArrayType.STR:
-        model_class = TextHFModel
+    assert task is not None, f"Task is not defined for {udf_obj.name}"
+    model_class = MODEL_FOR_TASK[task]
 
     return lambda: model_class(udf_obj)
--- a/eva/third_party/huggingface/binder.py
+++ b/eva/third_party/huggingface/binder.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 from eva.catalog.models.udf_catalog import UdfCatalogEntry
 from eva.third_party.huggingface.create import MODEL_FOR_TASK
+from eva.third_party.huggingface.utils import get_metadata_entry
 
 
 def assign_hf_udf(udf_obj: UdfCatalogEntry):
@@ -27,13 +28,7 @@ def assign_hf_udf(udf_obj: UdfCatalogEntry):
     # NOTE: Currently, we only support models that require a single input.
     assert len(inputs) == 1, "Only single input models are supported."
 
-    task = None
-    for metadata in udf_obj.metadata:
-        if metadata.key == "task":
-            task = metadata.value
-            break
-
-    assert task is not None, f"Task is not defined for {udf_obj.name}"
+    task = get_metadata_entry(udf_obj, "task")[1]
     model_class = MODEL_FOR_TASK[task]
 
     return lambda: model_class(udf_obj)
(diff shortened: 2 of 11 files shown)


eva/third_party/huggingface/model.py
```diff
@@ -48,3 +49,30 @@ def input_formatter(self, inputs: Any):
...
+    Base Model for all HF Models that take in audio as input
+    """
+
+    def input_formatter(self, inputs: Any):
+        audio = np.empty((0, 0))
+        files = set(inputs.iloc[:, 0].tolist())
+        for file in files:
+            reader = decord.AudioReader(file, mono=True, sample_rate=16000)
+            audio = np.append(audio, reader[0:].asnumpy()[0])
+        return audio
```
tushar-97, line 65: This is necessary because most models are trained with a 16000 sampling rate. In the future we could maintain a map of model to sampling rate

eva/parser/lark_visitor/_functions.py
```diff
@@ -101,7 +102,11 @@ def create_udf(self, tree):
                 elif child.data == "udf_metadata":
                     # Each UDF metadata is a key value pair
                     key_value_pair = self.visit(child)
-                    metadata.append((key_value_pair[0].value, key_value_pair[1].value)),
+                    # value can be an integer or string
+                    value = key_value_pair[1]
+                    if isinstance(value, ConstantValueExpression):
```
suryatejreddy, line 107: This should happen inside the visit imo (that's how it's done with the other nodes in this file). I think when you visit, if there is a constant value expression (decimal), then it should get the value and return it.

eva/udfs/abstract/hf_abstract_udf.py
```diff
@@ -43,12 +43,21 @@ def name(self) -> str:
...
+            else:
+                pipeline_args[entry.key] = entry.value
         self.hf_udf_obj = pipeline(**pipeline_args, device=device)
 
     def setup(self, *args, **kwargs) -> None:
         super().setup(*args, **kwargs)
 
+    @property
+    def default_pipeline_args(self) -> dict:
+        return {}
```
suryatejreddy, line 59: I am guessing this is for future when we would some default arguments. In that case, a comment here would be nice. Or else it's like a property that is always {}

eva/udfs/abstract/hf_abstract_udf.py
```diff
@@ -69,7 +80,7 @@ def output_formatter(self, outputs: Any):
         result_df = pd.DataFrame(result_list)
         return result_df
 
-    def forward(self, inputs, *args, **kwargs) -> None:
+    def forward(self, inputs, *args, **kwargs) -> pd.DataFrame:
```
suryatejreddy, line 83: 👍

eva/udfs/abstract/hf_abstract_udf.py
```diff
@@ -59,6 +68,8 @@ def output_formatter(self, outputs: Any):
         """
         Function that formats output from HuggingFace format to EVA format (pandas dataframe)
         """
+        if isinstance(outputs, dict):
```
suryatejreddy, line 71: Just for my understanding, can you give an example of this type of ouput. Is it of the type `{labels: [1,2,1,2,1], class: ['car', 'bus', 'car', 'bus', 'car']}. We probably need examples in the docstrings on what are the type of outputs that we expect here. I can do it for the others but you could mention for audio.

script/formatting/formatter.py
```diff
@@ -195,7 +195,7 @@ def format_file(file_path, add_header, strip_header, format_code):
             os.system(black_command)
 
             # AUTOFLAKE
-            autoflake_command = f"{FLAKE_BINARY} --config={FLAKE8_CONFIG} {file_path}"
+            autoflake_command = f"{FLAKE_BINARY} --config='{FLAKE8_CONFIG}' {file_path}"
```
suryatejreddy, line 198: 👍

test/integration_tests/test_huggingface_udfs.py
```diff
@@ -35,11 +35,19 @@ def setUp(self) -> None:
         query = """LOAD VIDEO 'data/ua_detrac/ua_detrac.mp4' INTO DETRAC;"""
         execute_query_fetch_all(query)
 
+        create_table = """CREATE TABLE IF NOT EXISTS VIDEOS (path TEXT(1000));"""
```
suryatejreddy, line 38: nit: VIDEO_SOURCES/VIDEO_PATHS as table name instead of VIDEOS.

test/integration_tests/test_huggingface_udfs.py
```diff
@@ -197,3 +205,53 @@ def test_text_classification(self):
...
+
+        # verify that output has one row and one column only
+        self.assertTrue(output.frames.shape == (1, 1))
+        # verify that speech was converted to text correctly
+        self.assertTrue(output.frames.iloc[0][0].count("touchdown") == 2)
+
+        drop_udf_query = f"DROP UDF {udf_name};"
+        execute_query_fetch_all(drop_udf_query)
+
+    def test_summarization_from_video(self):
```
suryatejreddy, line 228: Great E2E test case!

eva/third_party/huggingface/model.py
```diff
@@ -48,3 +49,30 @@ def input_formatter(self, inputs: Any):
...
+class AudioHFModel(AbstractHFUdf):
+    """
+    Base Model for all HF Models that take in audio as input
+    """
+
+    def input_formatter(self, inputs: Any):
+        audio = []
+        files = inputs.iloc[:, 0].tolist()
+        for file in files:
+            reader = decord.AudioReader(file, mono=True, sample_rate=16000)
```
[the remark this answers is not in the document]
suryatejreddy: Probably add a comment in the code so that future devs are aware.

eva/third_party/huggingface/create.py
```diff
@@ -21,13 +21,34 @@
...
+    ImageHFModel,
+    TextHFModel,
+)
 
 """
 We currently support the following tasks from HuggingFace.
 Each task is mapped to the type of input it expects.
 """
-SUPPORTED_TASKS = {
+SUPPORTED_TASKS = [
```
suryatejreddy, line 36: Feels like repetitive information. Any reason for breaking it into a List and a Dict? The keys are the same, can probably just change the name of the dictionary.

eva/third_party/huggingface/binder.py
```diff
@@ -12,28 +12,28 @@
...
     assert len(inputs) == 1, "Only single input models are supported."
-    input_type = inputs[0].array_type
 
-    # By default, we assume that the input is an image.
-    model_class = ImageHFModel
+    task = None
+    for metadata in udf_obj.metadata:
+        if metadata.key == "task":
+            task = metadata.value
+            break
```
suryatejreddy, line 35: Can probably make this a helper function, `get_metadata_entry(<key>)` that can be reused in other parts of the code. I believe we will use this for caching and batching as well.

suryatejreddy (commented):
Great work! The summarization usecase feels amazing.

eva/third_party/huggingface/model.py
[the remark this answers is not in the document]
gaurav274: This PR is based on applying UDF on a list of audio files. Any reason for not using the audio engine?

eva/udfs/abstract/hf_abstract_udf.py
tushar-97: For automatic speech recognition, this will be like {'text':<transcript_from_video>}>. The default code doesn't really handle it, plus this way looks cleaner. Will add an example

eva/udfs/abstract/hf_abstract_udf.py
tushar-97: Correct. I actually added default args for the ASR model

eva/third_party/huggingface/create.py
tushar-97: I agree its repetitive, but this just looks cleaner and makes it more obvious when we do operations on the list/dict. Renaming the dictionary can be a fair compromise. Will do.

eva/parser/lark_visitor/_functions.py
tushar-97: I agree, but strings are returned as ConstantValueExpression objects, and integers are returned as int objects. I don't want to change the existing methods which is why this check is necessary.

eva/third_party/huggingface/model.py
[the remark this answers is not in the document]
tushar-97: It was due to issues with batching when using decord reader. Fixed it now

PR #667, opened 2023-04-24
tutorial: toxicity classifier

hhh21u:


jarulraj:
@hhh21u Where is the HF-based UDF? I see only detoxify-based UDFs:

`https://github.com/unitaryai/detoxify/releases/download/`

jarulraj:
Thanks for the PR! Subsumed in #708

PR #677, opened 2023-04-28
feat: Adding Yolov8

aryan-rajoria:
Adds yolov8


eva/udfs/yolo_object_detector.py
```diff
@@ -41,10 +44,14 @@ class YoloV5(PytorchAbstractClassifierUDF):
     def name(self) -> str:
         return "yolo"
 
-    def setup(self, threshold=0.85):
+    def setup(self, threshold=0.60):
         logging.getLogger("yolov5").setLevel(logging.CRITICAL)  # yolov5
         self.threshold = threshold
-        self.model = torch.hub.load("ultralytics/yolov5", "yolov5s", verbose=False)
+        self.predict_func = YOLO("yolov8s.pt")
```
gaurav274, line 50: use more accurate model

aryan-rajoria: Okay, should I use the Medium one or large model?

gaurav274:
We need to change `YoloV5` in the entire code base

eva/udfs/yolo_object_detector.py
```diff
@@ -157,26 +159,29 @@ def forward(self, frames: Tensor) -> pd.DataFrame:
...
+
+        for pred in predictions:
+            single_result = pred.boxes
+            pred_class = [
+                self.predict_func.names[i] for i in single_result.cls.tolist()
+            ]
+            pred_score = single_result.conf.tolist()
+            pred_boxes = single_result.xyxy.tolist()
+
+            sorted_list = list(map(lambda i: i < self.threshold, pred_score))
```
gaurav274, line 177: get rid of the threshold from here and let us use confidence to control it in queries. Btw, is there a default value?

gaurav274: If there is no default value, then we can stick with 0.3.

aryan-rajoria: there is no default value for the model

eva/udfs/decorators/yolo_object_detection_decorators.py
```diff
@@ -46,7 +46,7 @@ def name(self) -> str:
     @setup(cachable=True, udf_type="object_detection", batchable=True)
     def setup(self, threshold=0.85):
         self.threshold = threshold
-        self.model = torch.hub.load("ultralytics/yolov5", "yolov5s", verbose=False)
+        self.model = torch.hub.load("ultralytics/Yolo", "Yolos", verbose=False)
```
gaurav274, line 49: Also, change this to new code. Btw, can you get rid of older code and let us only use decorators?

aryan-rajoria: Yes done

gaurav274:
#679

PR #679, opened 2023-04-29
feat: support for Yolov8 pipelines

gaurav274:
- [x] Simplify the creation yolo based UDFs

```
CREATE UDF IF NOT EXISTS Yolo
TYPE  ultralytics
'model' 'yolov8m.pt';
```

- [x] Speed up testing by using a smaller Yolo model


eva/executor/create_udf_executor.py
```diff
@@ -46,6 +48,12 @@ def handle_huggingface_udf(self):
             self.node.metadata,
         )
 
+    def handle_ultralytics_udf(self):
+        """Handle Ultralytics UDFs"""
+        # manually set the impl_path for yolo udfs
+        self.node._impl_path = Path(f"{EVA_DEFAULT_DIR}/udfs/yolo_object_detector.py")
```
suryatejreddy, line 54: Adding decorators to this file should fix the decorators problem. Because the create_udf_executor will read decorators from here and update it in the catalog.

suryatejreddy (commented):
I think the overall design looks okay to me (probably biased because it's similar to HF). I'm not sure what the right Database fundamentals are for the multiple if else inside the create_udf_executor.

eva/executor/create_udf_executor.py
```diff
@@ -46,6 +48,12 @@ def handle_huggingface_udf(self):
             self.node.metadata,
         )
 
+    def handle_ultralytics_udf(self):
```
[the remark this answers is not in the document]
gaurav274: It is too much engineering for what we have right now. Maybe in the future if we really need it.

eva/third_party/ultralytics/binder.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def assign_yolo_udf(udf_obj: UdfCatalogEntry):
+    """
+    Assigns the correct yolo model to the UDF. The model is provided as part of the metadata
+    """
+    try:
+        model = get_metadata_entry(udf_obj, "model")[1]
+    except Exception:
+        model = "yolov8m.pt"
```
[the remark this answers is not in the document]
gaurav274: I don't understand this. My assumption is if nothing is provided we fall to `yolov8m.pt`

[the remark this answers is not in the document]
suryatejreddy: This was our discussion related to letting the user just mention CREATE UDF TYPE ultralytics TASK CLASSIFICATION/DETECTION/SEGMENTATION but for now we can leave it as is.

eva/third_party/ultralytics/binder.py
```diff
@@ -0,0 +1,44 @@
...
+from eva.catalog.models.udf_metadata_catalog import UdfMetadataCatalogEntry
+from eva.third_party.utils import get_metadata_entry
+from eva.udfs.yolo_object_detector import Yolo
+
+
+def assign_yolo_udf(udf_obj: UdfCatalogEntry):
+    """
+    Assigns the correct yolo model to the UDF. The model is provided as part of the metadata
+    """
+    try:
```
suryatejreddy, line 27: nit: probably can have a function get_metadata_entry_or_val(udf_obj, key, default_value) in utils.py. I imaging this to be a common use case throughout eva, even for batching/caching.

suryatejreddy: Now that I think about it, this should be probably not be in third_party but somewhere inside utils. Because it has to do with metdata.

eva/third_party/ultralytics/binder.py
```diff
@@ -0,0 +1,44 @@
...
+        model = "yolov8m.pt"
+
+    return lambda: Yolo(model)
+
+
+def parse_yolo_args(metadata_list: List[UdfMetadataCatalogEntry]):
+    """
+    Parse the metadata information associated with yolo and it as dictionary
+    """
+    model = "yolov8m.pt"
```
suryatejreddy, line 39: seems like duplicate code for get_metadata_entry, can we merge?

eva/executor/create_udf_executor.py
```diff
@@ -118,3 +123,69 @@ def exec(self, *args, **kwargs):
         yield Batch(
             pd.DataFrame([f"UDF {self.node.name} successfully added to the database."])
         )
+
+    def _try_initializing_udf(
```
suryatejreddy, line 127: 👍

eva/executor/create_udf_executor.py
```diff
@@ -118,3 +123,69 @@ def exec(self, *args, **kwargs):
...
+
+        Raises:
+            RuntimeError: If an error occurs while initializing the UDF.
+        """
+        # load the udf class from the file
+        try:
+            # loading the udf class from the file
+            udf = load_udf_class_from_file(impl_path, self.node.name)
+            # initializing the udf class calls the setup method internally
+            udf(**udf_args)
```
suryatejreddy, line 149: I'm not sure I completely understnad this. It seems like you are passing in the model as  "yolov8m.pt". But that is also happening inside the lambda function that you are returning. this means we are essentially doing `Yolo("yolov8m.pt")("yolov8m.pt")`?.

eva/executor/create_udf_executor.py
```diff
@@ -118,3 +123,69 @@ def exec(self, *args, **kwargs):
...
+        except Exception as e:
+            err_msg = f"Error creating UDF: {str(e)}"
+            logger.error(err_msg)
+            raise RuntimeError(err_msg)
+
+        return udf
+
+    def _resolve_udf_io(self, udf: UdfCatalogEntry) -> List[UdfIOCatalogEntry]:
```
suryatejreddy, line 157: 👍

test/udfs/test_abstract_udf.py
```diff
@@ -30,6 +31,9 @@ def test_udf_abstract_functions(self, mock_pipeline):
         derived_udf_classes = list(get_all_subclasses(AbstractUDF))
         # Go over each derived class of AbstractUDF
         for derived_udf_class in derived_udf_classes:
+            # skip yolo to avoid downloading model
```
suryatejreddy, line 34: I remember facing the same issue for HF also, so I mocked the function call, maybe you can do the same? Just mock the Yolo call?

eva/executor/create_udf_executor.py
gaurav274: Which lambda function?

suryatejreddy: Ohh wait my bad, this is loading the UDF from file.

eva/udfs/udf_bootstrap_queries.py
```diff
@@ -185,8 +183,12 @@ def init_builtin_udfs(mode="debug"):
             ]
         )
 
-    if mode != "minimal":
-        queries.extend([YoloV5_udf_query])
+        if mode == "debug":
```
jarulraj, line 186: Why do we have multiple modes here?

setup.py
```diff
@@ -94,11 +94,11 @@ def read(path, encoding="utf-8"):
...
-    "easyocr>=1.5.0",          # OCR EXTRACTION
-    "ipython<8.13.0",          # NOTEBOOKS
-    "yolov5<=7.0.9",           # OBJECT DETECTION
-    "thefuzz",                 # FUZZY STRING MATCHING
-    "transformers==4.27.4"     # HUGGINGFACE
+    "easyocr>=1.5.0",  # OCR EXTRACTION
+    "ipython<8.13.0",
+    "ultralytics",          # OBJECT DETECTION
+    "thefuzz",                # FUZZY STRING MATCHINGz
+    "transformers==4.27.4",    # HUGGINGFACE
```
jarulraj, line 101: Do we only support this particular version of transformers?

test/benchmark_tests/conftest.py
```diff
@@ -12,21 +12,19 @@
...
 def setup_pytorch_tests():
     CatalogManager().reset()
     execute_query_fetch_all("LOAD VIDEO 'data/ua_detrac/ua_detrac.mp4' INTO MyVideo;")
     execute_query_fetch_all("LOAD VIDEO 'data/mnist/mnist.mp4' INTO MNIST;")
-    load_udfs_for_testing()
-    from eva.udfs.udf_bootstrap_queries import YoloV5_udf_query
-
-    execute_query_fetch_all(YoloV5_udf_query)
+    # Load all the release UDFs
+    # We want to use accurate yolo model
```
jarulraj, line 28: Can we specify the actual models in the comment -- one for release and another one for debug?

test/util.py
```diff
@@ -451,7 +451,7 @@ def create_dummy_4d_batches(
 
 
 def load_udfs_for_testing(mode="debug"):
-    # DEBUG OR RELEASE MODE: ALL UDFs
+    # DEBUG MODE: ALL UDFs
     # MINIMAL MODE: NO YOLO
```
jarulraj, line 455: Can we get rid of `minimal` mode?

jarulraj (approved):

eva/third_party/ultralytics/binder.py
gaurav274: right fixed!

eva/third_party/ultralytics/binder.py
gaurav274: Correct, fixed!

test/udfs/test_abstract_udf.py
gaurav274: I can't figure out mocking for yolo since it is loaded using a file path. Sticking with the current code isn't super important.

PR #682, opened 2023-04-29
feat: support for docker containerization

Bickor:
In this PR, we have added a Dockerfile that can be utilized to containerize the eva database, and several hugging face udfs to add functionality to the database. For example, the following models were registered as UDFs and tested: ehcalabres/wav2vec2-lg-xlsr-en-speech-emotion-recognition, jonatasgrosman/wav2vec2-large-xlsr-53-english, etc.


jarulraj:
The models in the notebook look useful. Thanks for sharing.

PR #683, opened 2023-04-29
feat: support for fine-tuning YOLO models using TUNE command

tangyuan233:
Created a new TUNE command and zip file support for existing LOAD command


jarulraj:
Thanks for the nice PR! @tangyuan233 We will look into merging it in the future as this gets into the "fine-tuning" space that EVA is not currently focused on.

PR #686, opened 2023-04-30
ci: reduce number of cache shards

gaurav274:
- [x] Fix memory failures on CI when running reuse test cases. An issue with `yolo` surfaces when the system is running on low memory. Explicitly calling garbage collection helped fix this issue.

### DIFF
--- a/eva/catalog/models/base_model.py
+++ b/eva/catalog/models/base_model.py
@@ -108,7 +108,7 @@ def init_db():
         logger.info("Database does not exist, creating database.")
         create_database(engine.url)
         logger.info("Creating tables")
-        BaseModel.metadata.create_all()
+        BaseModel.metadata.create_all(checkfirst=True)
 
 
 def drop_db():
--- a/eva/catalog/models/base_model.py
+++ b/eva/catalog/models/base_model.py
@@ -108,7 +108,7 @@ def init_db():
         logger.info("Database does not exist, creating database.")
         create_database(engine.url)
         logger.info("Creating tables")
-        BaseModel.metadata.create_all(checkfirst=True)
+        BaseModel.metadata.create_all()
 
 
 def drop_db():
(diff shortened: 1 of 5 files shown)


eva/utils/kv_cache.py
```diff
@@ -26,10 +26,10 @@ class DiskKVCache:
         `max_cache_size` (int, optional): an integer representing the maximum size of
             the cache. The system will make its best effort to enforce this limit, but it may slightly exceed it. The default value is 2**30.
         `shards` (int, optional): an integer representing the number of shards to use.
-            This can improve concurrent writes. The default value is 6.
+            This can improve concurrent writes. The default value is 3.
     """
 
-    def __init__(self, path: str, max_cache_size: int = 2**30, shards: int = 6):
+    def __init__(self, path: str, max_cache_size: int = 2**30, shards: int = 3):
```
jarulraj, line 32: This means 3 GB -- can we add a comment in the code itself?

eva/catalog/models/base_model.py
```diff
@@ -108,7 +108,7 @@ def init_db():
         logger.info("Database does not exist, creating database.")
         create_database(engine.url)
         logger.info("Creating tables")
-        BaseModel.metadata.create_all()
+        BaseModel.metadata.create_all(checkfirst=True)
```
jarulraj, line 111: What does this `checkfirst` option mean? Why is it now needed? We could add some comments here.

test/integration_tests/test_reuse.py
```diff
@@ -44,6 +45,7 @@ def tearDown(self):
         execute_query_fetch_all("DROP TABLE IF EXISTS DETRAC;")
 
     def _verify_reuse_correctness(self, query, reuse_batch):
+        gc.collect()
```
jarulraj, line 48: Should we add a comment here for why this is needed? Links to relevant pages would also be helpful.

jarulraj (changes requested):

eva/catalog/models/base_model.py
gaurav274: Removed. It is true by default.

eva/utils/kv_cache.py
gaurav274: Done

PR #701, opened 2023-05-06
bug: modify the catalog reset method 

gaurav274:
The reset method only clears the contents of the catalog table and preserves the existing schema.


xzdandy:
How do we reset the schema now? By deleting the `catalog_database_uri`?

gaurav274:
> How do we reset the schema now? By deleting the `catalog_database_uri`?

We used to drop the tables and also the database. I felt it is unnecessary and we should only clear the contents to the table. Hopefully, it is stable.

xzdandy (approved):

PR #706, opened 2023-05-08
bug: catalog reset 

gaurav274:
Drop table which are not part of catalog

### DIFF
--- a/eva/catalog/models/base_model.py
+++ b/eva/catalog/models/base_model.py
@@ -19,7 +19,7 @@ from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy_utils import create_database, database_exists
 
-from eva.catalog.sql_config import SQLConfig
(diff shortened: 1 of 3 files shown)


eva/catalog/models/base_model.py
```diff
@@ -113,13 +113,31 @@ def init_db():
...
+    engine = SQLConfig().engine
+    # reflect to refresh the metadata
+    BaseModel.metadata.reflect(bind=engine)
+    if database_exists(engine.url):
+        with contextlib.closing(engine.connect()) as con:
+            trans = con.begin()
+            for table in reversed(BaseModel.metadata.sorted_tables):
+                if table.name not in CATALOG_TABLES:
+                    if table.exists(con):
+                        table.drop(con)
```
jiashenC, line 142: Wondering, we don't need `con.execute` here?

gaurav274: It does not work for `drop` so had to do this way.

jiashenC (approved):

PR #712, opened 2023-05-09
fix: image loading

jiashenC:
* Update file path with hashing
* Not abort if subset of images are corrupted

### DIFF
--- a/eva/executor/load_multimedia_executor.py
+++ b/eva/executor/load_multimedia_executor.py
@@ -82,7 +82,9 @@ class LoadMultimediaExecutor(AbstractExecutor):
 
             # Get valid files.
             valid_files = [
-                str(path) for path, is_valid in zip(video_files, valid_bitmap) if is_valid
+                str(path)
+                for path, is_valid in zip(video_files, valid_bitmap)
+                if is_valid
             ]
 
             if not valid_files:
--- a/eva/executor/load_multimedia_executor.py
+++ b/eva/executor/load_multimedia_executor.py
@@ -12,7 +12,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import multiprocessing as mp
+
 from pathlib import Path
+from multiprocessing import Pool
 
 import pandas as pd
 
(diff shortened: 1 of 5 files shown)


jiashenC:
Profiling shows https://github.com/georgia-tech-db/eva/blob/5134f6a3d504d87be09e09f86637e8cb040c5112/eva/storage/abstract_media_storage_engine.py#L112 takes most the time during loading for many images. I think this is a very easy to parallelize function, but not sure whether it worth to try. From my experience, load data in other DB systems tend to be very long as well, but only happens one time, so no need aggressive engineering to optimize it.

jarulraj:
Parallelizing this would greatly improve user experience. But, how to do this safely without corrupting the storage engine?

jiashenC:
> Parallelizing this would greatly improve user experience. But, how to do this safely without corrupting the storage engine?

@jarulraj Profiling shows majority of time spent on the for loop here (https://github.com/georgia-tech-db/eva/blob/5134f6a3d504d87be09e09f86637e8cb040c5112/eva/storage/abstract_media_storage_engine.py#L116). It generates a new file path and create symbolic link from new file path to the original. So, I think it won't affect storage engine even if it is parallelized. 

We can do some exception handling for the parallel for loop as well. Last, we will do the database writes in a single process (https://github.com/georgia-tech-db/eva/blob/5134f6a3d504d87be09e09f86637e8cb040c5112/eva/storage/abstract_media_storage_engine.py#L128).

jiashenC:
Through more careful profiling, the loading time breakdown can be categorized to three components. Benchmark profiles loading 1M images, each has 400x400x3 dimension on a system with 128 CPU cores. 

1. Multimedia data validation ~1000 sec https://github.com/georgia-tech-db/eva/blob/de37afb78793f16e12ee09bdacd998c8e0f3ecbb/eva/executor/load_multimedia_executor.py#L62
2. File symbolic links creation ~250 sec https://github.com/georgia-tech-db/eva/blob/de37afb78793f16e12ee09bdacd998c8e0f3ecbb/eva/storage/abstract_media_storage_engine.py#L116
3. Update records to storage engine ~30 sec https://github.com/georgia-tech-db/eva/blob/de37afb78793f16e12ee09bdacd998c8e0f3ecbb/eva/storage/abstract_media_storage_engine.py#L128

I am thinking about parallelizing 1 and 2 here but not 3, so I don't think it will affect storage engine integrity.

I have also introduced parallel multimedia data validation which reduces the overhead from 1000 sec to 20 sec on 128 core system. 

Thoughts on issue of parallelizing component 2?

eva/executor/load_multimedia_executor.py
```diff
@@ -122,3 +131,17 @@ def _rollback_load(
...
+        file_path: Path,
+    ):
+        file_path = Path(file_path)
+        if validate_media(file_path, self.media_type):
+            return True
+        else:
+            err_msg = f"Load {self.media_type.name} failed due to invalid file {str(file_path)}"
+            # If image is corrupted, do not abord the entire loading process.
+            logger.critical(err_msg)
+            # raise ValueError(file_path)
```
jiashenC, line 146: @gaurav274 Thought on this? Should we just abort as what we did previously? Since, the media validation process is parallelized, so the abort happens much quickly. Or do we just silently pass those corrupted images but give a warning to users?

gaurav274: Will it be possible to return the files that we couldn't load?

jiashenC: We can return corrupted files, but then we need to raise exception after we process all files. Should I add this feature in this PR?

2 option sounds good to me. I think we can merge this PR for now by sticking to the original approach: raise exception and abort. I can file another PR to support 2 options.

gaurav274: Sounds good!

gaurav274: We can fix the build and merge this.

gaurav274 (approved):

PR #720, opened 2023-05-11
fix: similarity test

jiashenC:
* Foreign key is not correctly configured, which prevents dataset tables to be dropped.
* Storage read gets extra argument after storage engine change.
* Add test case just to make sure index creation and similarity search runs correctly.

### DIFF
--- a/test/integration_tests/test_similarity.py
+++ b/test/integration_tests/test_similarity.py
@@ -93,12 +93,18 @@ class SimilarityTests(unittest.TestCase):
             )
             base_img -= 1
 
+        # Create an actual image dataset.
+        load_image_query = f"LOAD IMAGE '{self.img_path}' INTO testSimilarityImageDataset;"
+        execute_query_fetch_all(load_image_query)
+
     def tearDown(self):
         shutdown_ray()
 
         drop_table_query = "DROP TABLE testSimilarityTable;"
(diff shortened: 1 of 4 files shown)


test/integration_tests/test_similarity.py
```diff
@@ -288,3 +296,15 @@ def test_should_not_do_faiss_index_scan_with_predicate(self):
...
+        create_index_query = """CREATE INDEX testFaissIndexImageDataset
+                                    ON testSimilarityImageDataset (DummyFeatureExtractor(data))
+                                    USING HNSW;"""
+        execute_query_fetch_all(create_index_query)
+        select_query = """SELECT data FROM testSimilarityImageDataset
+                            ORDER BY Similarity(DummyFeatureExtractor(Open("{}")), DummyFeatureExtractor(data))
+                            LIMIT 1;""".format(
+            self.img_path
+        )
+        execute_query_fetch_all(select_query)
```
xzdandy, line 310: Can we make this test case more meaningful? First, it does not assert on anything. Second it seems there is only one image in the `testSimilarityImageDataset `,  it always returns that image I assume. For example, can we have 2+ images loaded and check it returns the similar one?

xzdandy (approved):

PR #730, opened 2023-05-12
feat : load pdf

affan00733:
load pdf query 

1. added the load pdf changes to lark parser and executor statements files.
2. remaining part is to add the LoadPDFExecutor in the load_executor.py where we will be adding the logic to extract the text items from pdf and store in table


eva/catalog/catalog_utils.py
```diff
@@ -91,12 +92,29 @@ def get_image_table_column_definitions() -> List[ColumnDefinition]:
...
+    id: pdf id
+    data: pdf data
+    """
+    columns = [
+        ColumnDefinition(PDFColumnName.id.name, ColumnType.INTEGER, None, None),
+        ColumnDefinition(
+            PDFColumnName.data.name,
+            ColumnType.NDARRAY,
+            NdArrayType.UINT8,
+            (None, None, None),
```
xzdandy, line 106: Why the pdf data column is NDARRAY of UINT8?

affan00733: As the data will be the text content and the array will contains the paragraphs of the page in the array

xzdandy: Would that be `NdArrayType.STR`? On the other hand, pdf != text. Is the plan here to have an internal reader that only extracts the text from the pdf? A few more questions, it is completely fine if we haven't thought about it. Why we partition the text based on the paragraphs instead of sentences, words, or even the whole pdf as single string? Based the query, the needs can be different. And, do we have examples queries on how we plan to query pdf in EVA?

affan00733: yes it should be NdArrayType.STR, yes I had a discussion with professor joy where we will be  having  a internal reader that only extracts the text from the pdf.
yes , we can have the text based on sentences and words and also a single string based on pages.
we plan to use the EVA query with load pdf where it create a table which have the text items as per the requirements from the pdf

eva/executor/load_executor.py
```diff
@@ -37,6 +37,12 @@ def exec(self, *args, **kwargs):
         elif self.node.file_options["file_format"] == FileFormatType.CSV:
             executor = LoadCSVExecutor(self.node)
 
+        elif self.node.file_options["file_format"] == FileFormatType.PDF:
+            """
+                Executor for extractor from pdf
+            """
+            pass
```
gaurav274, line 44: So there isn't an executor for `PDF`?

gaurav274:
I assume this is WIP, as the PDF executor is missing. Also, you can consider adding a test case to load PDF. Thanks!

eva/executor/load_executor.py
affan00733: Yes, as of now it isn't been added. It is in-progress of development

eva/parser/types.py
```diff
@@ -62,6 +62,7 @@ class FileFormatType(EVAEnum):
     VIDEO  # noqa: F821
     CSV  # noqa: F821
     IMAGE  # noqa: F821
+    PDF
```
jarulraj, line 65: You will need to run the linter -- 
`bash script/test/test.sh -m "LINTER"`  
`bash script/test/test.sh -m "TEST"  `

jarulraj (changes requested):

eva/parser/types.py
affan00733: Sure, I will do it

affan00733: I had run the linter, which got executed successfully 
and from the below image we can see that the pdf is been added to the eva and it has been throwing error at the executor because as of now we haven’t provided any executor. The PDFexecutor is in progress and will be developed soon
image

affan00733:
From the below image we can see that the pdf is been added to the eva and it has been throwing error at the executor because as of now we haven’t provided any executor. The PDFexecutor is in progress and will be developed soon
image

eva/catalog/catalog_type.py
```diff
@@ -24,6 +24,7 @@ class TableType(EVAEnum):
     STRUCTURED_DATA  # noqa: F821
     VIDEO_DATA  # noqa: F821
     IMAGE_DATA  # noqa: F821
+    PDF_DATA  # noqa: F821
```
affan00733, line 27: Hello @gaurav274 @jarulraj 
because i have created the table type PDF_DATA I am getting the error :05-22-2023 00:29:21 ERROR [plan_executor:plan_executor.py:execute_plan:0177] Invalid table type PDF_DATA
when dropping the table , the error is associated at storage_engine.py as in the storage dict is not mentioned and it is required to pass it with corresponding table type and storage engine

affan00733: So for this the part in which i am got stuck is in storage engine so could please help me how I can solve it

affan00733:
Hello Team,
 I had push the latest changes , with a notebook at tutorials/load_pdf_test.ipynb from which we can test it for now till the test cases are written


image

affan00733:
Hello Team ,

added the test case to location integration_tests/test_load__pdf_executor/LoadExecutorTest/test_load_pdfs :
where in the test case provided the assert statement of checking the number of columns from the table and the number of pdf loaded in the table

affan00733:
Command Query `LOAD PDF "path_pdf" from table_name;`

test/integration_tests/test_load__pdf_executor.py
affan00733, line 1: Added test case for load pdf

affan00733 (commented):
Run Linter  -- Done

affan00733:
Hello Team,
@jarulraj  @gaurav274 

update the changes to load pdf query with pdf table schema in page level and add the corresponding test case related to it.

Could you please review my PR, as all the changes have been integrated

image

affan00733:
Hello Team , 

added the fix for getting the pdf data based on paragraph and added the corresponding test case related to it.

image

jarulraj:
subsumed in #769

PR #731, opened 2023-05-12
feat: extensible parallel execution

jiashenC:
Propose simpler design for operator-level parallelization. 

1. The parallelized plan becomes the inner plan of the exchange operator.
2. Modify execution to support construction of inner plan.
3. The parallelized operator connects with child operator and parent operator with input queue and output queue. 

Visual example of new design (https://www.dropbox.com/s/nej0d7ouknpnrr2/ray-design.jpg?dl=0)


eva/experimental/ray/executor/ray_remote.py
```diff
@@ -0,0 +1,61 @@
...
+def ray_parallel(
+    conf_dict: Dict[str, str],
+    executor: Callable,
+    input_queue: Queue,
+    output_queue: Queue,
+):
+    for k, v in conf_dict.items():
+        os.environ[k] = v
+
+    gen = executor(input_queue=input_queue)
```
xzdandy, line 50: For this to work, we need to add a `readRayInputQueueExecutor` at the leaf of the inner plan.

jiashenC: Maybe you meant this? https://github.com/georgia-tech-db/eva/blob/75554b29839b48dc469274204754fc72c39459a1/eva/experimental/ray/executor/exchange_executor.py#L62

xzdandy: Thanks for pointing it out. I missed that!

eva/experimental/ray/executor/ray_remote.py
```diff
@@ -0,0 +1,61 @@
...
+    for next_item in gen:
+        output_queue.put(next_item)
+
+
+@ray.remote(max_calls=1)
+def ray_pull(conf_dict: Dict[str, str], executor: Callable, input_queue: Queue):
+    for k, v in conf_dict.items():
+        os.environ[k] = v
+
+    for next_item in executor():
```
xzdandy, line 60: Do you think we should do an explicit `executor.ray_exec()` instead of using the `__call__`? In most cases, `ray_exec()` or the current `__call__` are identical to `exec()`. Say, to support partition in seq scan, we can have a different `ray_exec()`.

eva/experimental/ray/executor/exchange_executor.py
```diff
@@ -32,74 +33,70 @@ def __init__(self):
...
-        while is_top:
+
+        ray_wait_and_alert.remote([ray_pull_task], input_queue)
+        ray_wait_and_alert.remote(ray_parallel_task_list, output_queue)
+
+        while True:
             res = output_queue.get(block=True)
             if res is StageCompleteSignal:
                 break
             elif isinstance(res, ExecutorError):
```
xzdandy, line 99: Maybe we can find a better way to garbage collect the ray actors and handle errors. I will think about it. Let us know if you have any thoughts.

jiashenC: I see. Do you have some error cases example in mind, that may not be cleanly caught by current implementations?

eva/experimental/ray/executor/ray_remote.py
jiashenC: Yes, I think that is a good idea. @gaurav274

gaurav274: Are you suggesting having 2 different execution logics for seq scan, `ray_exec` and `exec`?

xzdandy: That is a possibility to support partition. We are not doing it in this pull request. I am suggesting we have an explicitly `ray_exec()` for readability. So later we modify the `ray_exec()` instead of `__call__` in case we want to have a different behavior when the executor is under ray.  By default, `ray_exec()` will `yield from exec()` in the abstract_executor like the current `__call__`

eva/experimental/ray/optimizer/rules/rules.py
```diff
@@ -55,28 +55,36 @@ def apply(self, before: LogicalExchange, context: OptimizerContext):
...
 
-    def apply(self, before: LogicalProject, context: OptimizerContext):
-        after = ProjectPlan(before.target_list)
-        for child in before.children:
-            after.append_child(child)
-        upper = ExchangePlan(
-            parallelism=1,
-            ray_conf={"num_gpus": 1} if Context().gpus else {"num_cpus": 1},
+    def apply(self, before: LogicalApplyAndMerge, context: OptimizerContext):
+        apply_plan = ApplyAndMergePlan(before.func_expr, before.alias, before.do_unnest)
```
xzdandy, line 71: Some functions we can not parallelize like trackers. We can add a new UDF attribute to the decorator and check here.

eva/experimental/ray/executor/ray_remote.py
jiashenC: I think it is a good idea. It could also be beneficial later if we want to extend to an operator that has more than one children.

eva/experimental/ray/optimizer/rules/rules.py
jiashenC: I see. Just curious, why tracker is not parallelizable?

eva/experimental/ray/executor/ray_remote.py
gaurav274: Can't we control this logic in `__call__`? I prefer keeping the logic in `ray_remote` clean.

eva/experimental/ray/optimizer/rules/rules.py
gaurav274: Tracker needs tuples in sequence. For it to run in parallel, we need to implement some nontrivial stitching code.

eva/experimental/ray/executor/ray_remote.py
jiashenC: Something like this

```python
def ray_remote_func(executor):
    executor(remote=True)

Executor:
def exec(remote=False):
    if remote:
        # remote logic
    else:
        # local logic
```

I guess either way should work.

gaurav274: Yes, I like this.

xzdandy: I feel that this makes the exec more complicated and contains both the logic for parallel and non-parallel execution. The idea is to do it in `__call__`, but I think it's better to have an explicit name `ray_exec()` .

gaurav274: Yes, my bad. I assumed it is in `__call__`

eva/optimizer/plan_generator.py
```diff
@@ -145,7 +144,7 @@ async def build(self, logical_plan: Operator):
             print("Optimizer timed out!")
 
         # Only run post-processing if Ray is enabled.
-        if ConfigurationManager().get_value("experimental", "ray"):
-            plan = self.post_process(plan)
+        # if ConfigurationManager().get_value("experimental", "ray"):
```
gaurav274, line 147: We can remove this code

gaurav274 (commented):
Please add the design docs as part of the readme inside ray. Also, add the PR description.

gaurav274:
Also, have a look at https://app.circleci.com/pipelines/github/georgia-tech-db/eva/3651/workflows/f416c4c4-0a26-46ee-a7d5-60e699edb9cd/jobs/17625 to see if we can make it work with `ray`.

eva/experimental/ray/executor/ray_remote.py
```diff
@@ -0,0 +1,61 @@
...
+def ray_wait_and_alert(tasks: List[ray.ObjectRef], queue: Queue):
+    try:
+        ray.get(tasks)
+        queue.put(StageCompleteSignal)
+    except RayTaskError as e:
+        logger.error(e)
+        queue.put(ExecutorError(e.cause))
+
+
+@ray.remote(max_calls=1)
```
xzdandy, line 40: Why we need `max_calls = 1`?

test/optimizer/test_cascade_optimizer.py
```diff
@@ -91,6 +92,7 @@ def test_optimizer_post_process_strip_with_branch(self):
         self.assertTrue(isinstance(nest_plan.children[0], StoragePlan))
         self.assertTrue(isinstance(nest_plan.children[1], StoragePlan))
 
+    @pytest.mark.skip("No post-process tree")
     def test_optimizer_post_process_strip_without_branch(self):
```
xzdandy, line 96: We can just remove these two test cases. I don't think we need them in the future also?

jiashenC:
> Also, have a look at https://app.circleci.com/pipelines/github/georgia-tech-db/eva/3651/workflows/f416c4c4-0a26-46ee-a7d5-60e699edb9cd/jobs/17625 to see if we can make it work with `ray`.

@gaurav274 I think it naturally passes (https://app.circleci.com/pipelines/github/georgia-tech-db/eva/3671/workflows/95879b2d-6a6c-4a24-9aca-295d98e92ccb/jobs/17854). However, the reuse with multiple recurrence test case fails on CI. It does pass when I test on my machine though.

eva/experimental/ray/executor/ray_remote.py
xzdandy: Sounds perfect! Please add a comment to the code, so we do not lose this insight over time.

eva/experimental/ray/executor/exchange_executor.py
jiashenC: I see. I added a test case for error ray exception. The current print consists output from the main process and all remote processes. I remove the error logging of remote processes (which is just duplicate of the main process), so now it only prints one line of clean error. E.g., 

> 05-18-2023 10:32:02 ERROR [plan_executor:plan_executor.py:execute_plan:0177] The dataset file could not be found. Please verify that the file exists in the specified path.

And, I think the main process always returns one error message because it immediately stops when one ExecutorError is raised by any of remote processes.

jiashenC:
> Can we somehow set up the current integration tests to check the results are identical with/without ray?

Yeah, I added a test case for this as well. Ready for another review @gaurav274 @xzdandy

gaurav274:
LGTM

eva/experimental/ray/executor/ray_remote.py
```diff
@@ -0,0 +1,66 @@
...
+):
+    for k, v in conf_dict.items():
+        os.environ[k] = v
+
+    gen = executor(input_queue=input_queue)
+    for next_item in gen:
+        output_queue.put(next_item)
+
+
+@ray.remote(max_calls=1)
```
gaurav274, line 60: Please add here the comment you shared on slack

gaurav274 (approved):

gaurav274:
PLease make the doc image part of a readme on github. We can slowly expand it moving forward. Thanks!

eva/experimental/ray/executor/ray_remote.py
jiashenC: Which comment?

jiashenC:
> Please make the doc image part of a readme on github. We can slowly expand it moving forward. Thanks!

Should we have a separate readme or just add to the main readme?

gaurav274:
> > Please make the doc image part of a readme on github. We can slowly expand it moving forward. Thanks!
> 
> Should we have a separate readme or just add to the main readme?

Separate readme is better

eva/experimental/ray/executor/ray_remote.py
gaurav274: My bad, you shared here. Reason for `max_call=1`

jiashenC:
> > > Please make the doc image part of a readme on github. We can slowly expand it moving forward. Thanks!
> > 
> > 
> > Should we have a separate readme or just add to the main readme?
> 
> Separate readme is better

Ok. Fixed those. I also rename ray folder to parallel to be more specific.

xzdandy (approved):

xzdandy:
If `bfs` and `dfs` in abstract_plan.py are not used anymore in the new ray design, we can remove them.

PR #732, opened 2023-05-14
feat: OVERWRITE statement

yulaicui:
Overwrite statement allows users to modify table loaded by images or videos in place by udf.

query example:
OVERWRITE table BY UDF(data);

use case: 
1. Apply incremental updates using popular cv2-based image transformations
- OVERWRITE table BY ToGrayScale(data);
- OVERWRITE table BY GaussianBlur(data);

2. Draw bounding boxes computed by object detectors on frames
- OVERWRITE table BY Annotate(Yolov5(data));


xzdandy (changes requested):
Could you update the description of the pull request to give a brief summary on the overwrite statement? 
BTW, it seems that we don't have a meaningful test cases for overwrite. Could you improve that? Thanks!

yulaicui:
> Could you update the description of the pull request to give a brief summary on the overwrite statement? BTW, it seems that we don't have a meaningful test cases for overwrite. Could you improve that? Thanks!

summary update pending. For tests, there is eva/test/integration_tests/test_overwrite_executor.py that tests for overwriting both video table and image table. Did you mean to improve it? If so, could you provide some insights on how to do so? thanks!

test/integration_tests/test_overwrite_executor.py
```diff
@@ -0,0 +1,75 @@
...
+        execute_query_fetch_all(create_udf_query)
+
+        query = f"LOAD VIDEO '{video_file_path}' INTO MyVideo;"
+        execute_query_fetch_all(query)
+        overwrite_query = "OVERWRITE MyVideo BY GaussianBlur(data);"
+        execute_query_fetch_all(overwrite_query)
+
+        query = f"LOAD IMAGE '{image_file_path}' INTO MyImage;"
+        execute_query_fetch_all(query)
+        overwrite_query = "OVERWRITE MyImage BY GaussianBlur(data);"
```
xzdandy, line 47: Ah. I missed the query here! `Overwrite` is what we want to test here. Please move it out of the `setUpClass` and put them in separate test classes.

xzdandy (changes requested):

xzdandy:
Thanks for the description. Having a high-level question, If my understanding is correct, after the `overwrite` query, we lose the original table (data). I think in reality, users will more often want to keep the original data for traceability.

yulaicui:
> Thanks for the description. Having a high-level question, If my understanding is correct, after the `overwrite` query, we lose the original table (data). I think in reality, users will more often want to keep the original data for traceability.

@xzdandy regarding "we lose the original table (data)", it is true that the original data is no longer on the table, but the actual contents are still kept intact. It is just that they are "un-linked" from the table. For instance, assume ./data/video.mp4 is loaded to TABLE. the modified table will now has frames that link to the video file ./data/modified/video.mp4 instead so TABLE no longer has a tuple such that name field is ./data/video.mp4. However, the file ./data/video.mp4 is never deleted from the user's drive. It is designed this way to address the exact concern. 

A more inspiring use case is like this: assume we want to run 3 analysis on blurred frames TABLE: Analysis1, Analysis2, Analysis3 (all UDFs). Without overwrite, the queries looks like this:
SELECT Analysis1(GaussianBlur(data)) FROM TABLE;
SELECT Analysis2(GaussianBlur(data)) FROM TABLE;
SELECT Analysis3(GaussianBlur(data)) FROM TABLE;

With overwrite, it looks like this:
OVERWRITE TABLE BY GaussianBlur(data);
SELECT Analysis1(data) FROM TABLE;
SELECT Analysis2(data) FROM TABLE;
SELECT Analysis3(data) FROM TABLE;

Notice that the latter prevents the user from calling the UDF many times, and this is very common in modern data analytics such that we pre-process data the way we want before running analysis on the processed data.

test/integration_tests/test_overwrite_executor.py
yulaicui: resolved in new commit

eva/binder/statement_binder.py
```diff
@@ -178,6 +179,13 @@ def _bind_rename_table_statement(self, node: RenameTableStatement):
             node.old_table_ref.table.table_obj.table_type != TableType.STRUCTURED_DATA
         ), "Rename not yet supported on structured data"
 
+    @bind.register(OverwriteStatement)
+    def _bind_overwrite_statement(self, node: OverwriteStatement):
+        print("node: ", node)
```
xzdandy, line 184: remove print

eva/executor/overwrite_executor.py
```diff
@@ -0,0 +1,186 @@
...
+
+class OverwriteExecutor(AbstractExecutor):
+    def __init__(self, node: OverwritePlan):
+        super().__init__(node)
+        self.catalog = CatalogManager()
+        self.table_ref = self.node.table_ref
+        self.operation = self.node.operation
+
+    def make_new_video_path(self, video_path: str):
+        dirs = video_path.split("/")
```
xzdandy, line 40: Try using the standard library such as pathlib for path manipulation, instead of directly working with "/"

eva/executor/overwrite_executor.py
```diff
@@ -0,0 +1,186 @@
...
+
+        new_image_path = "/".join(dirs)
+        return new_image_path
+
+    def exec(self, *args, **kwargs):
+        assert type(self.operation) == FunctionExpression
+
+        table_obj = self.table_ref.table.table_obj
+        storage_engine = StorageEngine.factory(table_obj)
+        batch_mem_size = 30000000
```
xzdandy, line 61: We should read the batch mem size from eva.yml instead of hard coding it. To be consistent with other executor, maybe it is better to read and set it in the optimizer.

eva/executor/overwrite_executor.py
```diff
@@ -0,0 +1,186 @@
...
+        return new_video_path
+
+    def make_new_image_path(self, image_path: str):
+        dirs = image_path.split("/")
+        dirs[-1] = "modified_" + dirs[-1]
+
+        new_image_path = "/".join(dirs)
+        return new_image_path
+
+    def exec(self, *args, **kwargs):
```
xzdandy, line 56: I think we may reuse the existing executors like seqscan, function expression to achieve this instead of implementing everything from scratch. This allows optimizer to improve the query plan (e.g., parallelism, reuse). I can help you on that later.

eva/storage/image_storage_engine.py
```diff
@@ -36,3 +36,15 @@ def read(self, table: TableCatalogEntry) -> Iterator[Batch]:
...
                     yield batch
+
+    def clear(self, table: TableCatalogEntry, image_paths: List[Path]):
+        try:
+            media_metadata_table = self._get_metadata_table(table)
+            for image_path in image_paths:
+                dst_file_name = self._xform_file_url_to_file_name(Path(image_path))
+                image_file = Path(table.file_url) / dst_file_name
+                self._rdb_handler.clear(media_metadata_table)
+                image_file.unlink()
```
xzdandy, line 47: This removes the original file. We probably don't want it.

eva/storage/video_storage_engine.py
```diff
@@ -58,3 +58,15 @@ def read(
...
                     yield batch
+
+    def clear(self, table: TableCatalogEntry, video_paths: List[Path]):
+        try:
+            media_metadata_table = self._get_metadata_table(table)
+            for video_path in video_paths:
+                dst_file_name = self._xform_file_url_to_file_name(Path(video_path))
+                image_file = Path(table.file_url) / dst_file_name
+                self._rdb_handler.clear(media_metadata_table)
+                image_file.unlink()
```
xzdandy, line 69: 1. I think the variable name should be `video_file` instead of `image_file`
2. Similar to image, this removes the original video.

eva/udfs/ndarray/annotate.py
```diff
@@ -0,0 +1,68 @@
...
+            bboxes = row[2]
+
+            for bbox in bboxes:
+                x1, y1, x2, y2 = np.asarray(bbox, dtype="int")
+
+                x1, y1, x2, y2 = int(x1), int(y1), int(x2), int(y2)
+
+                frame = cv2.rectangle(frame, (x1, y1), (x2, y2), color, thickness)
+
+                # cv2.putText(
```
xzdandy, line 54: remove commented code

eva/udfs/ndarray/annotate.py
```diff
@@ -0,0 +1,68 @@
...
+import pandas as pd
+
+from eva.udfs.abstract.abstract_udf import AbstractUDF
+
+color = (207, 248, 64)
+thickness = 4
+
+
+class Annotate(AbstractUDF):
+    def setup(self):
```
xzdandy, line 26: I think we should fill the @setup and @forward decorator for these set of UDFs.

xzdandy (changes requested):

test/integration_tests/test_overwrite_executor.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # integration test for overwrite
+    def test_should_overwrite_video(self):
+        overwrite_query = "OVERWRITE MyVideo BY GaussianBlur(data);"
+        execute_query_fetch_all(overwrite_query)
+
+        select_query = "SELECT * FROM MyVideo;"
+        actual_batch = execute_query_fetch_all(select_query)
+        modified_dir = actual_batch.column_as_numpy_array(actual_batch.columns[1])[0]
+        self.assertTrue("tmp/modified/dummy.avi" in modified_dir)
```
xzdandy, line 68: We should check the video content is actually transformed by `GaussionBlur` instead of the file path. Same for the image test case.

xzdandy (changes requested):

eva/udfs/ndarray/annotate.py
```diff
@@ -0,0 +1,68 @@
...
+
+         Returns:
+             ret (pd.DataFrame): The modified frame.
+        """
+
+        def annotate(row: pd.Series) -> np.ndarray:
+            row = row.to_list()
+            frame = row[0]
+
+            bboxes = row[2]
```
xzdandy, line 45: How do we pass bounding box information in a OVERWRITE query?

xzdandy:
> > Thanks for the description. Having a high-level question, If my understanding is correct, after the `overwrite` query, we lose the original table (data). I think in reality, users will more often want to keep the original data for traceability.
> 
> @xzdandy regarding "we lose the original table (data)", it is true that the original data is no longer on the table, but the actual contents are still kept intact. It is just that they are "un-linked" from the table. For instance, assume ./data/video.mp4 is loaded to TABLE. the modified table now has frames that link to the video file ./data/modified/video.mp4 instead so TABLE no longer has a tuple such that name field is ./data/video.mp4. However, the file ./data/video.mp4 is never deleted from the user's drive. It is designed this way to address the exact concern.
> 
> A more inspiring use case is like this: assume we want to run 3 analysis on blurred frames TABLE: Analysis1, Analysis2, Analysis3 (all UDFs). Without overwrite, the queries looks like this: SELECT Analysis1(GaussianBlur(data)) FROM TABLE; SELECT Analysis2(GaussianBlur(data)) FROM TABLE; SELECT Analysis3(GaussianBlur(data)) FROM TABLE;
> 
> With overwrite, it looks like this: OVERWRITE TABLE BY GaussianBlur(data); SELECT Analysis1(data) FROM TABLE; SELECT Analysis2(data) FROM TABLE; SELECT Analysis3(data) FROM TABLE;
> 
> Notice that the latter prevents the user from calling the UDF many times, and this is very common in modern data analytics such that we pre-process data the way we want before running analysis on the processed data.

Thanks a lot of the detailed explanation and engineering efforts in this pr. I have gone over them and left some comments. 

Have we thought about creating a new table instead of overwriting the existing one (I think your current implementation removes the existing videos or images, which probably wants to change)? Creating a new video can achieves the similar user experience improvement.

Think about the annotate case. If annotation overwrites the original video, no further analysis and modification can be made (e.g., users find out a mistake/improvement after the annotation) on the video.

yulaicui:
> > > Thanks for the description. Having a high-level question, If my understanding is correct, after the `overwrite` query, we lose the original table (data). I think in reality, users will more often want to keep the original data for traceability.
> > 
> > 
> > @xzdandy regarding "we lose the original table (data)", it is true that the original data is no longer on the table, but the actual contents are still kept intact. It is just that they are "un-linked" from the table. For instance, assume ./data/video.mp4 is loaded to TABLE. the modified table now has frames that link to the video file ./data/modified/video.mp4 instead so TABLE no longer has a tuple such that name field is ./data/video.mp4. However, the file ./data/video.mp4 is never deleted from the user's drive. It is designed this way to address the exact concern.
> > A more inspiring use case is like this: assume we want to run 3 analysis on blurred frames TABLE: Analysis1, Analysis2, Analysis3 (all UDFs). Without overwrite, the queries looks like this: SELECT Analysis1(GaussianBlur(data)) FROM TABLE; SELECT Analysis2(GaussianBlur(data)) FROM TABLE; SELECT Analysis3(GaussianBlur(data)) FROM TABLE;
> > With overwrite, it looks like this: OVERWRITE TABLE BY GaussianBlur(data); SELECT Analysis1(data) FROM TABLE; SELECT Analysis2(data) FROM TABLE; SELECT Analysis3(data) FROM TABLE;
> > Notice that the latter prevents the user from calling the UDF many times, and this is very common in modern data analytics such that we pre-process data the way we want before running analysis on the processed data.
> 
> Thanks a lot of the detailed explanation and engineering efforts in this pr. I have gone over them and left some comments.
> 
> Have we thought about creating a new table instead of overwriting the existing one (I think your current implementation also removes the existing videos or images, which we probably want to change)? Creating a new table can achieves the similar user experience improvement.
> 
> Think about the annotate case. If annotation overwrites the original video, no further analysis and modification can be made (e.g., users find out a mistake/improvement after the annotation) on the video.

@xzdandy thanks for the feedback! Again. the video on the video never deleted from the hard drive. A new video is created and linked to the same table. The old video is still there and can be reloaded to the table. The executor just creates a new video on a new path. I think the alternative proposed in the comment can be achieved with just using UDF, which can basically be achieved with existing infrastructure without the OVERWRITE executor.

test/integration_tests/test_overwrite_executor.py
yulaicui: This integration check is designed to check if the OVERWRITE statement infrastructure works as a whole, which means checking if a new video is created and linked to the table. I'll add a check for UDF, but perhaps not as an integration test.

eva/storage/video_storage_engine.py
yulaicui: 1. Agree. will change in the next commit.
2. It doesn't remove the original video. It removes the video from the table but not the video on the user's drive.

eva/executor/overwrite_executor.py
yulaicui: I thought of borrowing logics from executors like sequential scan but it seems the workflow is a bit different. The idea is to collect all existing video/image paths and create modified versions of those files and then link them back. Are there ideas on how to fit this into existing executor workflows?

eva/udfs/ndarray/annotate.py
xzdandy: If `OVERWRITE TABLE with Annotate(data, Yolov5(data));` works now, let's have a test case for that.

test/integration_tests/test_overwrite_executor.py
xzdandy: Integration test should check the end-to-end is working, which is that we want the video to be transformed via `GaussionBlur`. On the other hand, the fact that we store in the `tmp/modified/dummy.avi` is one implementation option we choose (but not the only one). So checking the `tmp/modified/dummy.avi` can be unitest case for our implementation choices, and the test can use mock to avoid really loading videos and running the UDF.

jarulraj:
Can we illustrate `OVERWRITE` command in the MNIST notebook in a separate PR?

xzdandy:
> > > > Thanks for the description. Having a high-level question, If my understanding is correct, after the `overwrite` query, we lose the original table (data). I think in reality, users will more often want to keep the original data for traceability.
> > > 
> > > 
> > > @xzdandy regarding "we lose the original table (data)", it is true that the original data is no longer on the table, but the actual contents are still kept intact. It is just that they are "un-linked" from the table. For instance, assume ./data/video.mp4 is loaded to TABLE. the modified table now has frames that link to the video file ./data/modified/video.mp4 instead so TABLE no longer has a tuple such that name field is ./data/video.mp4. However, the file ./data/video.mp4 is never deleted from the user's drive. It is designed this way to address the exact concern.
> > > A more inspiring use case is like this: assume we want to run 3 analysis on blurred frames TABLE: Analysis1, Analysis2, Analysis3 (all UDFs). Without overwrite, the queries looks like this: SELECT Analysis1(GaussianBlur(data)) FROM TABLE; SELECT Analysis2(GaussianBlur(data)) FROM TABLE; SELECT Analysis3(GaussianBlur(data)) FROM TABLE;
> > > With overwrite, it looks like this: OVERWRITE TABLE BY GaussianBlur(data); SELECT Analysis1(data) FROM TABLE; SELECT Analysis2(data) FROM TABLE; SELECT Analysis3(data) FROM TABLE;
> > > Notice that the latter prevents the user from calling the UDF many times, and this is very common in modern data analytics such that we pre-process data the way we want before running analysis on the processed data.
> > 
> > 
> > Thanks a lot of the detailed explanation and engineering efforts in this pr. I have gone over them and left some comments.
> > Have we thought about creating a new table instead of overwriting the existing one (I think your current implementation also removes the existing videos or images, which we probably want to change)? Creating a new table can achieves the similar user experience improvement.
> > Think about the annotate case. If annotation overwrites the original video, no further analysis and modification can be made (e.g., users find out a mistake/improvement after the annotation) on the video.
> 
> @xzdandy thanks for the feedback! Again. the video on the video never deleted from the hard drive. A new video is created and linked to the same table. The old video is still there and can be reloaded to the table. The executor just creates a new video on a new path. I think the alternative proposed in the comment can be achieved with just using UDF, which can basically be achieved with existing infrastructure without the OVERWRITE executor. The whole purpose of this PR is to have EVA to support in-place update on the table without removing the actual files on a user's hard drive. Like any traditional DBMS, a statement that UPDATES something is expected to modify the table and it is left to the user to exercise caution when making in-place changes

Sorry for the confusions. Let me step back. Let's say we support the following query `CREATE TABLE blurredVideo AS SELECT GaussianBlur(data) FROM Video`.  And if we want overwrite, we can allow `CREATE TABLE Video AS SELECT GaussianBlur(data) FROM Video`. I think this will be more generic and useful.

jarulraj:
@xzdandy  Any reason why the `OVERWRITE` command is not as good as `CREATE TABLE Video AS SELECT GaussianBlur(data) FROM Video`?

xzdandy:
> @xzdandy Any reason why the `OVERWRITE` command is not as good as `CREATE TABLE Video AS SELECT GaussianBlur(data) FROM Video`?
> 
> This syntax seems simpler to use: `OVERWRITE table BY ToGrayScale(data);`

@jarulraj In my opinion,  exploration is one main use case for database system. More often user does not know exactly what they want beforehand.  Under that assumption, from a user's perspective, `CREATE TABLE blurredVideo AS SELECT GaussianBlur(data) FROM Video` is more useful than `CREATE TABLE Video AS SELECT GaussianBlur(data) FROM Video`. 
Let's say user wants to try for two blur functions, if we use overwrite proposed, that means we need load the video again in the middle. Furthermore, if storage is not a concern, I do not see why user wants to get rid of the original video. 

Secondly, `CREATE TABLE Video AS SELECT GaussianBlur(data) FROM Video` can be composite with all other queries we supported in EVA (e.g., where) seamlessly.  So even if we want to support `OVERWRITE` for simplicity, i think it should also be a sugar syntax for `CREATE TABLE Video AS SELECT GaussianBlur(data) FROM Video`.

PR #733, opened 2023-05-14
docs: enable spellcheck on all python files, notebooks, and docs 

jarulraj:
We use `aspell`, `codespell`, and `pyenchant` with the list of ignored words located in a file -- `spelling.txt'

`aspell`: rst and python files
`pyenchant`: notebook markdown cells
`codespell`: rst and python files

We currently do not use `pylint` as it is slow and not very useful.


script/formatting/formatter.py
```diff
@@ -386,3 +413,19 @@ def check_file(file):
         if file.endswith(".ipynb"):
             notebook_file = os.path.join(EVA_NOTEBOOKS_DIR, file)
             check_notebook_format(notebook_file)
+
+    # CODESPELL
+    LOG.info("Codespell")
+    os.system("codespell eva/")
```
gaurav274, line 419: I feel we shouldn't break the build for spelling errors in code. What do you think?

script/formatting/formatter.py
```diff
@@ -386,3 +413,19 @@ def check_file(file):
...
+    os.system("codespell docs/source/*/*.rst")
+    os.system("codespell docs/source/*.rst")
+    os.system("codespell *.md")
+    os.system("codespell eva/*.md")
+
+    # GO OVER ALL DOCS
+    LOG.info("aspell")
+    for elem in Path(EVA_DOCS_DIR).rglob('*.*'):
+        if elem.suffix == ".rst":
+            os.system(f"aspell --lang=en --personal=spelling.txt check {elem}")
```
gaurav274: Breaking for errors in documentation makes sense.

gaurav274: It looks like we need to run the command in a noninteractive way.

script/formatting/formatter.py
jarulraj: This will break the build. However, aspell is an interactive checker -- so it won't exit -- will just keep waiting. Locally, we will fix things interactively with the `LINTER` mode.

script/formatting/formatter.py
jarulraj: There seems to be no easy way to run it in a noninteractive way.

gaurav274 (approved):

PR #735, opened 2023-05-15
feat: Ray redesign

kaushikravichandran:
EVA on Ray is currently unstable because we use Exchanges. The current design has the following limitations:

Introduction of exchange operator is hard-coded
Splits in plans cannot be handled by exchanges
Exchanges are not context free, as a top-level exchange depends on lower exchanges
Although there can be a lot of operators in between two exchanges that are run on Ray, only UDFs benefit
Costing is not easy
This PR removes the use of Exchange operator and localizes the use of ray to only FunctionExpressions. This has the following advantages:

The above problems are eliminated as we no longer need an Exchange operator as ray is just used as a distributed execution alternative for FunctionExpression
Use of Ray is localized only to FunctionExpressions
We cost FunctionExpressions, and can use this information to guide heuristics for ray
Todo:
[] Analysis (preliminary analysis showed 25% improvement of running FastRCNNObjectDetector on UADETRAC when increasing parallelism from 2 to 16)
[] Code cleanup
[] Tests

### DIFF
--- a/eva/eva.yml
+++ b/eva/eva.yml
@@ -23,11 +23,13 @@ storage:
   image_engine: "eva.storage.image_storage_engine.ImageStorageEngine"
 server:
   host: "0.0.0.0"
-  port: 5432
+  port: 7213
   socket_timeout: 60
 
 experimental:
   ray: False
+  ray_predicate: True
+  ray_parallelism: 8
 
 third_party:
   openai_api_key: ""
--- a/eva/eva.yml
+++ b/eva/eva.yml
@@ -23,13 +23,13 @@ storage:
   image_engine: "eva.storage.image_storage_engine.ImageStorageEngine"
 server:
   host: "0.0.0.0"
-  port: 7213
+  port: 5432
   socket_timeout: 60
 
 experimental:
   ray: False
-  ray_predicate: True
-  ray_parallelism: 8
+  ray_predicate: False
+  ray_parallelism: 16
 
 third_party:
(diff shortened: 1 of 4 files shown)


gaurav274:
@xzdandy @jiashenC I feel this overlaps with your new design for `ray.` What do you think?

xzdandy:
> @xzdandy @jiashenC I feel this overlaps with your new design for `ray.` What do you think?

If I understand correctly, the idea proposed here is to directly create ray actors in the executor for the predicate's expression and projection's target_list. This is different from Jiashen's idea in two aspects.
1. This approach does not align with the optimizer design, where we pull the function expression into apply, and it becomes the unit for reuse, ray parallelism rules. 
2. This approach makes every ray actor as a pipeline block, where we get all batches from the child and then create ray actors to parallelize among those batches.

jiashenC:
I haven't enabled parallelization rule for predicate operator in my PR yet https://github.com/georgia-tech-db/eva/pull/731. Maybe we can either enable that in my current PR or add a new PR later as a complement to this current PR?

xzdandy:
> I haven't enabled parallelization rule for predicate operator in my PR yet #731. Maybe we can either enable that in my current PR or add a new PR later as a complement to this current PR?

I think your current PR, which matches the applymerge is fine. Later, we can have another rule that pulls the UDF expression from predicate into applymerge. Then it will work with your PR.

PR #738, opened 2023-05-15
feat: allow users to specify host and port while launching server

gaurav274:
1. Allow users to specify hostname and port while launching the server
2. Change the default port to 8803 to avoid conflict with Postgres.


script/test/package.sh
```diff
@@ -1,10 +1,18 @@
 #!/bin/bash
 
+## backward compatibility
+EVA_VERSION=$(pip show evadb | grep Version | awk '{print $2}')
+if [[ "$EVA_VERSION" < "0.2.4" ]]; then
```
jarulraj, line 5: Good decision

eva/server/server.py
```diff
@@ -38,6 +38,7 @@ async def start_eva_server(self, host: string, port: int):
         port: port of the server
         """
         logger.info("Start Server")
+        print(f"EVA Server started at host {host} and port {port}", flush=True)
```
jarulraj, line 41: Is there a reason for using `print` instead of logger here?

eva/server/interpreter.py
```diff
@@ -56,9 +56,8 @@ async def read_from_client_and_send_to_server(
     stdin_reader: StreamReader, writer: StreamWriter, server_reader: StreamReader
 ):
     VERSION = VERSION_DICT["VERSION"]
-    intro = "eva (v " + VERSION + ')\nType "EXIT;" to exit the client' + "\n"
-    sys.stdout.write(intro)
-    sys.stdout.flush()
+    intro = f"eva (v{VERSION})\nType 'EXIT;' to exit the client \n"
+    print(intro, flush=True)
```
jarulraj, line 60: Is there a reason for using `print` instead of logger or sys.stdout here?

jarulraj (approved):

jarulraj:
Can we add a test case as well -- https://coveralls.io/builds/59972132/source?filename=eva%2Feva_cmd_client.py#L49

eva/server/server.py
gaurav274: Logger wasn't displaying anything. Maybe we have to set the level to `info`, wasn't sure if we want to do it.

eva/server/interpreter.py
gaurav274: `print` is just a wrapper around `sys.stdout`.

xzdandy (approved):

PR #750, opened 2023-05-19
release: automating the release script

jarulraj:
- [x] releaser.py


script/releasing/releaser.py
```diff
@@ -149,22 +303,140 @@ def release_version(NEXT_RELEASE):
...
+    changelog = get_changelog(release_date)
+    print(changelog)
+
+    # UPDATE CHANGELOG
+    append_changelog(changelog, current_version_str_without_dev)
+
+    # RELEASE VERSION
+    release_version(current_version_str_without_dev)
+
+    # PUBLISH WHEELS ON PYPI
```
gaurav274, line 434: We should not automate this to avoid accidentally uploading to PyPi (cannot be reverted)

gaurav274 (approved):

PR #751, opened 2023-05-19
feat: integration with langchain

gaurav274:
Introducing support for `langchain` document loaders! With this enhancement, each file is treated as a single document, aligning with the `langchain` design. For instance, when loading a PDF file with 16 pages, a single tuple will be returned for the entire file. 

Command: `LOAD DOCUMENT 'data/documents/*.pdf' INTO pdfs;`


jarulraj:
We need to add some dependencies in `setup.py`?

eva/executor/executor_utils.py
```diff
@@ -91,11 +92,17 @@ def validate_video(video_path: Path) -> bool:
...
+def validate_document(doc_path: Path) -> bool:
+    return doc_path.suffix in LOADER_MAPPING
+
+
 def validate_media(file_path: Path, media_type: FileFormatType) -> bool:
     if media_type == FileFormatType.VIDEO:
         return validate_video(file_path)
     elif media_type == FileFormatType.IMAGE:
         return validate_image(file_path)
+    elif media_type == FileFormatType.DOCUMENT:
```
jiashenC, line 104: I saw langchain support many different types of documents. Are we going to validate them all? And what is the standard way of validating those documents?

eva/readers/document/document_reader.py
```diff
@@ -0,0 +1,33 @@
...
+        super().__init__(*args, **kwargs)
+
+    def _read(self) -> Iterator[Dict]:
+        ext = Path(self.file_url).suffix
+        assert ext in LOADER_MAPPING, f"File Format {ext} not supported"
+        loader_class, loader_args = LOADER_MAPPING[ext]
+        loader = loader_class(self.file_url, **loader_args)
+        # load entire document as one entry
+        for data in loader.load():
+            yield {"data": data.page_content}
```
xzdandy, line 33: Is ` data.page_content` always a python `str` regardless of the loader? Type hint here will be useful for understanding.

test/integration_tests/test_load_executor.py
```diff
@@ -496,6 +496,14 @@ def test_parallel_load_should_raise_exception(self):
         # Clean up large scale image directory.
         shutil.rmtree(large_scale_image_files_path)
 
+    def test_load_pdfs(self):
+        execute_query_fetch_all(
+            f"""LOAD DOCUMENT '{EVA_ROOT_DIR}/data/documents/*.pdf' INTO pdfs;"""
+        )
+        result = execute_query_fetch_all("SELECT * from pdfs;")
+        self.assertEqual(len(result.columns), 3)
+        self.assertEqual(len(result), 2)
```
xzdandy, line 505: I think you did not push the pdf files. I assume we have two pdf files?

eva/executor/executor_utils.py
gaurav274: I don't plan to validate all documents. Just the ones we support in our minimal version

eva/readers/document/document_reader.py
gaurav274: Yes, as per the `langchain` documentation. I'll add a reference.

eva/executor/executor_utils.py
jarulraj: We can use -- https://pypi.org/project/python-magic/ -- in a future PR to check file types beyond just suffix matching for security

jarulraj (approved):

PR #752, opened 2023-05-22
fix: restart eva_server process for every tutorial

jiashenC:
Currently, eva_server is not restarted when running a tutorial. In this case, if ~/.eva directory is deleted or modified by user, the eva_server process enters a not clean state.

### DIFF
--- a/tutorials/00-start-eva-server.ipynb
+++ b/tutorials/00-start-eva-server.ipynb
@@ -1,6 +1,7 @@
 {
  "cells": [
   {
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


jarulraj:
I did this to speed up CI for notebooks. Why do things not work even if all the test cases pass?

jarulraj (approved):

jiashenC:
> I did this to speed up CI for notebooks. Why do things not work in the notebook even if all the test cases pass?

It won't affect test cases across different tutorials. But when users delete or modify `~/.eva` directory without restarting the `eva_server` process, I guess the process may cache database data in memory (e.g., image path) even though the dataset is already gone.

PR #766, opened 2023-05-23
feat: change default dir to "eva_data" and make it configurable

jarulraj:
`eva_server --database eva_data_2 --port 3424 --host localhost`


jarulraj:
It's not a super elegant change. We should clean up this PR before merging.

gaurav274 (approved):

Issue #739, opened 2023-05-16
Improve CREATE TABLE query for better usability 


xzdandy:
### Search before asking

- [X] I have searched the EVA DB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar feature requests.

### Description

We'd like introduce a `CREATE TABLE` query to support common modifications (e.g., blur, flip, annotate) on a video/image table. The plan is as follows:
- [x] Add common modification UDFs from #732 
- [x] Simplify the query so that user does not need to provide a column name list. 
- [x] Rename current `CREATE MATERIALIZED VIEW` to `CREATE TABLE`
- [ ] Summarize and fix `SELECT` queries that are currently not supported with `CREATE TABLE` 
- [ ] ~~Support writing numpy array as video or image in _StorageEngine_. Reuse the implementation from #732~~ 
- [ ] ~~Integrating the new _StorageEngine_ with the `CREATE TABLE` executor~~

### Use case

We will be able to support following queries after accomplishing this feature:
- Q1: `CREATE TABLE BlurredVideo AS SELECT GaussianBlur(data) FROM Video;`
- Q2: `SELECT UDF1(data) FROM BlurredVideo;`
- Q3: `SELECT UDF2(data) FROM BlurredVideo;`
- Q4: `CREATE TABLE AnnotatedVideo AS SELECT Annotate(data, Yolo(data).bboxes) FROM Video;`
- Q5: `CREATE TABLE Video AS SELECT GaussianBlur(data) FROM Video;`

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

xzdandy:
We have serialization support in sqlite_storage_engine.py. https://github.com/georgia-tech-db/eva/blob/c526f78ae25d7d54e7661fd0af0d16e062d1fba6/eva/storage/sqlite_storage_engine.py#L51

So the step `Support writing numpy array as video or image in StorageEngine` and `Integrating the new StorageEngine with the CREATE TABLE executor` become optional. We will use the serialization support in sqlite_storage_engine.py for now. If we meet performance issues, we will rethink the need to offer a specialized writer for video/images. 

On the other hand, @yulaicui found a few queries are not supported with the current `CREATE MATERIALIZED VIEW` (or `CREATE TABLE` after renaming). Yulai will summarize those queries with error messages, and we will prioritize and fix them.

jarulraj:
@yulaicui This query in motif mining notebook is not working -- 

[...]

PR #768, opened 2023-05-23
feat: create materialized view infers column name

yulaicui:
Per issue #739 
Create materialized view is able infer column names. Column name is therefore optional on the user end.

Before:
CREATE MATERIALIZED VIEW dummy_view **(id, label)** AS SELECT id, DummyObjectDetector(data).label FROM MyVideo;

After:
CREATE MATERIALIZED VIEW dummy_view AS SELECT id, DummyObjectDetector(data).label FROM MyVideo;

### DIFF
--- a/test/integration_tests/test_mat_executor.py
+++ b/test/integration_tests/test_mat_executor.py
@@ -54,18 +54,36 @@ class MaterializedViewTest(unittest.TestCase):
         execute_query_fetch_all("DROP TABLE UATRAC;")
 
     def test_should_mat_view_with_dummy(self):
-        materialized_query = """CREATE MATERIALIZED VIEW dummy_view (id, label)
+        materialized_query = """CREATE MATERIALIZED VIEW dummy_view0 (id, label)
             AS SELECT id, DummyObjectDetector(data).label FROM MyVideo;
         """
         execute_query_fetch_all(materialized_query)
 
-        select_query = "SELECT id, label FROM dummy_view;"
+        select_query = "SELECT id, label FROM dummy_view0;"
         actual_batch = execute_query_fetch_all(select_query)
         actual_batch.sort()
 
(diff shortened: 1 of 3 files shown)


jarulraj:
@yulaicui We could add some test cases for scenarios for additional and fewer columns as well in a separate PR: 

```CREATE MATERIALIZED VIEW dummy_view (id, label, additional_column) AS SELECT id, DummyObjectDetector(data).label FROM MyVideo;```

File: test/integration_tests/test_mat_executor.py [hunk 1]
```diff
@@ -54,18 +54,36 @@ def tearDownClass(cls):
         execute_query_fetch_all("DROP TABLE UATRAC;")
 
     def test_should_mat_view_with_dummy(self):
-        materialized_query = """CREATE MATERIALIZED VIEW dummy_view (id, label)
+        materialized_query = """CREATE MATERIALIZED VIEW dummy_view0 (id, label)
```
jarulraj, line 57: Any reason for these changes `dummy_view` -> `dummy_view0`?

jarulraj (approved):

test/integration_tests/test_mat_executor.py
yulaicui: I was trying to distinguish the view names used for tests that have column names and the tests that don't have column name. It has no reason besides having a set of consistent names

jarulraj: I see, we could maybe get rid of the numbers as they serve no purpose. Ideally we should support both types of queries, right?

yulaicui: Agree. the numbers are a bit confusing and I'll come up with better names. And yes, this PR is already able to infer column types. What about I subsume the naming fixes to my next PR that adds more test cases?

PR #772, opened 2023-05-25
doc: start troubleshoot section

yulaicui:
This PR adds a trouble shooting section to doc that list common errors with corresponding fixes. Reviewers are welcomed to suggest any more issues and fixes.

### DIFF
--- a/docs/trouble_shoot/trouble_shoot.rst
+++ b/docs/trouble_shoot/trouble_shoot.rst
@@ -1,0 +1,68 @@
+Troubleshoot
+=============
+
+cannot install certain dependencies
+----
+
+fix #1
+~~~~
+
+Make ensure that local python version >= 3.7.0
+
+
+eva_server fails
+----
+
+fix #1
+~~~~
+
+Check if a process is already running on the target port (port_number). If so, terminate the process with process_id running on the port:
+
+.. code-block:: bash
+
+    sudo lsof -i :<port_number>
+    kill -9 <process_id>
+
+fix #2
+~~~~
+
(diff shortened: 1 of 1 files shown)


yulaicui:
I just created this doc to gather problems and their fixes: [https://docs.google.com/document/d/1K6Jvf66uT3INseqK2qxG67c5q9y0MU_i42FOke_3ItU/edit?usp=sharing]

Feel free to add contents to this document and I'll compile them

jiashenC:
- Computation cannot run on GPUs

docs/trouble_shoot/trouble_shoot.rst
```diff
@@ -0,0 +1,68 @@
...
+    kill -9 <process_id>
+
+fix #2
+~~~~
+
+Remove eva's running directory. To do this, open a new terminal and run
+
+.. code-block:: bash
+
+    rm -rf .eva
```
xzdandy, line 33: This is out of date? I think the latest version of eva does not use `.eva`?

xzdandy:
For some issues like `eva_client stuck at a simple query without responding`. I think we should fix it to improve the usability instead of putting them in the trouble shooting section.

docs/trouble_shoot/trouble_shoot.rst
gaurav274: Yes, we moved to `eva_data`.

aryan-rajoria:
> For some issues like `eva_client stuck at a simple query without responding`. I think we should fix it to improve the usability instead of putting them in the trouble shooting section.

We could

jarulraj:
subsumed in #801

PR #774, opened 2023-05-25
Improve create mat

xzdandy:


eva/binder/statement_binder.py
```diff
@@ -151,7 +152,34 @@ def _bind_delete_statement(self, node: DeleteTableStatement):
...
+            else:
+                raise BinderError("Unsupported expression type {}.".format(expr.etype))
+
+        assert (
+            len(node.col_list) == 0 or len(node.col_list) == num_projected_columns
+        ), "Projected columns mismatch, expected {} found {}".format(len(node.col_list), num_projected_columns)
+        binded_col_list = []
+        idx = 0
+        for expr in node.query.target_list:
+            output_objs = [(expr.col_name, expr.col_object)] if expr.etype == ExpressionType.TUPLE_VALUE else zip(expr.projection_columns, expr.output_objs)
```
yulaicui, line 171: just made a commit here. `expr.col_object.name` does not always infer the correct column name (i.e. 'labels' != 'label'). `expr.col_name` fixed this.

xzdandy: Thanks. Maybe I missed. What is the `expr.col_object.name` then?

yulaicui: I'm not sure where it is set. It seems that the object's name attribute doesn't always match column name. For instance, the previous failed test had object name = "labels" but object column name = "label", and "label" should be the correct one passed all the way down to sql storage engine.

xzdandy: @gaurav274 Is this a BUG or intended behavior?

yulaicui (approved):

eva/binder/statement_binder.py
```diff
@@ -151,7 +152,36 @@ def _bind_delete_statement(self, node: DeleteTableStatement):
...
+        ), "Projected columns mismatch, expected {} found {}.".format(len(node.col_list), num_projected_columns)
+        binded_col_list = []
+        idx = 0
+        for expr in node.query.target_list:
+            output_objs = [(expr.col_name, expr.col_object)] \
+                if expr.etype == ExpressionType.TUPLE_VALUE \
+                else zip(expr.projection_columns, expr.output_objs)
+            for col_name, output_obj in output_objs:
+                binded_col_list.append(
+                    ColumnDefinition(
```
gaurav274, line 176: Why is the bound column a ColumnDefinition? There is nothing to bind here. We should change the variable name to avoid confusion. Thanks!

eva/binder/statement_binder.py
```diff
@@ -151,7 +152,36 @@ def _bind_delete_statement(self, node: DeleteTableStatement):
...
+                num_projected_columns += len(expr.output_objs)
+            else:
+                raise BinderError("Unsupported expression type {}.".format(expr.etype))
+
+        assert (
+            len(node.col_list) == 0 or len(node.col_list) == num_projected_columns
+        ), "Projected columns mismatch, expected {} found {}.".format(len(node.col_list), num_projected_columns)
+        binded_col_list = []
+        idx = 0
+        for expr in node.query.target_list:
```
gaurav274, line 170: This logic is convoluted (too much nesting). It can be simplified. Can you point to the case where we cannot only rely on `objs`? They are required to be always correct.

PR #777, opened 2023-05-25
feat: Hugging face entity extraction

affan00733:
In the hugging face pipeline added the add on functionality of named entity recognition (NER) models


jarulraj:
@suryatejreddy Any feedback on this PR?

eva/third_party/huggingface/create.py
```diff
@@ -46,6 +46,7 @@
     "image-to-text": HFInputTypes.IMAGE,
     "object-detection": HFInputTypes.IMAGE,
     "depth-estimation": HFInputTypes.IMAGE,
+    "ner": HFInputTypes.TEXT,
```
suryatejreddy, line 49: nit: Move it along with the other `HFInputTypes.TEXT` types.

eva/third_party/huggingface/create.py
```diff
@@ -61,11 +62,12 @@
     "image-to-text": ImageHFModel,
     "object-detection": ImageHFModel,
     "depth-estimation": ImageHFModel,
+    "ner": TextHFModel,
```
suryatejreddy, line 65: Same as above.

eva/third_party/huggingface/create.py
```diff
@@ -122,7 +124,7 @@ def infer_output_name_and_type(**pipeline_args):
 
     # Get a dictionary of output names and types from the output
     output_types = {}
-    if isinstance(model_output, list):
+    if isinstance(model_output, list) and model_output != []:
```
suryatejreddy, line 127: Do we need this additional check? I see you changed the default example to include entities, so model_output would never be equal to `[]`. You can do away with this check probably.

eva/udfs/abstract/hf_abstract_udf.py
```diff
@@ -80,11 +80,14 @@ def output_formatter(self, outputs: Any):
             return pd.DataFrame(outputs, index=[0])
         # PERF: Can improve performance by avoiding redundant list creation
         result_list = []
-        for row_output in outputs:
-            # account for the case where we have more than one prediction for an input
-            if isinstance(row_output, list):
-                row_output = {k: [dic[k] for dic in row_output] for k in row_output[0]}
-            result_list.append(row_output)
+        if outputs != [[]]:
```
suryatejreddy, line 83: I think this can be made cleaner by returning an empty Dataframe when NER model returns empty list.

eva/udfs/abstract/hf_abstract_udf.py
```diff
@@ -80,11 +80,14 @@ def output_formatter(self, outputs: Any):
...
+            for row_output in outputs:
+                # account for the case where we have more than one prediction for an input
+                if isinstance(row_output, list):
+                    row_output = {k: [dic[k] for dic in row_output] for k in row_output[0]}
+                result_list.append(row_output)
+        elif self.pipeline_args["task"] == "ner":
+            result_list.append({"entity": "", "score": 0, 'index': 0, 'end': 0, 'word': 0, 'start': 0})
```
suryatejreddy, line 90: Why are we hardcoding this here? The above code should take care of it automatically right? Is there a case when it doesn't work?

test/integration_tests/test_huggingface_udfs_ner.py
```diff
@@ -0,0 +1,62 @@
+# coding=utf-8
```
suryatejreddy, line 1: This test case can be mvoed to the huggingface test file. Do not need a separate file for ner.

test/integration_tests/test_huggingface_udfs_ner.py
```diff
@@ -0,0 +1,62 @@
...
+        CatalogManager().reset()
+
+        query = """LOAD PDF 'data/documents/pdf_sample1.pdf' INTO MyPDFs;"""
+        execute_query_fetch_all(query)
+
+    def tearDown(self) -> None:
+        execute_query_fetch_all("DROP TABLE IF EXISTS MyPDFs;")
+
+    @pytest.mark.benchmark
+    def test_named_entity_recognition_model(self):
```
suryatejreddy, line 41: Also add a test case for when there are no named entities found in a text document. We want to ensure that EVA exits gracefully in that case.

suryatejreddy (changes requested):
@affan00733 left some comments. cc @jarulraj

eva/udfs/abstract/hf_abstract_udf.py
affan00733: by returning the empty dataframe it dosen't work it gives unknown columns finds error

eva/udfs/abstract/hf_abstract_udf.py
affan00733: it is not taking care of it automatically , it is giving error of unknown column found as shown in below image 

image

eva/third_party/huggingface/create.py
affan00733: made this changes and pushed

eva/third_party/huggingface/create.py
affan00733: made this changes and pushed

eva/third_party/huggingface/create.py
affan00733: made this changes and pushed

test/integration_tests/test_huggingface_udfs_ner.py
affan00733: made this changes and pushed and the test case name is test_named_entity_recognition_model_all_pdf_data

test/integration_tests/test_huggingface_udfs_ner.py
affan00733: made this changes and pushed and the test case name is test_named_entity_recognition_model_no_ner_data_exists

affan00733:
> @affan00733 left some comments. cc @jarulraj

@suryatejreddy added some comments and fixed all the issues reported and added the addition test case.
Only the remaining part is for ner where we are adding the hardcoded condition , because it is giving error of unknown column found. Regarding this added the related images in the block where review has been reported.

eva/udfs/abstract/hf_abstract_udf.py
jarulraj: @affan00733 Why is there an unknown column coming up with only the NER task? What happens with some other task like text classification? Why does this code not work: 

`row_output = {k: [dic[k] for dic in row_output] for k in row_output[0]}`

We should avoid specializing the code for a particular task as much as possible

eva/udfs/abstract/hf_abstract_udf.py
suryatejreddy: Yeah but in the sample case, we do have entities so I believe that shouldn't happen?

suryatejreddy:
@affan00733 could you try running again with the new sample text case that you added which actually has entities?

eva/udfs/abstract/hf_abstract_udf.py
affan00733: yes, In sample case we have the entities and where this unknown columns issue hasn't been occurred

eva/udfs/abstract/hf_abstract_udf.py
affan00733: For the same text classification it is working properly , none of the issues have been encountered

affan00733:
> @affan00733 could you try running again with the new sample text case that you added which actually has entities?

@suryatejreddy i have tried with new sample case where it is working and getting some entities , you can see in the below image

image

affan00733:
@jarulraj @suryatejreddy when carefully inspecting the issue I have found that unknown columns have occurred only when no entities have been detected. Taking that scenario I think that hardcoded block will be needed for ner

jarulraj:
@affan00733 Is it possible to generally handle scenario where entities have been detected -- instead of specializing it to just this task?

affan00733:
> @affan00733 Is it possible to generally handle scenario where entities have been detected -- instead of specializing it to just this task?

@jarulraj I had tried but couldn’t find a way, as when ner failed to detect the items then in batch.py in project function their has been an assert statement for the checking of unknown and verified columns as seen in image. where verified is columns detected by the ner model so for ner no items are their then it will be empty and unknown is columns from UDF frames that have been detected. As I haven’t had much understanding of batch.py file so @suryatejreddy could you please guide me in this scenario what possible solutions can be done.
image

jarulraj:
Fixed it by changing the code in `function_expression`. That is the right place to fix this \ cc @affan00733

affan00733:
> Fixed it by changing the code in `function_expression`. That is the right place to fix this \ cc @affan00733

Thanks 😊 professor, I will remove the existing hard coded code

jarulraj:
@gaurav274 @suryatejreddy Is this behavior expected when the UDF does not return any inputs?

jarulraj:
**Query:**

`SELECT data, {udf_name}(data)
                  FROM MyPDFs
                  WHERE page = 3
                  AND paragraph >= 1 AND paragraph <= 3;`

Output (**when entities are detected**)
image-1

Output (**when no entities are detected**)
image-2

PR #778, opened 2023-05-25
feat: saliency map

vivek-mandal:
Saliency Maps feature with updated UDF Design and test cases


jarulraj:
For context, @vivek-mandal is generalizing the implementation to an off-the-shelf Resnet feature extractor so that it is not tailored to only the MRI dataset.

jarulraj:
Closing this PR for now

PR #784, opened 2023-05-26
feat: support relational apis

gaurav274:
- [ ] Auto start-server : deferred to another PR
- [x] select
- [x] filter
- [x] load
- [x] create_vector_index
- [x] execute
- [x] order
- [x] limit
- [x] query
- [x] df


eva/interfaces/relational/db.py
```diff
@@ -23,18 +37,43 @@ def __init__(self, reader, writer):
...
+    ) -> EVARelation:
+        return self.cursor().load(file_regex, table_name, format, **kwargs)
+
+    def table(self, table_name: str) -> EVARelation:
+        return self.cursor().table(table_name)
+
+    def query(self, sql_query: str) -> EVARelation:
+        return self.cursor().query(sql_query)
+
+    def df(self) -> pandas.DataFrame:
```
jarulraj, line 59: Will `self.cursor().query(second_query)` will work even if the first query got a empty result?

eva/interfaces/relational/db.py
```diff
@@ -23,18 +37,43 @@ def __init__(self, reader, writer):
...
+        if not self._result:
+            raise Exception("No valid result with the current connection")
+        return self._result.frames
+
+    def create_vector_index(
+        self, index_name: str, table_name: str, expr: str, using: str
+    ) -> "EVACursor":
```
jarulraj, line 66: Why is this a string instead of a type?

eva/interfaces/relational/utils.py
```diff
@@ -0,0 +1,117 @@
...
+        "target_list",
+        "groupby_clause",
+        "orderby_list",
+        "limit_count",
+    ]
+
+    assert clause in accepted_clauses, f"Unknown clause: {clause}"
+
+    # If the clause being set is "target_list" and the value is equal to
+    # "*", the "SELECT *" portion is reset.
```
jarulraj, line 104: Why is it reset -- may be expand the comment?

jarulraj (approved):

eva/interfaces/relational/db.py
gaurav274: This is lazy, so you need to call `.df()` on the output to run the query.

eva/interfaces/relational/db.py
gaurav274: This is a type from the documentation standpoint. This is a way to refer to a class that hasn't been initialized yet. We can change it if you suggest another approach.

eva/interfaces/relational/relation.py
```diff
@@ -0,0 +1,220 @@
...
+        """Execute and fetch all rows as a pandas DataFrame
+
+        Returns:
+            pandas.DataFrame:
+        """
+        batch = self.execute()
+        assert batch is not None, "relation execute failed"
+        return batch.frames
+
+    def execute(self) -> Batch:
```
jiashenC, line 94: I understand this function is needed for trigger execution, but I am not sure whether return a `Batch` to users is useful.

Instead, I think providing a `show` function like other systems is more straightforward, which just directly prints results.

gaurav274: We have `.df()`. I earlier had `show` as well

eva/parser/utils.py
```diff
@@ -0,0 +1,60 @@
+from eva.parser.load_statement import LoadDataStatement
+
+from eva.parser.parser import Parser
+from eva.parser.select_statement import SelectStatement
+
+
+def parse_expression(expr: str):
+    mock_query = f"SELECT {expr} FROM DUMMY;"
```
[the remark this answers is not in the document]
gaurav274: We don't care about `dummy`. We just need the parsed equivalent of `expr`. The caller takes care of populating i.

eva/interfaces/relational/db.py
```diff
@@ -23,18 +37,43 @@ def __init__(self, reader, writer):
...
 
+    def load(
+        self, file_regex: str, table_name: str, format: str, **kwargs
+    ) -> EVARelation:
+        return self.cursor().load(file_regex, table_name, format, **kwargs)
+
+    def table(self, table_name: str) -> EVARelation:
+        return self.cursor().table(table_name)
+
+    def query(self, sql_query: str) -> EVARelation:
```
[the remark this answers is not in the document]
gaurav274: Sure 👍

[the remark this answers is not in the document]
gaurav274: `duckdb` has both. Will do the same

eva/interfaces/relational/utils.py
gaurav274: Done!

test/interfaces/relational/test_relational_api.py
```diff
@@ -0,0 +1,147 @@
...
+        self.images = f"{EVA_ROOT_DIR}/data/detoxify/*.jpg"
+
+    def tearDown(self):
+        shutdown_ray()
+        # todo: move these to relational apis as well
+        execute_query_fetch_all("""DROP TABLE IF EXISTS mnist_video;""")
+        execute_query_fetch_all("""DROP TABLE IF EXISTS meme_images;""")
+        # os.system("nohup eva_server --stop")
+
+    def test_relation_apis(self):
```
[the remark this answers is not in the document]
gaurav274: Done

eva/interfaces/relational/db.py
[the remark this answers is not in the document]
gaurav274: Done

eva/interfaces/relational/relation.py
gaurav274: Added `show()`

PR #787, opened 2023-05-28
feat: enable ray by default

xzdandy:
This pr intends to enable ray by default in EVA. 

- [x] Ray is enabled by default in eva server, test cases, notebooks  
- [x] Remove experimental code directory
- [ ] ~~Introduce a new UDF decorator `parralizable`  and uses it in the optimizer to filter out inexpensive functions.~~ 
   - This is not possible now. We need UDF decorator to be associated with the class instead of the setup function, so we can get the UDF properties without initializing the UDF.


xzdandy:
Currently, the optimizer rule for ray does not work for UDFs in the projection's target list. So only the following two notebooks uses ray. Their results look expected. 
- [tutorials/03-emotion-analysis.ipynb](https://github.com/georgia-tech-db/eva/blob/6b5a496bb3929c4d0a0047af42a2bf2b4eaa9af1/tutorials/03-emotion-analysis.ipynb)
- [tutorials/11-similarity-search-for-motif-mining.ipynb](https://github.com/georgia-tech-db/eva/blob/6b5a496bb3929c4d0a0047af42a2bf2b4eaa9af1/tutorials/11-similarity-search-for-motif-mining.ipynb)

Planning to add a new optimizer rule that extracts the UDF out of the projection's target list.

xzdandy:
@gaurav274 @jiashenC Any idea, why we have the ray skip tag for [chatgpt test cases](https://github.com/georgia-tech-db/eva/blob/97e91273d05f35c96b775203db97e0a5c7cfca56/test/integration_tests/test_chatgpt.py#LL67C21-L67C21)? Also, this seems to be unit test cases instead of integration test cases.

gaurav274:
> @gaurav274 @jiashenC Any idea, why we have the ray skip tag for [chatgpt test cases](https://github.com/georgia-tech-db/eva/blob/97e91273d05f35c96b775203db97e0a5c7cfca56/test/integration_tests/test_chatgpt.py#LL67C21-L67C21)? Also, this seems to be unit test cases instead of integration test cases.

It is a unit test because of the missing `openai_api_key.` Not sure about ray.

jiashenC:
> @jiashenC testcases are failing because of the coverage issue, right?

Yes, we probably still need to separate ray testing without generating coverage reports. Otherwise, it will cause timeout issue because of too many coverage trace to parse which are generated by many ray processes.

eva/experimental/parallel/optimizer/rules/rules.py
```diff
@@ -87,45 +86,39 @@ def apply(self, before: LogicalApplyAndMerge, context: OptimizerContext):
         yield exchange_plan
 
 
-class LogicalGetToSeqScan(Rule):
+class LogicalProjectToPhysical(Rule):
```
jiashenC, line 89: Just double check, so we won't get any functional expression in the SeqScan operator anymore?

jarulraj:
We can get rid of `experimental` folder?

eva/experimental/parallel/optimizer/rules/rules.py
xzdandy: Yes. It seems that the current statement to operator generation always creates a project operator.

xzdandy:
> > @jiashenC testcases are failing because of the coverage issue, right?
> 
> Yes, we probably still need to have a separate ray testing without generating coverage reports. Otherwise, it will cause timeout issue because of too many coverage trace to parse which are generated by many ray processes.

It is pending on local also, after all testcases are completed before the coverage report is generated. It seems that there may be some ray related resources are not correctly garbage cleaned.

xzdandy:
> We can get rid of `experimental` folder?

Make sense. We should get rid of it. Will do.

xzdandy:
**Update**:
- Rollback the UDF decorator changes for ray. The current UDF decorator design can not support this feature in a proper manner.
- After removing the experimental directory, the coverage is dropping, given we are collecting coverage with ray disabled. Need to update .coveragerc and add some unit ray testcases even when ray is disabled.

gaurav274 (approved):

jiashenC (approved):

PR #789, opened 2023-05-28
feat: create table from select query

yulaicui:
#739 
This PR enables creating tables from select query outputs.

Example:
assume a table loaded from video `MyVideo` and an object detector UDF `DummyObjectDetector` already exist, The query `CREATE TABLE dummy_table AS SELECT id, DummyObjectDetector(data).label FROM MyVideo;` should create a table of object detector results.


eva/parser/eva.lark
```diff
@@ -30,7 +30,7 @@ rename_table: RENAME TABLE table_name TO table_name
 create_udf: CREATE UDF if_not_exists? udf_name INPUT create_definitions OUTPUT create_definitions TYPE udf_type IMPL udf_impl udf_metadata* | CREATE UDF if_not_exists? udf_name IMPL udf_impl udf_metadata* | CREATE UDF if_not_exists? udf_name TYPE udf_type udf_metadata*
 
 // Create Materialized View
-create_materialized_view: CREATE MATERIALIZED VIEW if_not_exists? table_name ("(" uid_list ")") AS select_statement | CREATE MATERIALIZED VIEW if_not_exists? table_name AS select_statement
+create_materialized_view: CREATE MATERIALIZED VIEW if_not_exists? table_name ("(" uid_list ")")? AS select_statement
```
xzdandy, line 33: You can remove the create materialized code.

eva/optimizer/operators.py
```diff
@@ -547,6 +547,7 @@ class LogicalCreate(Operator):
         video {TableRef}: [video table that is to be created]
         column_list {List[ColumnDefinition]}:
         if_not_exists {bool}: [create table if exists]
+        query {SelectStatement}: Optional, only if new table is created by a query output
```
xzdandy, line 550: Shall we remove it, since the operator does not have the query as a member variable?

eva/optimizer/plan_generator.py
```diff
@@ -49,13 +49,15 @@ def build_optimal_physical_plan(
...
         root_grp = optimizer_context.memo.groups[root_grp_id]
         best_grp_expr = root_grp.get_best_expr(PropertyType.DEFAULT)
-        physical_plan = best_grp_expr.opr
 
-        for child_grp_id in best_grp_expr.children:
-            child_plan = self.build_optimal_physical_plan(
-                child_grp_id, optimizer_context
-            )
-            physical_plan.append_child(child_plan)
+        if best_grp_expr is not None:
```
xzdandy, line 53: For my education, in which case, the `best_grp_expr` can be None?

eva/parser/eva.lark
```diff
@@ -20,7 +20,7 @@ create_database: CREATE DATABASE if_not_exists? uid
 
 create_index: CREATE INDEX uid ON table_name index_elem vector_store_type?
 
-create_table: CREATE TABLE if_not_exists? table_name create_definitions 
+create_table: CREATE TABLE if_not_exists? table_name (create_definitions | (AS select_statement))
```
xzdandy, line 23: Do we plan to not support `CREATE TABLE if_not_exists? table_name create_definitions AS select_statement`? So we can still rename the column. It is fine that we do not include in this pr. Check the plan here.

xzdandy (approved):
The implementation looks clean! Thanks. Leave some comments. Please check before merging.

eva/optimizer/operators.py
yulaicui: agree. this can be deleted

eva/optimizer/plan_generator.py
yulaicui: It shouldn't happen if plans are implemented correctly. Will clean it up in the next commit.

eva/parser/eva.lark
yulaicui: Ideally yes.

eva/parser/eva.lark
yulaicui: agree. I'll make a separate PR for this

PR #791, opened 2023-05-29
bug: minor fixes for postgres connector

gaurav274:
👋 Thanks for submitting a Pull Request to EVA DB!

🙌 We want to make contributing to EVA DB as easy and transparent as possible. Here are a few tips to get you started:

- 🔍 Search existing EVA DB [PRs](https://github.com/georgia-tech-db/eva/pulls) to see if a similar PR already exists.
- 🔗 Link this PR to a EVA DB [issue](https://github.com/georgia-tech-db/eva/issues) to help us understand what bug fix or feature is being implemented.
- 📈 Provide before and after profiling results to help us quantify the improvement your PR provides (if applicable).

👉 Please see our ✅ [Contributing Guide](https://evadb.readthedocs.io/en/stable/source/contribute/index.html) for more details.

PR #795, opened 2023-05-30
Import config should persist nothing

gaurav274:
1. Removed singleton classes, `ConfigurationManager`, `SQLConfig`, and `CatalogManager`. They are now encapsulated within the `EVADatabase` class, which is owned by each server instance.
2. The Pythonic APIs no longer rely on server-client architecture. The current implementation allows the Pythonic APIs to run serverless. This design choice was made for stability, and we plan to reintroduce the server-client architecture. Please note that the command-line client still utilizes the server-client architecture.
3. For backward compatibility, we renamed `connect` to `connect_remote` for notebooks that use old server-client architecture.
4. The `eva.yml` file is no longer used. In the current implementation, configuration changes cannot be made post-launch. We have plans to introduce a `set` command to support configuration modifications after launch.


eva/executor/storage_executor.py
```diff
@@ -15,6 +15,7 @@
 from typing import Iterator
 
 from eva.catalog.catalog_type import TableType
+from eva.database import EVADB
```
jarulraj, line 18: We could use `EVADatabase` instead of `EVADB` to avoid confusion with the system name

jarulraj:
Does this PR also subsume #791?

gaurav274:
> Does this PR also subsume #791?

No

jiashenC (commented):
I actually like the previous approach that modules like CatalogManager, ConfiguraitonManagaer stays global and singleton and only initialized when called. So, we also do not need to pass instance around. Any particular reason that we adopt a new design?

xzdandy (approved):

PR #800, opened 2023-06-01
test: api testing  -> similarity between text and relevance keyword

jarulraj:
>  UDF : similarity between the paragraph from pdfs and relevant word passed through run time


jarulraj:
This PR looks good to me. We should rename `SimiliarityFeatureExtract` to `SpacyFeatureExtractor`. 

@jiashenC  Is there a difference in performance and accuracy between `SpacyFeatureExtractor` and `SentenceFeatureExtractor`?

jarulraj:
@gaurav274 @jiashenC The constructed similarity search query defaults to `ASC`. Why is it not sorted in descending order (`DESC`) based on similarity score? Intuitively, similarity score will be highest for similar items, right?

jarulraj:
This query works -- 

```
  query = (
            conn.table("PDFss")
            .order(
                """Similarity(
                    SpacyFeatureExtractor('When was the NATO created?'), SpacyFeatureExtractor(data)
                ) DESC"""
            )
            .limit(3)
            .select("data")
        )
```

In ascending order (default), it does not work:

```
  query = (
            conn.table("PDFss")
            .order(
                """Similarity(
                    SpacyFeatureExtractor('When was the NATO created?'), SpacyFeatureExtractor(data)
                ) """
            )
            .limit(3)
            .select("data")
        )
```

jiashenC:
> This PR looks good to me. We should rename `SimiliarityFeatureExtract` to `SpacyFeatureExtractor`.
> 
> @jiashenC Is there a difference in performance and accuracy between `SpacyFeatureExtractor` and `SentenceFeatureExtractor`?

I pick `sentence transformer` mainly because it does not introduce any more lib dependencies because we already have `huggingface`. `SpaCy` seems to support GPU as well, but it uses a completely different eco-system, so we may add a lot other libraries.

jarulraj:
Got it, then let's just stick with `SentenceFeatureExtractor`. @affan00733 Can you try running these two feature extractors and compare their accuracy and runtime?

affan00733:
> Got it, then let's just stick with `SentenceFeatureExtractor`. @affan00733 Can you try running these two feature extractors and compare their accuracy and runtime?

@jarulraj what specifically need to be done , is it to write a test case to get the features from SentenceFeatureExtractor and SpacyFeatureExtractor and get the similarity from these both features and show their respective result.
but as of now  SentenceFeatureExtractor is not available in master so what can be done

affan00733:
Also earlier after file name changing udf file is not available so test case are failing in checks , so fixed it

PR #804, opened 2023-06-02
feat: create udf interface for pythonic api

yulaicui:
create udf support for pythonic api

example: 
`CREATE UDF IF NOT EXISTS DummyObjectDetectorDecorators IMPL  'test/util.py';`

can be transformed to `conn.create_udf("DummyObjectDetector", "test/util.py")`


eva/interfaces/relational/db.py
```diff
@@ -86,6 +87,31 @@ def df(self) -> pandas.DataFrame:
             raise Exception("No valid result with the current connection")
         return self._result.frames
 
+    def create_udf(
+        self,
+        udf_name: str,
+        implementation: str,
+        type: str = None,
+        task: str = None,
```
gaurav274, line 95: Do you think we should make `task` and `model` part of kwargs since they are specific to huggingFace? @jarulraj

eva/interfaces/relational/db.py
```diff
@@ -279,6 +305,32 @@ def load(
...
+        """
+        Create a udf in the database.
+
+        Args:
+            udf_name (str): Name of the udf to be created.
+            implementation (str): Path string to udf's implementation
+            type (str): Type of the udf (e.g. HuggingFace).
+            task (str): ML task (only applicable if loading an ML model).
+            model (str): ML model.
+            **kwargs: Additional keyword arguments for configuring the load operation.
```
gaurav274, line 326: incorrect doc string

gaurav274 (commented):
Nicely done! Minor doc fixes. 

Also, add this function to the docs. https://github.com/georgia-tech-db/eva/tree/master/docs/source/reference/python 

Also, add the steps to add a new API in the developer docs. 

Thanks!

eva/interfaces/relational/db.py
yulaicui: I think it's a good idea

xzdandy: Not relevant to this pr. But I think we can support cachable, batchable, and parallelizable in the kwargs.

eva/parser/utils.py
```diff
@@ -39,6 +40,23 @@ def parse_table_clause(expr: str):
...
+    mock_query = f"CREATE UDF IF NOT EXISTS {udf_name}"
+    if type is not None:
+        mock_query += f" TYPE {type}"
+        if task is not None and model is not None:
+            mock_query += f" 'task' '{task}' 'model' '{model}'"
+    else:
+        mock_query += f" IMPL '{implementation}'"
+    mock_query += ";"
+
+    stmt = Parser().parse(mock_query)[0]
```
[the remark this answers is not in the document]
gaurav274: We need to take care of constructing Abstract Expressions etc. Plus, code maintenance.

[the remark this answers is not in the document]
xzdandy: It can be nontrivial to construct the original query if we have a very long .calls. For an unrealistic example,  .project().filter().apply().grouby().orderby().apply().project().filter().join()....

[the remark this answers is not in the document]
gaurav274: The current design is not constructing the entire query at once. It is rather incrementally constructing the query. I think it should be able to take care of it. The idea is quite similar to constructing the plan but I wanted to reuse the binder,  so went with this design. Do you think of any obvious issue?

eva/interfaces/relational/db.py
jarulraj: yes, we should make task and model part of `kwargs` -- `langchain` relies on `kwargs` a lot.

yulaicui: addressed in new commit

eva/parser/utils.py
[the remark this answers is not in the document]
xzdandy: I think you meant reuse parser. I guess the benefit is that we can do `print(mnist_query.sql_query())` which prints out the SQL query.

jarulraj:
@gaurav274 Can you stylistically review the `create_udf` test in `test_relational_api.py`?

eva/parser/utils.py
```diff
@@ -39,6 +40,22 @@ def parse_table_clause(expr: str):
     return stmt.from_table
 
 
+def parse_create_udf(udf_name: str, udf_file_path: str, type: str, **kwargs):
+    mock_query = f"CREATE UDF IF NOT EXISTS {udf_name}"
```
gaurav274, line 44: We should make `IF NOT EXISTS` a function argument.

eva/parser/create_udf_statement.py
```diff
@@ -93,7 +93,10 @@ def __str__(self) -> str:
                 metadata_str += f"{key}={value}, "
             metadata_str = metadata_str.rstrip(", ")
 
-        return f"CREATE UDF {self._name} INPUT ({input_str}) OUTPUT ({output_str}) TYPE {self._udf_type} IMPL {self._impl_path.name} ({metadata_str}))"
+        if self._impl_path:
+            return f"CREATE UDF {self._name} INPUT ({input_str}) OUTPUT ({output_str}) TYPE {self._udf_type} IMPL {self._impl_path.name} ({metadata_str}))"
```
gaurav274, line 97: There is no need to print `INPUT` and `OUTPUT.` We are anyways getting rid of it.

eva/parser/utils.py
yulaicui: in what situation would we not want IF NOT EXISTS when creating a UDF?

gaurav274: If the `UDF` already exists, we don't want this to throw an error but instead raise a warning. 
This is important in exploratory analysis when a user creates the UDF once but runs the Python script multiple times.

PR #810, opened 2023-06-03
feat: grouping paragraphs in documents and samples in audio

jarulraj:
Added support for 
`GROUP BY '<> samples'`
`GROUP BY '<> paragraphs'`


tutorials/01-mnist.ipynb
```diff
@@ -255,86 +384,42 @@
...
-       "4                          5  "
-      ]
-     },
-     "execution_count": 5,
-     "metadata": {},
-     "output_type": "execute_result"
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "SELECT data, MnistImageClassifier(data) FROM MNISTVid WHERE ((id = 30) OR ((id = 50) OR ((id = 70) OR ((id = 0) OR (id = 140)))))\n"
```
xzdandy, line 390: UDF is not created when using the new API.

xzdandy: Given the current Pythonic API, which is serverless. I updated the notebook to directly use `EVACursor`, instead of using `connect`, since current `connect` still tries to look for eva server. We also need to call `init_builtin_udfs` explictly. Maybe #795 will improve the situation. @gaurav274 Please check the usage of Pythonic API in the notebook.

jarulraj: @gaurav274 @xzdandy  With the auto-start server logic in #795, can we get rid of the need to call  `init_builtin_udfs` and just do `cursor = connect().cursor()` -- similar to https://dev.mysql.com/doc/connector-python/en/connector-python-api-mysqlcursor.html: 

```
import mysql.connector

cnx = mysql.connector.connect(database='world')
cursor = cnx.cursor()

PR #811, opened 2023-06-04
refactor: move eva -> evadb

gaurav274:
1. Moved `eva` -> `evadb`
2. Exposed `from evadb import connect` so now users can run the following commands.

```
import evadb 
conn = evadb.connect("eva_data")
```


jarulraj:
we will also need to rename this repo from `eva` to `evadb`

gaurav274:
Duplicate #820

PR #813, opened 2023-06-04
feat: Expose apis

gaurav274:
Exposed `from eva import connect` so now users can run the following commands.

```
import eva
conn = eva.connect("eva_data")
```


jarulraj:
Should we just get rid of the `start-server` notebook and use the new `connect` call in each notebook?

gaurav274:
> Should we just get rid of the `start-server` notebook and use the new `connect` call in each notebook?

Yes. What should we do about the `pip install ` ?

PR #814, opened 2023-06-04
feat: benchmark question answering v1

jiashenC:
Benchmark the embedding feature extraction step with parallelization on the pride and prejudice book.

Some issues that I discover that needs to be fixed:

- [ ] Batch size needs to be set manually, other texts are squeezed to one batch.
- [ ] GPU configurations need to be fixed.


jarulraj:
```
evadb.executor.executor_utils.ExecutorError: (builtins.LookupError) 'NDARRAY' is not among the defined enum values. Enum name: columntype. Possible values: BOOLEAN, INTEGER, FLOAT, ..., ANY
[SQL: INSERT INTO udfio_catalog (name, type, is_nullable, array_type, array_dimensions, is_input, udf_id) VALUES (?, ?, ?, ?, ?, ?, ?)]
[parameters: [{'is_nullable': False, 'array_type': <NdArrayType.STR: 11>, 'udf_id': 9, 'is_input': True, 'name': 'data', 'type': <ColumnType.NDARRAY: 5>, 'array_dimensions': '(None, 1)'}]]
```

PR #817, opened 2023-06-05
fix: GPU ids and degree of parallelism

jiashenC:
This PR addresses two issues.

1. We previously only allows parallelism when there is more than two GPUs available. I think this is too conservative. First, some functional expressions only run on CPUs and even parallelization of CPU functional expression can bring performance benefit. Additionally, even if there is only one GPU available, it still provide performance benefit when parallelization is enabled because of better GPU utilizations. Currently, I just hardcode the DOP, but I am thinking to allow users to configure this themselves through the `eva.yml`. Any feedback on this? @xzdandy @gaurav274 @jarulraj 

2. I realize a bug related to the previous approach. This will happen for more than 1 GPU case. I will give a concrete example here. Let's assume we have two GPUs. The GPU id that we get from the `Context` for the second GPU is `1` because context runs in the main process, which detects two GPUs on the PyTorch side (e.g., `GPU ID = [0, 1]`. However, within the Ray process, after we set the environmental variable `CUDA_VISIBLE_DEVICES=1`, the PyTorch indices is reset. Because it only accesses one GPU within the Ray process, it only gets GPU ids like `GPU ID = [0]`. Thus, it causes the no GPU error when it moves to GPU `to('cuda:1')`. The fix is very easy, that we just simply expose all GPUs to every Ray process, so we don't need to worry about PyTorch device id reset. To expose all GPUs, we just find the max GPU id from the context and expose all GPU ids that are lower than the max in the environmental variables.

### DIFF
--- a/evadb/executor/execution_context.py
+++ b/evadb/executor/execution_context.py
@@ -39,6 +39,7 @@ class Context:
 
     @property
     def gpus(self):
+        print("GPU Devices", self._gpus)
         return self._gpus
 
     def _populate_gpu_from_config(self) -> List:
--- a/evadb/executor/execution_context.py
+++ b/evadb/executor/execution_context.py
@@ -39,7 +39,6 @@ class Context:
 
     @property
     def gpus(self):
-        print("GPU Devices", self._gpus)
         return self._gpus
 
     def _populate_gpu_from_config(self) -> List:
--- a/evadb/optimizer/rules/rules.py
+++ b/evadb/optimizer/rules/rules.py
@@ -1220,6 +1220,20 @@ class LogicalVectorIndexScanToPhysical(Rule):
         yield after
 
 
+"""
+Rules to optimize Ray.
+"""
+
+
+def get_ray_env_dict():
+    # Get the highest GPU id and expose all GPUs that have id lower than
+    # the max id.
+    max_gpu_id = max(Context().gpus) + 1
+    return {
+        "CUDA_VISIBLE_DEVICES": ",".join([str(n) for n in range(max_gpu_id)])
+    }
+
+
 class LogicalExchangeToPhysical(Rule):
     def __init__(self):
         pattern = Pattern(OperatorType.LOGICALEXCHANGE)
@@ -1254,15 +1268,15 @@ class LogicalApplyAndMergeToRayPhysical(Rule):
     def apply(self, before: LogicalApplyAndMerge, context: OptimizerContext):
         apply_plan = ApplyAndMergePlan(before.func_expr, before.alias, before.do_unnest)
 
-        parallelism = 2 if len(Context().gpus) > 1 else 1
-        ray_parallel_env_conf_dict = [
-            {"CUDA_VISIBLE_DEVICES": str(i)} for i in range(parallelism)
-        ]
+        parallelism = 2
+
+        ray_process_env_dict = get_ray_env_dict()
+        ray_parallel_env_conf_dict = [ray_process_env_dict for _ in range(parallelism)]
 
         exchange_plan = ExchangePlan(
             inner_plan=apply_plan,
             parallelism=parallelism,
-            ray_pull_env_conf_dict={"CUDA_VISIBLE_DEVICES": "0"},
+            ray_pull_env_conf_dict=ray_process_env_dict,
             ray_parallel_env_conf_dict=ray_parallel_env_conf_dict,
         )
         for child in before.children:
@@ -1293,15 +1307,15 @@ class LogicalProjectToRayPhysical(Rule):
                 project_plan.append_child(child)
             yield project_plan
         else:
-            parallelism = 2 if len(Context().gpus) > 1 else 1
-            ray_parallel_env_conf_dict = [
-                {"CUDA_VISIBLE_DEVICES": str(i)} for i in range(parallelism)
-            ]
+            parallelism = 2
+
+            ray_process_env_dict = get_ray_env_dict()
+            ray_parallel_env_conf_dict = [ray_process_env_dict for _ in range(parallelism)]
 
             exchange_plan = ExchangePlan(
                 inner_plan=project_plan,
                 parallelism=parallelism,
-                ray_pull_env_conf_dict={"CUDA_VISIBLE_DEVICES": "0"},
+                ray_pull_env_conf_dict=ray_process_env_dict,
                 ray_parallel_env_conf_dict=ray_parallel_env_conf_dict,
             )
             for child in before.children:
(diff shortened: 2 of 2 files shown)


gaurav274:
> This PR addresses two issues.
> 
> 1. We previously only allows parallelism when there is more than two GPUs available. I think this is too conservative. First, some functional expressions only run on CPUs and even parallelization of CPU functional expression can bring performance benefit. Additionally, even if there is only one GPU available, it still provide performance benefit when parallelization is enabled because of better GPU utilizations. Currently, I just hardcode the DOP, but I am thinking to allow users to configure this themselves through the `eva.yml`. Any feedback on this? @xzdandy @gaurav274 @jarulraj
> 2. I realize a bug related to the previous approach. This will happen for more than 1 GPU case. I will give a concrete example here. Let's assume we have two GPUs. The GPU id that we get from the `Context` for the second GPU is `1` because context runs in the main process, which detects two GPUs on the PyTorch side (e.g., `GPU ID = [0, 1]`. However, within the Ray process, after we set the environmental variable `CUDA_VISIBLE_DEVICES=1`, the PyTorch indices is reset. Because it only accesses one GPU within the Ray process, it only gets GPU ids like `GPU ID = [0]`. Thus, it causes the no GPU error when it moves to GPU `to('cuda:1')`. The fix is very easy, that we just simply expose all GPUs to every Ray process, so we don't need to worry about PyTorch device id reset. To expose all GPUs, we just find the max GPU id from the context and expose all GPU ids that are lower than the max in the environmental variables.

1. I agree. We should have config settings for `numCPUCores` and `GPUs`. Regarding `DOP`, we can have a simple heuristic based on model size and latency. What do you think?

2. Sounds good!

jiashenC:
> > This PR addresses two issues.
> > 
> > 1. We previously only allows parallelism when there is more than two GPUs available. I think this is too conservative. First, some functional expressions only run on CPUs and even parallelization of CPU functional expression can bring performance benefit. Additionally, even if there is only one GPU available, it still provide performance benefit when parallelization is enabled because of better GPU utilizations. Currently, I just hardcode the DOP, but I am thinking to allow users to configure this themselves through the `eva.yml`. Any feedback on this? @xzdandy @gaurav274 @jarulraj
> > 2. I realize a bug related to the previous approach. This will happen for more than 1 GPU case. I will give a concrete example here. Let's assume we have two GPUs. The GPU id that we get from the `Context` for the second GPU is `1` because context runs in the main process, which detects two GPUs on the PyTorch side (e.g., `GPU ID = [0, 1]`. However, within the Ray process, after we set the environmental variable `CUDA_VISIBLE_DEVICES=1`, the PyTorch indices is reset. Because it only accesses one GPU within the Ray process, it only gets GPU ids like `GPU ID = [0]`. Thus, it causes the no GPU error when it moves to GPU `to('cuda:1')`. The fix is very easy, that we just simply expose all GPUs to every Ray process, so we don't need to worry about PyTorch device id reset. To expose all GPUs, we just find the max GPU id from the context and expose all GPU ids that are lower than the max in the environmental variables.
> 
>     1. I agree. We should have config settings for `numCPUCores` and `GPUs`. Regarding `DOP`, we can have a simple heuristic based on model size and latency. What do you think?
> 
>     2. Sounds good!

For DOP, I was just thinking a very simple approach. We can default to `2` in the config file to allow users to have some parallelization but the value is small enough to not cause any other issue (e.g., memory error, etc). For advanced uses, they can tune the parameter themselves based on their own system. I feel inferring DOP automatically is hard and also may cause system fail if DOP is not correctly inferred.

eva/optimizer/rules/rules.py
```diff
@@ -1262,6 +1262,21 @@ def apply(self, before: LogicalVectorIndexScan, context: OptimizerContext):
...
+Rules to optimize Ray.
+"""
+
+
+def get_ray_env_dict():
+    # Get the highest GPU id and expose all GPUs that have id lower than
+    # the max id.
+    if len(Context().gpus) > 0:
+        max_gpu_id = max(Context().gpus) + 1
+        return {"CUDA_VISIBLE_DEVICES": ",".join([str(n) for n in range(max_gpu_id)])}
```
xzdandy, line 1275: Shall we use the GPU ids specified in the eva.yml?

jiashenC: I assume you asked why I take max here?

I was thinking of a case that there may exist 3 GPUs [0, 1, 2] and the user wants to use [1, 2]. If we only do `CUDA_VISIBLE_DEVICES=1,2`, the device id on the pytorch side will become [0, 1] inside Ray. Then, it will error out when we do `.to(cuda:2`. 

So, I am thinking to expose all GPUs that are within the range of max id. And this is just to maintain a legit GPU id, but the system access to GPU will still be governed by `Context().gpus` from user setting, so system won't access GPU[0]. 

Let me know if that makes sense.

xzdandy: Get it.

xzdandy: Just come to my mind. By default, all GPUs are exposed to ray. It seems that is what we want. If so, can we just not set `CUDA_VISIBLE_DEVICES`?

jarulraj:
@jiashenC @xzdandy Are there some not hard-coded things from this PR we should add to the main repo?

jiashenC:
> @jiashenC @xzdandy Are there some not hard-coded things from this PR we should add to the main repo?

I think GPU id fix can be added. Parallelism anyway will stay hard coded for now.

jarulraj:
Got it, can you then update this PR so that it is suitable for merging (or maybe close this and raise a new PR if you prefer to leave these changes in history)?

xzdandy:
@jiashenC For my understanding, why we need `CUDA_VISIBLE_DEVICES`? Is that by default all GPUS are exposed to ray? If I misunderstand, please correct me. Thanks!

If we have a `SET` command, we can then have a simple SET DOP to change the Parallelism on fly instead of modifying the rule file. For now, parallelism will be hardcoded.

jiashenC:
> @jiashenC For my understanding, why we need `CUDA_VISIBLE_DEVICES`? Is that by default all GPUS are exposed to ray? If I misunderstand, please correct me. Thanks!
> 
> If we have a `SET` command, we can then have a simple SET DOP to change the Parallelism on fly instead of modifying the rule file. For now, parallelism will be hardcoded.

Sorry, missed the last msg.

My understanding is by default, none GPU is exposed to a ray process. When you do `@remote(num_gpus=1`, ray will assign a GPU to that ray process by setting `CUDA_VISIBLE_DEVICES=<some GPU id>`. Then, the similar issue will occur inside ray process that there will be only one GPU exposed inside ray and the GPU id from `Context` will be wrong.

PR #829, opened 2023-06-07
feat : langchain gpt4all QA

affan00733:
- loaded a pdf
- get the similarity by sentence transformer feature executor udf between question and pdf table -> top 3 results
- based on these results passed it to the GPT4ALL QA udf for answer


setup.py
```diff
@@ -54,10 +54,12 @@ def read(path, encoding="utf-8"):
...
     "boto3",
     "nest_asyncio",
-    "langchain",
+    "langchain==0.0.190",
     "pymupdf",
     "pdfminer.six",
-    "sentence-transformers"
+    "sentence-transformers",
+    "requests",
+    "gpt4all"
```
gaurav274, line 62: We can move this to `udf_libs`

evadb/udfs/GPT4ALL.py
```diff
@@ -0,0 +1,130 @@
...
+        output_signatures=[
+            PandasDataframe(
+                columns=["answers"],
+                column_types=[NdArrayType.STR],
+                column_shapes=[(1)],
+            )
+        ],
+    )
+    def forward(self, df: pd.DataFrame) -> pd.DataFrame:
+        def _forward(row: pd.Series) -> np.ndarray:
```
gaurav274, line 101: Please add a few comments to say what is going on here.

affan00733: sure 👍🏼

test/integration_tests/test_lanchain_llm.py
```diff
@@ -0,0 +1,89 @@
...
+    load_udfs_for_testing,
+    shutdown_ray,
+    suffix_pytest_xdist_worker_id_to_dir,
+)
+
+from evadb.configuration.constants import EVA_DATABASE_DIR, EVA_ROOT_DIR
+from evadb.interfaces.relational.db import connect
+
+
+class TestLangchainLLM(unittest.TestCase):
```
gaurav274, line 26: A notebook/example app would be better suited for this. We can compare the performance of not doing filtering vs. filtering.

test/integration_tests/test_lanchain_llm.py
```diff
@@ -0,0 +1,92 @@
...
+            "GPT4AllQaUDF",
+            True,
+            f"{EVA_ROOT_DIR}/evadb/udfs/GPT4ALL.py",
+        )
+        udf.execute()
+        # fetching Similarity based on features from SentenceTransformerFeatureExtractor udf and returning top 3 results
+        pdf_table_similarity = (
+            cursor.table("PDFss")
+            .order(
+                """Similarity(
```
jiashenC, line 75: I am a bit confused here. If we are already ranking the source similarity by using sentence transformer. Why do we need another index and embedding inside the GPT4All UDF?

evadb/udfs/GPT4ALL.py
```diff
@@ -0,0 +1,136 @@
...
+
+from evadb.catalog.catalog_type import NdArrayType
+from evadb.configuration.configuration_manager import ConfigurationManager
+from evadb.configuration.constants import EVA_ROOT_DIR
+from evadb.udfs.abstract.abstract_udf import AbstractUDF
+from evadb.udfs.decorators.decorators import forward, setup
+from evadb.udfs.decorators.io_descriptors.data_types import PandasDataframe
+
+
+class GPT4AllQaUDF(AbstractUDF):
```
jiashenC, line 54: Should we change the file name to GPT4AllQA.py as well? I think this is a bit different from the original GPT4All model aleady.

test/integration_tests/test_lanchain_llm.py
affan00733: we are not creating index two times for sentence transformer and in chat4gptall udf , we are just creating the index in chat4gptall all , as you can see in the udf and test it must be clear @jiashenC

evadb/udfs/GPT4ALL.py
affan00733: sure

affan00733: changes added

test/integration_tests/test_lanchain_llm.py
jiashenC: 1. My confusion is why you need this similarity search outside UDF? You do have an FAISS index inside UDF, which will be doing similarity search when you pass the question and source to it.
2. More of a question to @gaurav274, do you want to keep an index inside UDF here, which is transparent to the system.

affan00733: so , first we are doing similarity for filtering and pass only the top 3 or 5 results for answers to the gpt4allqa udf

PR #841, opened 2023-06-09
app: private gpt

gaurav274:


apps/privategpt/privateGPT.py
```diff
@@ -12,3 +12,46 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from gpt4all import GPT4All
```
xzdandy, line 15: gpt4all is not in the setup.py. We need to install it manually.

apps/privategpt/privateGPT.py
```diff
@@ -12,3 +12,46 @@
...
+            "content": f"Answer this question based on context: {question}",
+        },
+    ]
+    return llm.chat_completion(messages, verbose=False, streaming=False)
+
+
+## Take input of queries from user in a loop
+while True:
+    question = input("Enter your question (type 'exit' to stop): ")
+    if question == "exit" or "quit":
```
xzdandy, line 50: This is a typo. We always exit now.

apps/privategpt/privateGPT.py
```diff
@@ -12,3 +12,46 @@
...
+        break
+    answer = query(question)
+
+    print("\n\n> Question:")
+    print(query)
```
xzdandy, line 55: This print does not output useful information.

PR #852, opened 2023-06-10
feat: optional import of decord, fix for sentence feature extractor

jarulraj:


evadb/udfs/sentence_feature_extractor.py
```diff
@@ -32,6 +32,7 @@ def setup(self):
             "sentence-transformers/all-MiniLM-L6-v2"
         )
         self.model = AutoModel.from_pretrained("sentence-transformers/all-MiniLM-L6-v2")
+        self.device = None
```
jiashenC, line 35: Can you remind me why do we need this var?

evadb/udfs/sentence_feature_extractor.py
```diff
@@ -32,6 +32,7 @@ def setup(self):
             "sentence-transformers/all-MiniLM-L6-v2"
```
jiashenC, line 32: Do we converge on which UDF to keep this or the `sentence_transformer_feature_extractor`? I think they are the same model, but the other implementation looks cleaner than this.

evadb/udfs/sentence_feature_extractor.py
jarulraj: We need to set `self.model_device` -- not sure if `self.device` also needs to be set.

jarulraj: On a machine with only a CPU, we get this error -- `has no attribute` `model_device`

evadb/udfs/sentence_feature_extractor.py
jarulraj: Which one did you add?

jarulraj: Let's just call this `text_feature_extractor`?

jiashenC: Sure. I added this one, but i think other one is much simpler and also the device issue will disappear. I just found that this UDF’s implementation has a bug.

PR #860, opened 2023-06-12
Summary :  'progressive' filter while downloding streams in youtube_qa app.

Chitti-Ankith:
Since it is possible that some YouTube streams may not have audio codec present,
    added a filter on 'progressive' field to ensure that the first stream that we
    select and download gets transcribed into text without any 'Audio Not Found' errors.
    
    Also, added a filter on mp4 file extension to prevent unnecessary time in changing
    file format.
    
    Test Plan : Manual
    Reviewers : joy.alurraj, yulai.cui

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,8 +59,8 @@ workflows:
             mode: "COV"
         - Windows:
             name: "Windows | v3.10"
-        - MacOS:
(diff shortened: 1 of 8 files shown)


jarulraj:
Thanks for the PR, @Chitti-Ankith.

PR #868, opened 2023-06-15
Improve create index

gaurav274:


evadb/optimizer/statement_to_opr_converter.py
```diff
@@ -313,13 +319,85 @@ def visit_explain(self, statement: ExplainStatement):
...
+
+        # We need to also store the primary keys of the table in the index to later do
+        # the mapping. Typically, the vector indexes support setting an integer ID with
+        # the embedding. Unfortunately, we cannot support multi-column primary keys for
+        # cases like video table and document table.
+        # Hack: In such cases, we convert them into a unique ID using the following
+        # logic: We assume that the maximum number of files in the table is <=
+        # MAGIC_NUMBER and the number of frames or chunks for each video/document is <=
+        # MAGIC_NUMBER. Based on this assumption, we can safely say that
+        # `_row_id` * MAGIC_NUMBER + `chunk_id` for document table and
```
jiashenC, line 345: How do you determine the magic number? And when will be the chunking be done?

gaurav274: Right now, I set it to an arbitrarily large value. The idea is that as long as the number of files in the table is fewer than this number, our assumption will hold. 

Reg chunking: It is done when we read the document similar to frame decoding in videos.

jiashenC: I see. And for video, each video will get a unique ID and each frame will be assigned a different frame ID? Is this assumption correct?

gaurav274: We already have these IDs (`_row_id` for video and `id` for frame). Building on top of it. The assumption is these ids won't change across runs. `_row_id` is persisted, so no issue there. `id` is generated at runtime, and as long as the `reader` is deterministic across runs, we don't have a problem.

evadb/readers/pdf_reader.py
```diff
@@ -34,10 +34,12 @@ def _read(self) -> Iterator[Dict]:
         doc = fitz.open(self.file_url)
 
         # PAGE ID, PARAGRAPH ID, STRING
+        # Maintain a global paragraph number per PDF
+        global_paragraph_no = 0
```
jiashenC, line 38: I added this so that the calculated id for the create index method stays unique. Otherwise, it is still not unique because multiple paragraphs at different pages can have the same paragraph id. @gaurav274

jiashenC:
Close this for now. https://github.com/georgia-tech-db/evadb/pull/1073 is merged as a fix.

PR #869, opened 2023-06-15
Large scale medical pdfs benchmark

xzdandy:
```
python download_medical_pdfs.py
python ingest.py -d medical_pdfs
```

Nit: It will good to show the progress of a long running query, so the users are aware that we are making progress.


xzdandy:
- ingest 204 pdfs
- ray parallelism: 16
- GPU util: ~21%
```
real 6m10.840s
user 2m28.445s
sys 1m7.104s
```

PR #888, opened 2023-06-23
Support for multiple videos in youtube_qa app.

Chitti-Ankith:
Created an app which can parse multiple videos and answer questions across all of them. The questions to be asked need to be places in the questions.txt file. The video ids to download, transcribe and embed needs to be specified in the yt_video_ids.

Will work on making the interface more User Friendly in a separate commit. 

For Benchmarks and more details, please refer to https://docs.google.com/document/d/1CGPVwSazBaymxyFuij6uP0NkwrjWs8-7UGSRGNBDLcQ/edit


evadb/udfs/chatgpt.py
```diff
@@ -97,21 +97,26 @@ def completion_with_backoff(**kwargs):
         results = []
 
         for prompt, query in zip(prompts, queries):
-            if prompt != "None":
-                query = prompt + ": " + query
+            # if prompt != "None":
+            #     query = prompt + ": " + query
```
yulaicui, line 101: perhaps clean up clean code

evadb/udfs/chatgpt.py
```diff
@@ -97,21 +97,26 @@ def completion_with_backoff(**kwargs):
...
                 "temperature": self.temperature,
                 "messages": [
                     {
                         "role": "user",
-                        "content": query,
+                        "content": f"Context to answer the question : {query}",
+                    },
+                    {
+                        "role": "user",
+                        "content": f"Answer the question based on context : {prompt} Answer only from the context, don't generate responses on your own",
```
yulaicui, line 113: I believe the model generates answers conservatively with temperature = 0. It is good that the query is specific, but perhaps we don't want it to be too lengthy because it takes api tokens. Do we have proof that querying in this format gives better answers? Plus, is there a reason to not let the llm to only answer based on the context? A user may ask very broad questions.

evadb/udfs/chatgpt.py
```diff
@@ -97,21 +97,26 @@ def completion_with_backoff(**kwargs):
...
                     }
                 ],
             }
 
             response = completion_with_backoff(**params)
+            print("Response : ", response.choices[0].message.content)
```
yulaicui, line 119: is this a debug message or is this needed for the workflow? if not, we probably don't need it

apps/youtube_qa/multi_youtube_video_qa/README.md
```diff
@@ -0,0 +1,24 @@
+# YouTube Question Answering
```
yulaicui, line 1: It is strange to have an app within an app. A user would be confused. I suggest we have a separate app for this or that we make provide users with the option to add multiple videos in the same app. What do you think? @jarulraj

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
```diff
@@ -0,0 +1,324 @@
...
+        grouped_transcript = []
+        print("Create transcript csv")
+        if len(transcripts) > 0:
+            for transcript in transcripts:
+                group_transcript(transcript, grouped_transcript)
+
+            # print(grouped_transcript)
+            df = pd.DataFrame(grouped_transcript)
+            # df = pd.DataFrame(transcripts) 
+            # print(df)
```
yulaicui, line 211: perhaps eliminate dead code

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
```diff
@@ -0,0 +1,324 @@
...
+        st = time.time()
+        for question in questions:
+            print(question)
+            question_start = time.time()
+            
+            # instead of passing all the documents to the LLM, we first do a
+            # semantic search over the embeddings and get the most relevant rows.
+            cursor.query("DROP TABLE IF EXISTS EMBED_TEXT;").execute()
+            text_summarization_query = f"""
+                CREATE MATERIALIZED VIEW
```
yulaicui, line 294: CREATE MATERIALIZED VIEW is outdated. We are removing it soon. Perhaps use CREATE TABLE instead.

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
```diff
@@ -0,0 +1,324 @@
...
+                CREATE MATERIALIZED VIEW 
+                EMBED_TEXT(text) AS 
+                SELECT text FROM embedding_table
+                ORDER BY Similarity(embedding('{question}'), features) DESC
+                LIMIT 3; 
+                """
+            cursor.query(text_summarization_query).execute()
+            print(cursor.table('EMBED_TEXT').select("*").execute())
+
+            chatgpt_udf = """
```
yulaicui, line 303: we should use api instead of explicit query. for instance, cursor.table(EMBED_TEXT).select(f"ChatGPT('{question}', text)")

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
```diff
@@ -0,0 +1,324 @@
...
+
+        questions = open('questions', 'r')
+        st = time.time()
+        for question in questions:
+            print(question)
+            question_start = time.time()
+            
+            # instead of passing all the documents to the LLM, we first do a
+            # semantic search over the embeddings and get the most relevant rows.
+            cursor.query("DROP TABLE IF EXISTS EMBED_TEXT;").execute()
```
yulaicui, line 292: same as below, use drop table udf instead of explicit query

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
```diff
@@ -0,0 +1,324 @@
...
+        cursor.query(
+            f"""CREATE TABLE embedding_table AS
+            SELECT embedding(text), text FROM Transcript;"""
+        ).execute()
+        eft = time.time()
+        print(f"Creating embeddings took {eft - est} seconds")
+
+        # Create search index on extracted features.
+        cursor.query(
+            f"CREATE INDEX faiss_index ON embedding_table (features) USING FAISS;"
```
yulaicui, line 278: same as below, user create index api instead of explicit query.

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
```diff
@@ -0,0 +1,324 @@
...
+        
+        udf_query = """CREATE UDF IF NOT EXISTS embedding
+            IMPL  '../../../evadb/udfs/sentence_feature_extractor.py';
+            """
+        cursor.query("DROP UDF IF EXISTS embedding;").execute()
+        cursor.query(udf_query).execute()
+
+        cursor.query("DROP UDF IF EXISTS ChatGPT;").execute()
+        cursor.query(""" CREATE UDF IF NOT EXISTS ChatGPT
+          IMPL '../../../evadb/udfs/chatgpt.py';""").execute()
```
yulaicui, line 265: ChatGPT udf is now a builtin udf. There is no need to create it here.

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
```diff
@@ -0,0 +1,324 @@
+# coding=utf-8
```
yulaicui, line 1: This script is based on an old version of youtube_qa.py. The script is now updated with new logics as evadb has improved since then. Perhaps incorporate some changes from youtube_qa.py in the new master branch.

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
```diff
@@ -0,0 +1,324 @@
...
+        cursor.query(
+            """CREATE TABLE IF NOT EXISTS Transcript (text TEXT(100));"""
+        ).execute()
+        cursor.load("transcript.csv", "Transcript", "csv").execute()
+        print(f"Loading transcripts into DB took {time.time() - load_start_time} seconds")
+        
+        print("Creating embeddings and Vector Index")
+        
+        udf_query = """CREATE UDF IF NOT EXISTS embedding
+            IMPL  '../../../evadb/udfs/sentence_feature_extractor.py';
```
yulaicui, line 258: perhaps use create_udf api instead of the query. Plus, it is perhaps a better practice to declare this path as a macro  so that we can change them more easily. Check youtube_qa.py in master branch.

evadb/udfs/chatgpt.py
Chitti-Ankith: Done

evadb/udfs/chatgpt.py
Chitti-Ankith: Was put in for debugging, removed it. Thanks for pointing it out.

apps/youtube_qa/multi_youtube_video_qa/README.md
Chitti-Ankith: Moved it to a separate app folder for now. We can integrate it into the youtube_qa app in a separate commit.

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
Chitti-Ankith: done

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
Chitti-Ankith: In some runs, I had encountered the issue of ChatGPT UDF not existing while triggering the script, hence added this query. Will remove the drop_udf command and instead only create the UDF if it does not exist already.

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
Chitti-Ankith: Sure

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
Chitti-Ankith: Done

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
Chitti-Ankith: Had used API only. Missed removing this declaration.

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
Chitti-Ankith: Done

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
Chitti-Ankith: Done

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
Chitti-Ankith: the new logic of summarization / blog post generation may not be needed currently for this app. Have rebased the other changes. Please let me know if there is anything else to incorporate

evadb/udfs/chatgpt.py
Chitti-Ankith: While benchmarking, I was observing that sometimes the model was generating answers to the questions on its own and the generated answer doesnt match the info in the given video. Hence i add the restriction to answer based on the context. 

Removed it for now.

Chitti-Ankith:
I have added support for querying a specific youtube channel as well. @yulaicui @jarulraj please have a look.

apps/youtube_qa/multi_youtube_video_qa/multi_youtube_video_qa.py
yulaicui: init_builtin_udfs function is being run deterministically. It should not fail sometimes and succeed during other runs. There could be at least 2 reasons why it failed during some runs:
1. it is dropped somewhere in this app where it is not meant to be dropped
2. chatgpt_udf.py is modified. perhaps rerun `pip install .` when you modify the udf.

The code works. It's just that we need to present to users that we have those built-in functionalities and hide all the api-creation. When a user runs this app in a separate repo (which is what ultimately going to happen), they wouldn't want to come to evadb repo, copy-paste the chatgpt udf, and create the udf on their own.

jarulraj:
@Chitti-Ankith Thanks for putting together this app. Is there a way to test it on a subset of the urls -- like this test for the original youtube app -- `test_should_run_youtube_qa_app`?

What additional functionality are we adding in this app? Is this better stored in a separate repo (maybe in your account)?

jarulraj:
@Chitti-Ankith It looks like `scrapetube` is having some issues with the youtube URL. The app looks good to me. Let's keep it self-contained. Look at my changes, let's see if we can get rid of the custom ChatGPT UDF and instead make the built-in UDF take an optional argument ("like only use the context") in the query itself (instead of changing the UDF implementation file).

jarulraj:
Subsumed in #896

Chitti-Ankith:
Thanks for the feedback @jarulraj. I wasn't aware of UTs for the apps, I will add one in a separate commit.

With respect to new functionality introduced:

1) This app lets the user ask questions across any number of custom YT videos. The questions can be pre-aggregated or asked in a QA loop.

2) The app also lets the user to download any number of videos from a particular YT channel and ask their required queries.

I will add support for custom prompt as a separate argument in ChatGPT UDF. (This will also be useful for some of the additional app ideas I had shared with you) 

I also didnt find any issues with scrapetube and Channel URL. The following command works fine:

`scrapetube.get_channel(channel_url = "https://www.youtube.com/@LinusTechTips", limit = 10)`

The app currently only allows the user to specify the channel name and not the url, we can change that if needed.

PR #895, opened 2023-06-25
fix: Renaming Cursor Functions

aryan-rajoria:
- [x] Renaming Cursor Functions to better understand their uses.
- [x] Fixing create_vector_index() docstring


jarulraj:
@aryan-rajoria Please update the docs as well, that's very important.

jarulraj:
@aryan-rajoria We need to update the notebooks and test cases

PR #900, opened 2023-06-26
app: unit test for youtube_channel app, prompt option in ChatGPT

Chitti-Ankith:
Added a pytest for youtube_channel_qa app in tests/app_tests/ in test_youtube_channel_qa.py.

Added an optional prompt option in ChatGPT UDF so that apps can pass in custom requirements instead of modiying the chatgpt UDF directly.

Modifying the youtube_qa.py app on the local video to have a custom prompt as "Generate all responses as a pirate" created:

'Arrr, matey! The war in Ukraine has been raging for over 400 days, but it's not Russia or Ukraine that's winning - it's the US oil companies. These companies have made over $200 billion in profits from the instability caused by the war. Exxon Mobile alone made $56 billion in 2022. This is not the first time American oil companies have profited from war - they did the same thing in Iraq. The US officials sold a lie to go to war in Iraq, and while they didn't find any weapons of mass destruction, they did find lots of oil. Now, they're repeating the same story in Ukraine, feeding another war and filling the coffers of US oil companies. Arggg, it's a sad tale indeed.'


evadb/udfs/chatgpt.py
```diff
@@ -95,30 +96,45 @@ def completion_with_backoff(**kwargs):
             len(openai.api_key) != 0
         ), "Please set your OpenAI API key in evadb.yml file (third_party, open_api_key) or environment variable (OPENAI_KEY)"
 
-        prompts = text_df[text_df.columns[0]]
         queries = text_df[text_df.columns[0]]
```
jarulraj, line 99: @Chitti-Ankith Can you add comments with two examples of how this works for different types of inputs -- what would be `prompts` and `queries` etc.

For these two examples `ChatGPT(prompt, query)` and `ChatGPT(query)`. We could add this in the UDF doc string as well.

jarulraj:
Subsumed in #904

PR #908, opened 2023-06-27
feat: migrate to sqlalchemy 2.0

gaurav274:


jarulraj (approved):

jarulraj:
Is it a good idea to keep the old code and do a version check for `sqlalchemy`? Have most packages moved to 2.0?

gaurav274:
> Is it a good idea to keep the old code and do a version check for `sqlalchemy`? Have most packages moved to 2.0?

I tested with the older version of `sqlalchemy`. It works so we should be good.

Issue #786, opened 2023-05-26
Create materialized view with lateral join violates integrity constraints (column uniqueness)


yulaicui:
### Search before asking

- [X] I have searched the EVA DB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar bug report.

### Bug

Assume table UATRAC has been created by query `LOAD VIDEO "ua_detrac.mp4" INTO UATRAC;`

Consider the query below:
`CREATE MATERIALIZED VIEW IF NOT EXISTS uadtrac_fastRCNN AS SELECT * FROM UATRAC JOIN LATERAL Yolo(data) AS T(label, bbox, score) WHERE id < 5;`

EVA reports an integrity constraint violation error: 
Error: (sqlite3.IntegrityError) UNIQUE constraint failed: column_catalog.name, column_catalog.table_id
[SQL: INSERT INTO column_catalog (name, type, is_nullable, array_type, array_dimensions, table_id) VALUES (?, ?, ?, ?, ?, ?)]
[parameters: (‘_row_id’, ‘INTEGER’, 0, None, ‘()’, 5)]

This is because EVA automatically creates column "_row_id" to new tables, and the resulting view of this lateral join would have two columns with name "_row_id".

In general, EVA does not allow a CREATE MATERIALIZED VIEW if the select statement after AS has the form "SELECT *". Is there a reason. Should this be an issue worthy of fixes?

### Environment

- EVA DB v0.2.5
- OS: MacOS 12.3.1
- Python version: 3.10

### Minimal Reproducible Example

`LOAD VIDEO "ua_detrac.mp4" INTO UATRAC;`

`CREATE MATERIALIZED VIEW IF NOT EXISTS uadtrac_fastRCNN AS SELECT * FROM UATRAC JOIN LATERAL Yolo(data) AS T(label, bbox, score) WHERE id < 5;`

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

PR #912, opened 2023-06-29
feat: propose a simpler way to get software row_id

jiashenC:
The current approach to get the row_id in the base branch is a bit complicated. It requires us to modify or insert an expression to derive the row_id. It is easy for create index, but it becomes problematic for index scan. Index scan is enabled during optimization. We need to take care of the row_id expression during optimization by traversing multiple operators. I think this is not very clean. 

I am thinking we can just generate runtime time row id for storage besides structure table. The assumption is for those data table, they won't be used for operations like join, etc, so it is safe to just use a runtime row id which is different from the row_id actually stored on disk. 

Implementation-wise, I think it becomes simpler that we don't need manually insert any expression. Feedback is appreciated if any major issue is overlooked.

### DIFF
--- a/evadb/optimizer/statement_to_opr_converter.py
+++ b/evadb/optimizer/statement_to_opr_converter.py
@@ -14,10 +14,7 @@
 # limitations under the License.
 from evadb.catalog.catalog_type import TableType
 from evadb.catalog.catalog_utils import get_table_primary_columns
-from evadb.constants import MAGIC_NUMBER
 from evadb.expression.abstract_expression import AbstractExpression, ExpressionType
-from evadb.expression.arithmetic_expression import ArithmeticExpression
-from evadb.expression.constant_value_expression import ConstantValueExpression
 from evadb.expression.tuple_value_expression import TupleValueExpression
 from evadb.optimizer.operators import (
     LogicalCreate,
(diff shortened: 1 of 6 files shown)


xzdandy:
The following is confusing, if we do so, we shall give another name. 
> a runtime row id which is different from the row_id actually stored on disk.

I am in to redesign the row_id. The current implementation is complicated and requires special treatments in different components of the system.

jiashenC:
> The following is confusing, if we do so, we shall give another name.
> 
> > a runtime row id which is different from the row_id actually stored on disk.
> 
> I am in to redesign the row_id. The current implementation is complicated and requires special treatments in different components of the system.

One alternative is that we can rename it to "id"? For normal tables, "id" should be equal to "row_id". For others, it will be created at runtime. Any idea?

gaurav274:
> The current approach to get the row_id in the base branch is a bit complicated. It requires us to modify or insert an expression to derive the row_id. It is easy for create index, but it becomes problematic for index scan. Index scan is enabled during optimization. We need to take care of the row_id expression during optimization by traversing multiple operators. I think this is not very clean.
> 
> I am thinking we can just generate runtime time row id for storage besides structure table. The assumption is for those data table, they won't be used for operations like join, etc, so it is safe to just use a runtime row id which is different from the row_id actually stored on disk.
> 
> Implementation-wise, I think it becomes simpler that we don't need manually insert any expression. Feedback is appreciated if any major issue is overlooked.

I like the idea! However, this won't work for `delete,` which is problematic. How about we create a runtime unique id similar to what is done in the master, but we do it in the storage engine? This ensures that the rest of the system does not need to worry about it. We can make sure the `_row_id` (or some other column name) is always unique.

jiashenC:
> > The current approach to get the row_id in the base branch is a bit complicated. It requires us to modify or insert an expression to derive the row_id. It is easy for create index, but it becomes problematic for index scan. Index scan is enabled during optimization. We need to take care of the row_id expression during optimization by traversing multiple operators. I think this is not very clean.
> > I am thinking we can just generate runtime time row id for storage besides structure table. The assumption is for those data table, they won't be used for operations like join, etc, so it is safe to just use a runtime row id which is different from the row_id actually stored on disk.
> > Implementation-wise, I think it becomes simpler that we don't need manually insert any expression. Feedback is appreciated if any major issue is overlooked.
> 
> I like the idea! However, this won't work for `delete,` which is problematic. How about we create a runtime unique id similar to what is done in the master, but we do it in the storage engine? This ensures that the rest of the system does not need to worry about it. We can make sure the `_row_id` (or some other column name) is always unique.

Ok. A few clarifications.

1. Can you remind me why it would not work for delete?
2. `How about we create a runtime unique id similar to what is done in the master, but we do it in the storage engine?` Are you suggesting that we just create a new column with a different name for runtime unique id?

xzdandy:
If I understand correctly, ROWID needs to be permanent (https://www.ibm.com/docs/en/db2-for-zos/11?topic=types-row-id-values), while the ROWNUM can be generated at run time. 

I like the idea of storage engine. For structure data, we can get rowid from the sqlalchemy (or underlying database). So we only need handle the multimedia case to generate a unique id. 

PS: I suggest that `SELECT * does not return ROWID column` (https://www.ibm.com/docs/en/informix-servers/12.10?topic=statements-rowid-values-in-select), which makes the column implicit. It helps the case like` CREATE TABLE x AS SELECT * ...` (#786).

jiashenC:
Close this for now. https://github.com/georgia-tech-db/evadb/pull/1073 is merged as a fix.

PR #922, opened 2023-07-07
Title : Run EvaDB on Postgres

Chitti-Ankith:
Summary : Made changes to provide an option that would run EVA on top of Postgres server instead of SQLite. Created a CircleCI Job to validate the changes. Currently the job is only run for Python 3.10, we can add more configurations if needed. 

Also fixed the UTs that were failing on CircleCI runs in default master.

Test Plan : UT
Reviewers : @jarulraj @gaurav274


Chitti-Ankith:
Circle CI link : https://app.circleci.com/pipelines/github/Chitti-Ankith/eva/19

evadb/configuration/bootstrap_environment.py
```diff
@@ -46,7 +48,14 @@ def get_base_config(evadb_installation_dir: Path) -> Path:
...
 def get_default_db_uri(evadb_dir: Path):
-    return f"sqlite:///{evadb_dir.resolve()}/{DB_DEFAULT_NAME}"
+    import yaml
+
+    f = open(Path(EvaDB_INSTALLATION_DIR) / "evadb.yml", "r+")
+    config_obj = yaml.load(f, Loader=yaml.FullLoader)
+    if config_obj["experimental"]["use_postgres_backend"] is False:
+        return f"sqlite:///{evadb_dir.resolve()}/{DB_DEFAULT_NAME}"
+    else:
+        return f"postgresql://postgres:password@localhost:5432/{PG_DB_DEFAULT_NAME}"
```
gaurav274, line 58: Can we move this string to `catalog_database_uri` in `evadb.yml`?

evadb/configuration/constants.py
```diff
@@ -28,5 +28,6 @@
 CACHE_DIR = "cache"
 DATASET_DATAFRAME_NAME = "dataset"
 DB_DEFAULT_NAME = "evadb.db"
+PG_DB_DEFAULT_NAME = "circle_test"
```
gaurav274, line 31: If this is only for `circleci` testing, we can hardcode it.

evadb/database.py
```diff
@@ -39,7 +44,14 @@ def catalog(self) -> "CatalogManager":
...
 def get_default_db_uri(evadb_dir: Path):
-    return f"sqlite:///{evadb_dir.resolve()}/{DB_DEFAULT_NAME}"
+    import yaml
+
+    f = open(Path(EvaDB_INSTALLATION_DIR) / "evadb.yml", "r+")
+    config_obj = yaml.load(f, Loader=yaml.FullLoader)
+    if config_obj["experimental"]["use_postgres_backend"] is False:
+        return f"sqlite:///{evadb_dir.resolve()}/{DB_DEFAULT_NAME}"
+    else:
+        return f"postgresql://postgres:password@localhost:5432/{PG_DB_DEFAULT_NAME}"
```
gaurav274, line 54: Same as above, we should allow users to specify an existing Postgres db server.

setup.py
```diff
@@ -52,6 +52,7 @@ def read(path, encoding="utf-8"):
     "aenum>=2.2.0",
     "diskcache>=5.4.0",
     "retry>=0.9.2",
+    "psycopg2",
```
gaurav274, line 55: Let us not make it a default requirement.

evadb/configuration/bootstrap_environment.py
Chitti-Ankith: Added fields in the yml file to specify the username, password, host and port instead. This gives the clients a more flexible option to connect.

setup.py
Chitti-Ankith: Done

evadb/database.py
Chitti-Ankith: Done

evadb/configuration/constants.py
Chitti-Ankith: There is some issue in circleci wherein the postgres image pulled is not honoring the environment variable for db_name, and is instead getting defaulted to 'circle_test'. Hence I named it as such here. Should I still change it, for local workflows, the constant can be changed as desired.

Chitti-Ankith: There is already a bug open on this : https://github.com/CircleCI-Public/cimg-postgres/issues/94

evadb/catalog/models/utils.py
```diff
@@ -36,8 +36,8 @@ def init_db(engine: Engine):
     if not database_exists(engine.url):
         logger.info("Database does not exist, creating database.")
         create_database(engine.url)
-        logger.info("Creating tables")
-        BaseModel.metadata.create_all(bind=engine)
+    logger.info("Creating tables")
```
gaurav274, line 39: Just realized. We have function duplication here. We should be able to remove one instance.

evadb/configuration/bootstrap_environment.py
```diff
@@ -46,7 +48,27 @@ def get_base_config(evadb_installation_dir: Path) -> Path:
...
+        return f"sqlite:///{evadb_dir.resolve()}/{DB_DEFAULT_NAME}"
+    else:
+        # Custom Postgres server arguments need to be set in the evadb.yml file
+        pg_host = config_obj["postgres"]["pg_host"]
+        pg_port = config_obj["postgres"]["pg_port"]
+        pg_user = config_obj["postgres"]["pg_user"]
+        pg_password = config_obj["postgres"]["pg_password"]
+
+        if not pg_password:
+            return f"postgresql://{pg_user}@{pg_host}:{pg_port}/{PG_DB_DEFAULT_NAME}"
```
gaurav274, line 68: What do you think if we ask the user to provide this string directly instead of 4 parameters?

gaurav274: Right now, `catalog_database_uri` in `config.yml`  is populated with `sqlite` string. I suggest we use it for postgres. Users can change it to Postgres. 
Let me know if I'm missing some complexity with this approach. Thanks!

Chitti-Ankith: Main concern was if users set 'use_postgres_backend' to true without setting the catalog_database_uri field, they will get an authentication issue since we don't have a default pg uri. However your suggested approach is much simpler, I have made the changes and added comments to ensure the field is set.

evadb/evadb.yml
```diff
@@ -15,17 +15,15 @@ executor:
...
 server:
   host: "0.0.0.0"
   port: 8803
   socket_timeout: 60
 
 experimental:
-  ray: False 
+  ray: False
+  # Set the postgres db uri to connect to in 'catalog_database_uri' 
+  use_postgres_backend: False
```
gaurav274, line 26: Do we need this flag? We can infer it directly from the `catalog_database_uri`. We can avoid asking users for an extra thing.

gaurav274:
Overall LGTM! Fix the linter issue and address a minor comment. Thanks!

evadb/evadb.yml
Chitti-Ankith: Done, removed it.

Chitti-Ankith:
Thanks for the review @gaurav274 !

gaurav274 (approved):

PR #923, opened 2023-07-08
feat: support create table from select api

yulaicui:
This PR adds api support for create table from select.
for instance
`cursor.query("CREATE TABLE transcript AS SELECT SpeechRecognizer(audio) FROM video").execute()`
can be written as 
`select_rel = cursor.table("video").select("SpeechRecognizer(audio)")`
`cursor.create_table("transcript", query=select_rel).execute()`


jarulraj:
This was resolved in #1125

PR #924, opened 2023-07-11
fix tutorials 11-similarity-search-for-motif-mining.ipynb

xzdandy:
Currently, the tutorials failed due to the error `database is locked`. 

This is triggered by the commit in the [catalog.upsert_udf_cost_catalog_entry](https://github.com/georgia-tech-db/evadb/blob/d9a10de0d1c1ea6084ee88d1517c61a54f63d530/evadb/executor/executor_utils.py#L50) after the Yolo call.

We are able to fix the problem by removing the `batch_mem_size=12` at [image_storage_engine.py](https://github.com/georgia-tech-db/evadb/blob/d9a10de0d1c1ea6084ee88d1517c61a54f63d530/evadb/storage/image_storage_engine.py#L30). It is not clear why this fixes the `database is locked` issue. Need more investigation to locate the root cause. 

PS: it is not clear where the number 12 comes from.

### DIFF
--- a/evadb/storage/image_storage_engine.py
+++ b/evadb/storage/image_storage_engine.py
@@ -27,7 +27,7 @@ class ImageStorageEngine(AbstractMediaStorageEngine):
         super().__init__(db)
 
     def read(self, table: TableCatalogEntry) -> Iterator[Batch]:
-        for image_files in self._rdb_handler.read(self._get_metadata_table(table), 12):
+        for image_files in self._rdb_handler.read(self._get_metadata_table(table)):
             for _, (row_id, file_name) in image_files.iterrows():
                 system_file_name = self._xform_file_url_to_file_name(file_name)
                 image_file = Path(table.file_url) / system_file_name
--- a/tutorials/11-similarity-search-for-motif-mining.ipynb
+++ b/tutorials/11-similarity-search-for-motif-mining.ipynb
@@ -56,10 +56,10 @@
    "id": "aa5181dc",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2023-06-24T18:28:36.681648Z",
-     "iopub.status.busy": "2023-06-24T18:28:36.681200Z",
-     "iopub.status.idle": "2023-06-24T18:28:49.116256Z",
-     "shell.execute_reply": "2023-06-24T18:28:49.115289Z"
+     "iopub.execute_input": "2023-07-11T07:28:08.375931Z",
+     "iopub.status.busy": "2023-07-11T07:28:08.375426Z",
+     "iopub.status.idle": "2023-07-11T07:28:20.429793Z",
+     "shell.execute_reply": "2023-07-11T07:28:20.429076Z"
     }
    },
    "outputs": [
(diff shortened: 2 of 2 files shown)


xzdandy:
Please do not merge before locating the root cause.

gaurav274:
> credits to @xzdandy , we have identified the root cause. evadb hangs at commit because when a generator variable "result" is created in `read()` from `sqlite_storage_engine.py`, there is another db session that is trying to update UDF cost by writing to `udf_cost_catalog`. In short, there are two sessions running on the same time with one trying to write while another is still reading. I am able to recreate the scenario with the following snippet:
> 
> ```python
> from sqlalchemy import create_engine, event
> from sqlalchemy.orm import scoped_session, sessionmaker
> from sqlalchemy.sql import text
> from sqlalchemy import Column, Integer
> 
> def _enable_sqlite_pragma(dbapi_con, con_record):
>     dbapi_con.execute("pragma foreign_keys=ON")
>     dbapi_con.execute("pragma synchronous=NORMAL")
> 
> engine = create_engine('sqlite:////home/<user>/eva/evadb_data/evadb.db', connect_args={"timeout": 1000})
> event.listen(engine, "connect", _enable_sqlite_pragma)
> session = scoped_session(sessionmaker(bind=engine))
> table_catalog = session.execute(text('SELECT * FROM "<image table>";')) # read session is still running on generator
> print(table_catalog.__next__())
> 
> session1 = scoped_session(sessionmaker(bind=engine)) # create another session
> session1.execute(text('INSERT INTO udf_cost_catalog (udf_id, name, cost) VALUES (9, "Yolo", 1);'))
> session1.commit() # the script hangs here
> ```
> 
> This problem only happens when we set a low batch_mem_size (like 12) in image storage engine because the the read function doesn't escape the for loop on yield, leaving the session active with a running generator, whereas a large batch_mem_size yields all rows and escapes the for loop. This is a potential vulnerability in evadb's design that may impact scenarios with tables having many rows (say 100,000).
> 
> A hacky fix is to change `result = self._sql_session.execute(table_to_read.select())` from `sqlite_storage_engine.py` to `result = list(self._sql_session.execute(table_to_read.select()))`. However, this could compromise performance and is certainly not what generators are meant to be used.
> 
> **Fix pending...**
> 
> cc @jarulraj @gaurav274

This is happening because we disabled the `WAL` flag. Without it, it is not possible to support concurrent read/write. Can you try reenabling it?

xzdandy:
> credits to @xzdandy , we have identified the root cause. evadb hangs at commit because when a generator variable "result" is created in `read()` from `sqlite_storage_engine.py`, there is another db session that is trying to update UDF cost by writing to `udf_cost_catalog`. In short, there are two sessions running on the same time with one trying to write while another is still reading. I am able to recreate the scenario with the following snippet:
> 
> ```python
> from sqlalchemy import create_engine, event
> from sqlalchemy.orm import scoped_session, sessionmaker
> from sqlalchemy.sql import text
> from sqlalchemy import Column, Integer
> 
> def _enable_sqlite_pragma(dbapi_con, con_record):
>     dbapi_con.execute("pragma foreign_keys=ON")
>     dbapi_con.execute("pragma synchronous=NORMAL")
> 
> engine = create_engine('sqlite:////home/<user>/eva/evadb_data/evadb.db', connect_args={"timeout": 1000})
> event.listen(engine, "connect", _enable_sqlite_pragma)
> session = scoped_session(sessionmaker(bind=engine))
> table_catalog = session.execute(text('SELECT * FROM "<image table>";')) # read session is still running on generator
> print(table_catalog.__next__())
> 
> session1 = scoped_session(sessionmaker(bind=engine)) # create another session
> session1.execute(text('INSERT INTO udf_cost_catalog (udf_id, name, cost) VALUES (9, "Yolo", 1);'))
> session1.commit() # the script hangs here
> ```
> 
> This problem only happens when we set a low batch_mem_size (like 12) in image storage engine because the the read function doesn't escape the for loop on yield, leaving the session active with a running generator, whereas a large batch_mem_size yields all rows and escapes the for loop. This is a potential vulnerability in evadb's design that may impact scenarios with tables having many rows (say 100,000).
> 
> A hacky fix is to change `result = self._sql_session.execute(table_to_read.select())` from `sqlite_storage_engine.py` to `result = list(self._sql_session.execute(table_to_read.select()))`. However, this could compromise performance and is certainly not what generators are meant to be used.
> 
> **Fix pending...**
> 
> cc @jarulraj @gaurav274

@yulaicui, I recall when we debugged together, the simplified scenario is working with only `dbapi_con.execute("pragma synchronous=NORMAL")`. Is the `"pragma foreign_keys=ON"` the one causes the deadlock? Or have we missed some factors during that time? Thanks.

xzdandy:
Shall we merge @yulaicui ?

yulaicui (approved):

PR #928, opened 2023-08-11
feat: improve the use case doc (v1)

jiashenC:
* Image classification
* Image similarity search
* Q&A application


jarulraj (approved):
1. Any reason why `source/tutorials/08-chatgpt.ipynb` is removed from TOC?
2. Can we start using `functions` instead of `UDFs` everywhere to reduce confusion?

PR #932, opened 2023-08-15
feat: postgres integration - create database  

gaurav274:
1. Add support for creating database
```
CREATE DATABASE db
WITH ENGINE = "postgres",
PARAMETERS = {
    "user": "user",
    "password": "password",
    "host": "127.0.0.1",
    "port": "5432",
    "database": "demo"
};
```
2. Persist the info in the catalog
3. Todo: Verify if the provided info is valid - next PR


evadb/server/command_handler.py
```diff
@@ -42,9 +43,16 @@ def execute_query(
...
     with query_compile_time:
         stmt = Parser().parse(query)[0]
-        StatementBinder(StatementBinderContext(evadb.catalog)).bind(stmt)
-        l_plan = StatementToPlanConverter().visit(stmt)
-        p_plan = plan_generator.build(l_plan)
+
+        # For certain statements, we plan to omit binder and optimizer to keep the code
+        # clean. So, we handle such cases here and pass the statement directly to the
+        # executor.
+        if stmt not in SKIP_BINDER_AND_OPTIMIZER_STATEMENTS:
```
gaurav274, line 50: @xzdandy @jiashenC Added a skipping behavior in the system to avoid having to write boilerplate code for binder, optimizer.

xzdandy:
Looks good to me. To make it clear, the default engine (for now, sqlite3) in [evadb/catalog/sql_config.py](https://github.com/georgia-tech-db/evadb/blob/cae98dc78d56af5962b006a97deae5a7d29b3cb6/evadb/catalog/sql_config.py) will not have an entry in DatabaseCatalog (otherwise, we have a loop. Maybe we can actually, so it is not treated differently with any other databases). 
Some decisions we need to think are
- Is the default engine only storing EvaDB's metadata or it can also store user's data? In other words, can user use the EvaDB without connecting to any other structure databases?
-  Is the default engine always local or it can also connect to a remote database? Sharing model and external database connection with other users.

evadb/parser/create_statement.py
```diff
@@ -183,3 +183,42 @@ def __hash__(self) -> int:
                 self.query,
             )
         )
+
+
+class CreateDatabaseStatement(AbstractStatement):
```
xzdandy, line 188: Should the CreateDatabaseStatement inherit SKIP_BINDER_AND_OPTIMIZER_STATEMENTS?

xzdandy: Figured it out how it works.

xzdandy:
The `Test | v3.10 | Linux | Postgres` test actually failed. Need to take a look.

gaurav274:
> Looks good to me. To make it clear, the default engine (for now, sqlite3) in [evadb/catalog/sql_config.py](https://github.com/georgia-tech-db/evadb/blob/cae98dc78d56af5962b006a97deae5a7d29b3cb6/evadb/catalog/sql_config.py) will not have an entry in DatabaseCatalog (otherwise, we have a loop. Maybe we can actually, so it is not treated differently with any other databases created later by the user). Some decisions we need to think are
> 
> * Is the default engine only storing EvaDB's metadata or it can also store user's data? In other words, can user use the EvaDB without connecting to any other structure databases?
> * Is the default engine always local or it can also connect to a remote database? Sharing model and external database connection with other users.

Will document it in our design docs

xzdandy (approved):

PR #936, opened 2023-08-17
Use statement

jiashenC:
Working example

```sql
USE postgres {
    CREATE TABLE (...)
}
```


jiashenC:
@xzdandy @gaurav274 You can give read about the design. I will add more implementations to check in binder/catalog.

xzdandy:
For use query, we have the option to skip the binder and optimizer using [SKIP_BINDER_AND_OPTIMIZER_STATEMENTS](https://github.com/georgia-tech-db/evadb/blob/0f1bcf6b4313605dc5f2f6725bb78bfe5945faac/evadb/parser/utils.py#L29). This can simplify the implementation.

jiashenC:
> For use query, we have the option to skip the binder and optimizer using [SKIP_BINDER_AND_OPTIMIZER_STATEMENTS](https://github.com/georgia-tech-db/evadb/blob/0f1bcf6b4313605dc5f2f6725bb78bfe5945faac/evadb/parser/utils.py#L29). This can simplify the implementation.

I meant just checking if the data source is valid. Not gonna check anything for the native query.

evadb/parser/evadb.lark
```diff
@@ -172,11 +174,19 @@ explain_statement: EXPLAIN explainable_statement
 
 explainable_statement : select_statement | insert_statement | update_statement | delete_statement | create_table
 
+// Context Statements
+
+use_statement: USE database_name "{" query_string "}" // One shortcoming that query string cannot have parenthesis
```
xzdandy, line 179: Is this the limitation of the lark or ? Why we can not have parenthesis?

jiashenC: I will remove this comment. The native query can have parenthesis after I change to use curly bracket if that makes sense
```sql
// now
USE postgres {
   // some query
}

// before
USE postgres (
    // some query
)
```

If I use the parenthesis and the native query has arbitrary parentheses as well, I have some trouble of coming up with a grammar that works for all cases.

evadb/optimizer/operators.py
```diff
@@ -1239,3 +1240,37 @@ def __hash__(self) -> int:
                 self.search_query_expr,
             )
         )
+
+
+class LogicalUse(Operator):
```
gaurav274, line 1245: Can you also bypass optimizer for use? Just to simplify the code

evadb/parser/evadb.lark
gaurav274: Let me take care of it

evadb/optimizer/operators.py
gaurav274: Will push the code to take care of this

evadb/executor/sqlalchemy_executor.py
```diff
@@ -0,0 +1,54 @@
...
+
+        conn_str = generate_sqlalchemy_conn_str(
+            db_catalog_entry.engine,
+            db_catalog_entry.params,
+        )
+
+        engine = create_engine(conn_str)
+
+        with engine.connect() as con:
+            if "SELECT" in self._query_string or "select" in self._query_string:
```
gaurav274, line 50: We need to make this robust for cases where select is used for creating tables etc. Also, if query contains `Select` etc

xzdandy: What is the output for `CREATE TABLE` from sqlalchemy? Is it possible to simply execute the query and yield whatever output sqlalchemy yields in dataframe type?

gaurav274: I like this idea

gaurav274: How about we replicate this? https://github.com/mindsdb/mindsdb/blame/b417ecff78cca2e7f09d4e7daa3461595c77ad0d/mindsdb/integrations/handlers/postgres_handler/postgres_handler.py

I'm using it in the other PR for testing the connection.

jiashenC: https://github.com/sqlalchemy/sqlalchemy/issues/5433#issuecomment-652705643

Based on what I see, it is still true for this API for latest SQLAlchemy. There is no rows returned besides `SELECT`.

PR #945, opened 2023-08-23
fix: circle ci pip + setuptools conflict

jiashenC:


jiashenC:
@xzdandy @gaurav274 Current commit solves the previous error but a new error related to fitz lib occurs.

jiashenC:
I think this is in a good state. The cap is due to issues https://github.com/pymupdf/PyMuPDF/issues/2617 and https://github.com/pymupdf/PyMuPDF/issues/2616 in new release.

xzdandy (approved):
Thanks a lot!

PR #947, opened 2023-08-24
Integrating Slack into EvaDB

aryan-rajoria:
- [ ] Adding support for slack-sdk into EvaDB
- [ ] Fetch query - Using select statement
~- [ ] Delete query - Using delete statement~
~- [ ] Post query - Using insert statement~


kaushikravichandran:
As discussed, let's not go ahead with delete and post queries for now and keep that as future work.

PR #949, opened 2023-08-25
feat: improve circle ci

jiashenC:
Test folder reorg
```
test
|__ unit_tests
|__ integration_tests
     |__ short
     |__ long
```

For open PR, we will run unit tests and short integration tests. 

Once it is merged to staging branch, we will run long integration tests across different versions and with both ray disabled or enabled.

Once it passes, it will be merged to master, which will run full test (unit tests + all integration tests) with different python version and with both ray disabled and enabled. 

The testing for open PR is now around 8 mins, but majority of it is installing package. I plan to also add cache for that but it is not really working now.


xzdandy:
What is the staging branch we are going to use? Can we set up the repo so that is the default?

jiashenC:
> What is the staging branch we are going to use? Can we set up the repo so that is the default?

I think we can set up the rule once the new release is cut and when this is fully ready. Otherwise, it can impact other WIP PRs.

jiashenC:
Should be ready.

With pip wheel cache enabled, CI runs under 5mins for an open PR.

test/models/catalog/test_frame_info.py
xzdandy, line 1: Is this fine removed?

xzdandy (approved):

test/models/storage/test_batch.py
xzdandy, line 1: Is this test cases removed?

jiashenC: I see. Those test cases accidently match with some path in .gitignore so they are not added. I add them now.

PR #935, opened 2023-08-17
Ludwig-based model train and tune support.

xzdandy:
Provide support for the following queries. 

```
CREATE UDF IF NOT EXISTS PredictHouseRent FROM
(SELECT * FROM home_rentals)
TYPE Ludwig
'predict' 'rental_price';
```

```
SELECT PredictHouseRent(...) FROM home_rentals;
```

- [x] Basic implementation  
- [x] Documentation on the model training feature.
- [x] Set up circleci for ludwig tests

👋 Thanks for submitting a Pull Request to EvaDB!

🙌 We want to make contributing to EvaDB as easy and transparent as possible. Here are a few tips to get you started:

- 🔍 Search existing EvaDB [PRs](https://github.com/georgia-tech-db/eva/pulls) to see if a similar PR already exists.
- 🔗 Link this PR to a EvaDB [issue](https://github.com/georgia-tech-db/eva/issues) to help us understand what bug fix or feature is being implemented.
- 📈 Provide before and after profiling results to help us quantify the improvement your PR provides (if applicable).

👉 Please see our ✅ [Contributing Guide](https://evadb.readthedocs.io/en/stable/source/contribute/index.html) for more details.

PR #969, opened 2023-08-26
Forecasting in EVA

americast:
Implemented standalone forecasting in EVA (using [statsforecast](https://nixtla.github.io/statsforecast) package). You can run it via the following commands:

```sql
DROP TABLE IF EXISTS AirData;

CREATE TABLE AirData (
    unique_id TEXT(30),
    ds TEXT(30),
    y INTEGER);

LOAD CSV 'data/forecasting/air-passengers.csv' INTO AirData;

DROP UDF IF EXISTS Forecast;

CREATE UDF Forecast
FROM (SELECT unique_id, ds, y FROM AirData)
TYPE Forecasting
'predict' 'y';

SELECT Forecast(12) FROM AirData;
```
Here `Forecast(12)` signifies a horizon length of `12`.

Thanks!


xzdandy:
> Implemented standalone forecasting in EVA (using [statsforecast](https://nixtla.github.io/statsforecast) package). You can run it via the following commands:
> 
> ```sql
> DROP TABLE IF EXISTS AirData;
> 
> CREATE TABLE AirData (
>     unique_id TEXT(30),
>     ds TEXT(30),
>     y INTEGER);
> 
> LOAD CSV 'data/forecasting/air-passengers.csv' INTO AirData;
> 
> DROP UDF IF EXISTS Forecast;
> 
> CREATE UDF Forecast IMPL 'evadb/udfs/forecast.py';
> 
> SELECT Forecast(unique_id, ds, y) FROM AirData;
> ```
> 
> I plan to add more features to this. Tests and documentation are still pending.
> 
> Thanks!

Hi @americast, the design looks great. Is the idea that training will be implicit here? When the user runs `SELECT Forecast(unique_id, ds, y) FROM AirData;`, EvaDB will train and forecast underlying together. The rationale is that the trained forecast model can only apply to the same data source.

americast:
Thanks @xzdandy for your review. As of now, yes the training is implicit. Since `statsforecast` trains really fast, that should be fine. However, if we were to incorporate DL-based forecasting into this, we might want to train explicitly in the background.

jyotigoyal09:
Hi @americast, I see you are using Forecast(unique_id, ds, y) to forecast for a single time series data. I was wondering if you have a functionality to forecast for panel data. If yes, how would you make final model selection if groups in the panel do not have same seasonality/trend pattern.

gaurav274:
> Hi @americast, I see you are using Forecast(unique_id, ds, y) to forecast for a single time series data. I was wondering if you have a functionality to forecast for panel data. If yes, how would you make final model selection if groups in the panel do not have same seasonality/trend pattern.

Hi Jyoti, Thanks for showing interest in evadb. We are in the early phases of adding forecasting. @americast @xzdandy Thoughts?

americast:
@xzdandy I have pushed some commits and now the training is more like the Ludwig-style (#935). The training occurs when the UDF is created and the trained model is stored using a unique model file name generated from the data. Now, whenever the UDF is called, the user only needs to specify the horizon. I have updated the commands in https://github.com/georgia-tech-db/evadb/pull/969#issue-1868192153 accordingly.

xzdandy:
Thanks for the contribution! The implementation looks good to me at high level. 

We can add
- Add a long integration test cases to verify it worked end-to-end.
- Clean up the code (remove redundant print, redundant import)

americast:
> Thanks for the contribution! The implementation looks good to me at high level.
> 
> We can add
> 
> * Add a long integration test cases to verify it worked end-to-end.
> * Clean up the code (remove redundant print, redundant import)

@xzdandy Thanks. I have added the test and the docs.

xzdandy (changes requested):
Thanks for adding the tests and documentations. They look great.

Minor improvements. 

1. Fix the linter: `bash script/test/test.sh -m LINTER`
2. Add statsforecast dependency in setup.py
3. Install the statsforecast dependency in .cricle/config.yml, so the test will be run in long integration tests.

americast:
> Thanks for adding the tests and documentations. They look great.
> 
> Minor improvements.
> 
> 1. Fix the linter: `bash script/test/test.sh -m LINTER`
> 2. Add statsforecast dependency in setup.py
> 3. Install the statsforecast dependency in .cricle/config.yml, so the test will be run in long integration tests.

Thanks. I have updated them. Please let me know if this looks all good.

xzdandy (approved):
Great work!

americast:
> Great work! Verified the long integration test works on python3.10.

Awesome, thanks!

PR #972, opened 2023-08-27
feat: improve dev doc

jiashenC:


docs/source/contribute/index.rst
```diff
@@ -11,26 +11,22 @@ We welcome all kinds of contributions to EvaDB.
...
 
-   pip install .         # reinstall EvaDB package to include local changes 
-   pkill -9 evadb_server   # kill running EvaDB server (if any)
-   evadb_server&           # launch EvaDB server with newly installed package
+.. code-block:: bash
 
+   python3 -m venv test_evadb_venv
+   source test_evadb_venv/bin/activate
+   pip install --upgrade pip
+   pip install -e ".[dev]"
```
xzdandy, line 27: There are extra dependency besides [dev]. Shall we summarize them? Maybe we should have that information on the installation page and refer it here, since it will be helpful for users not only developers.

xzdandy (approved):

docs/source/contribute/index.rst
jiashenC: Sure. That is a good idea.

jiashenC: I think the installation page is no longer used. What would be a good place to show this information?

xzdandy: I think we should add one.

xzdandy (approved):

PR #976, opened 2023-08-28
feat: Slack chatbot

kaushikravichandran:


setup.py
```diff
@@ -54,7 +54,11 @@ def read(path, encoding="utf-8"):
     "retry>=0.9.2",
     "pydantic<2", # ray-project/ray#37019.
     "psutil",
-    "thefuzz"
+    "thefuzz",
+    "slack_sdk",
+    "slack-bolt",
+    "slack",
+    "slackclient"
```
xzdandy, line 61: Hi Kaushik, it will be better that slack dependency is specified in a requirements.txt under evadb/third_party/slack. So its dependency will only be installed when a connection to slack is created. You can check https://evadb.readthedocs.io/en/latest/source/contribute/new-data-source.html for more details.

kaushikravichandran: Fixed this. Will merge with Aryan's branch as earlier, instead of merging to staging. Closing this PR

PR #981, opened 2023-08-29
fix: use the same interface

jiashenC:
* Add some missing qdrant marker and check actual existence of qdrant.
* In-sync with interface.

### DIFF
--- a/evadb/server/command_handler.py
+++ b/evadb/server/command_handler.py
@@ -23,9 +23,35 @@ from evadb.models.storage.batch import Batch
 from evadb.optimizer.plan_generator import PlanGenerator
 from evadb.optimizer.statement_to_opr_converter import StatementToPlanConverter
 from evadb.parser.parser import Parser
+from evadb.parser.statement import AbstractStatement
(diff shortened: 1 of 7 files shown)


evadb/server/command_handler.py
```diff
@@ -23,11 +23,37 @@
...
+        p_plan = stmt
+    output = PlanExecutor(evadb, p_plan).execute_plan(
+        do_not_raise_exceptions, do_not_print_exceptions
+    )
+    if output:
+        batch_list = list(output)
+        return Batch.concat(batch_list, copy=False)
+
+
 def execute_query(
```
xzdandy, line 57: What is the difference between `execute_query`, `execute_query_fetch_all` and `execute_statement` now? It seems that they all do the same thing. I think originally `execute_query` is returning a generator, which intends to be used to implement fetch_all, fetch_one, though we never implemented fetch_one.

gaurav274 (approved):

evadb/server/command_handler.py
gaurav274: Surpassing this comment as this is breaking the build, and we need to hot fix it.

PR #1001, opened 2023-08-30
fix: pypi wheel does not work with third party

jiashenC:
* requirements.txt for dependencies is not packaged in the wheel.
* dependencies installation based on the absolute path. The relative path only works for dev testing, but errors when users actually install the wheel.


gaurav274:
@jiashenC Do we need this if we plan to avoid auto-installing packages?

jiashenC:
We can close this for now.

PR #1007, opened 2023-08-30
fix: create Table fails when projecting column from native table

gaurav274:
1. Database handler `get_columns` also returns the column datatypes
2. Throw Import errors instead of auto-installing missing dependencies.


evadb/third_party/databases/interface.py
```diff
@@ -26,11 +26,16 @@ def get_database_handler(engine: str, **kwargs):
...
-    # Dynamically install dependencies.
-    dynamic_install(engine)
-
     # Dynamically import the top module.
-    mod = dynamic_import(engine)
+    try:
+        mod = dynamic_import(engine)
+    except ImportError:
+        req_file = os.path.join(
+            "evadb", "third_party", "databases", engine, "requirements.txt"
```
xzdandy, line 34: Is this path accessible if user directly installs from pip instead of source?

evadb/third_party/databases/postgres/postgres_handler.py
```diff
@@ -108,3 +110,25 @@ def execute_native_query(self, query_string: str) -> DBHandlerResponse:
...
+            "smallint": int,
+            "numeric": float,
+            "real": float,
+            "double precision": float,
+            "character": str,
+            "character varying": str,
+            "text": str,
+            "boolean": bool,
+            # Add more mappings as needed
+        }
```
xzdandy, line 127: Why we do pg->python->evadb instead of pg->evadb?

gaurav274: Keep evadb types private from third-party integrations.

evadb/third_party/databases/interface.py
jiashenC: Yeah, I realize this path will only work for dev testing. We should switch using absolute path. In addition, we also need to package `requirements.txt`.

Reference to my closed PR: https://github.com/georgia-tech-db/evadb/pull/1001/files

xzdandy: Maybe just add a new property function to the handler that returns the list of package. No need to worry about path .

gaurav274: We won't be able to import the class to access the property function.

xzdandy (approved):

PR #1008, opened 2023-08-31
fix: CREATE TABLE persists entry in the catalog even if the query fails

gaurav274:
Rollback the catalog entry if the create query fails


xzdandy:
The PR also contains the change in #1007. We will finalize that first and then look into this.

xzdandy (approved):

PR #1019, opened 2023-08-31
feat: add CI for MySQL

jiashenC:


jiashenC:
I will move the CI to `staging` and `master` only once it is approved. For now, just make sure CI runs fine.

evadb/third_party/databases/mysql/mysql_handler.py
```diff
@@ -109,3 +110,23 @@ def execute_native_query(self, query_string: str) -> DBHandlerResponse:
...
+            "float": float,
+            "double": float,
+            # Add more mappings as needed
+        }
+
+        if mysql_type in mapping:
+            return mapping[mysql_type]
+        else:
+            raise Exception(
+                f"Unsupported column {mysql_type} encountered in the mysql table. Please raise a feature request!"
```
gaurav274, line 131: Can you also add the link to our Slack in this message (https://evadb.ai/community)?

test/third_party_tests/test_native_executor.py
```diff
@@ -186,10 +186,6 @@ def test_should_run_query_in_mysql(self):
         self._execute_native_query()
         self._execute_evadb_query()
 
-        # Test error.
-        self._raise_error_on_multiple_creation()
```
gaurav274, line 190: Why are we removing it?

evadb/third_party/databases/mysql/mysql_handler.py
gaurav274: Add `self.connection = None` in the base class. Else it throws property not found in some cases.

test/third_party_tests/test_native_executor.py
jiashenC: This test is actually invariant on which database engine we use. I currently just test on `Postgres`, so we should skip here.

evadb/third_party/databases/mysql/mysql_handler.py
jiashenC: What do you mean by the link?

xzdandy (approved):

PR #1047, opened 2023-09-05
fix: update udf to function in notebooks and docs

jiashenC:


jiashenC:
We can only run all of those successfully after the release.

xzdandy (approved):

PR #1058, opened 2023-09-06
created a new udf for multivariate

IshSiva:
time series forecasting

### DIFF
--- a/evadb/functions/multivariate_forecast.py
+++ b/evadb/functions/multivariate_forecast.py
@@ -1,0 +1,58 @@
+
+
+import pandas as pd
+
+from evadb.catalog.catalog_type import NdArrayType
+from evadb.functions.abstract.abstract_function import AbstractFunction
+from evadb.functions.decorators.decorators import forward, setup
+from evadb.functions.decorators.io_descriptors.data_types import PandasDataframe
+from statsforecast.adapters.prophet import AutoARIMAProphet
+
+class MultivariateForecast(AbstractFunction):
+    @property
+    def name(self) -> str:
+        return "MultivariateForecast"
+
+
+    @setup(cacheable=False, function_type="Forecasting", batchable=True)
+    def setup(self, train_end_date=None):
+        self.train_end_date = train_end_date
+        
+    
+    @forward(
+        input_signatures=[],
+        output_signatures=[
+            PandasDataframe(
+                columns=["ds", "y"],
+                column_types=[
+                    NdArrayType.STR,
+                    NdArrayType.FLOAT32,
+                ],
+                column_shapes=[(None,), (None,)],
+            )
+        ],
+    )
+    def forward(self, data) -> pd.DataFrame:
+        
+        train_df = data[data['ds']<=self.train_end_date]
+        test_df = data[data['ds']>self.train_end_date]
+        
+        #instantiate the model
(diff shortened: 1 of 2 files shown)


xzdandy:
Hi Ish, what is the difference between the forecasting in this PR and existing [forecasting](https://evadb.readthedocs.io/en/stable/source/reference/ai/model-forecasting.html) in EvaDB

IshSiva:
> Hi Ish, what is the difference between the forecasting in this PR and existing [forecasting](https://evadb.readthedocs.io/en/stable/source/reference/ai/model-forecasting.html) in EvaDB?

Hi Andy, this is for multivariate forecasting. The previous PR works for univariate forecasting only.

xzdandy:
> > Hi Ish, what is the difference between the forecasting in this PR and existing [forecasting](https://evadb.readthedocs.io/en/stable/source/reference/ai/model-forecasting.html) in EvaDB?
> 
> Hi Andy, this is for multivariate forecasting. The previous PR works for univariate forecasting only.

Thanks Ish! Does multivariate subsumed univariate? If so, we can direclty change the existing function instead of creating a new one.

IshSiva:
> > > Hi Ish, what is the difference between the forecasting in this PR and existing [forecasting](https://evadb.readthedocs.io/en/stable/source/reference/ai/model-forecasting.html) in EvaDB?
> > 
> > 
> > Hi Andy, this is for multivariate forecasting. The previous PR works for univariate forecasting only.
> 
> Thanks Ish! Does multivariate subsume univariate? If so, we can direclty change the existing function instead of creating a new one.

Hi Andy, Prophet works for univariate also. I think we can update to having a single function. 
`
CREATE FUNCTION MultivariateForecast IMPL 'evadb/functions/multivariate_forecast.py'
train_end_date '2005-02-07' 
type 'MULTIVARIATE'
model 'PROPHET'
horizon 365
frequency 'd'
;`

Will this syntax work? 
we can have default parameters for univariate (ARIMA) and multivariate (PROPHET) types. 

@americast can you also share your thoughts on this

americast:
Hi @IshSiva, thanks for the PR. I think for multivariate forecasting, the user would have to also provide multiple columns that need to be considered for forecasting. Thus, the user would have to provide the target column, as well as columns that would be considered as additional regressors.

You can check out statsforecast's implementation at [this link](https://nixtla.github.io/statsforecast/src/adapters.prophet.html#holiday-prophet). If you would like to directly use `prophet`, it's even simpler as they have an [`add_regressor` option.](https://rdrr.io/cran/prophet/man/add_regressor.html).

IshSiva:
> this link

Hi @americast , I have used this statsforecast implementation of Prophet for the multivariate forecasting. The target column can be given as an additional argument in the create udf function. So can we just update this new function to support univariate also? (because it is just an extension).

americast:
@IshSiva I think we can give statsforecast's `Prophet Holiday` a try. The only change that might be needed is support for more columns as user input while creating the UDF that would act as a reference for forecasting. The `model` option is also a parameter already implemented (I think I missed adding it to the docs), so the user can simply choose `AutoARIMAProphet` if they need multivariate forecasting. Or, we can create an alias for the same.

jarulraj:
Let's go with just `statsforecast` and `neuralforecast`.

PR #1082, opened 2023-09-08
feat: create index if exists

jiashenC:


xzdandy (approved):

xzdandy:
Maybe we can take this opportunity to update the documentation with `CREATE INDEX`. We can at least have a brief example in https://evadb.readthedocs.io/en/stable/source/reference/evaql/create.html for now. Later we can improve it.

Issue #1081, opened 2023-09-08
Improve the univariate statsforecast function in EvaDB


xzdandy:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar feature requests.

### Description

* [x] The univariate statsforecast function train and predicts on the exact same input relation, so there is no need for a separate training procedure. Currently `SELECT Forecast(12) FROM AirData;` does not make sense.
* [ ] The timeseries column is not properly handled. statsforecast has a required format for the timeseries column. https://nixtla.github.io/statsforecast/docs/getting-started/getting_started_short.html
* [x] The univariate statsforecast expects a fixed schema for the input dataframe. Renaming the column is not handled properly now. 
* [x] Update documentation with all available parameters. 

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

americast:
> The univariate statsforecast function train and predicts on the exact same input relation, so there is no need for a separate training procedure. Currently SELECT Forecast(12) FROM AirData; does not make sense.

I believe we can simply do `SELECT Forecast(12);`. The `FROM` part is a little redundant, but I am not sure if that is in line with SQL syntax.

@xzdandy I'll take care of `2`, you can assign it to me. Thanks!

xzdandy:
> > The univariate statsforecast function train and predicts on the exact same input relation, so there is no need for a separate training procedure. Currently SELECT Forecast(12) FROM AirData; does not make sense.
> 
> I believe we can simply do `SELECT Forecast(12);`. The `FROM` part is a little redundant, but I am not sure if that is in line with SQL syntax.
> 
> @xzdandy I'll take care of `2`, you can assign it to me. Thanks!

Thanks @americast! `SELECT Forecast(12)` the syntax is not supported. We can add that. I will handle  `3` first, which does not allow me to doing forecast on tables with customized column names.  

For time data type, it can be tricky. For example, I am using [House Property Sales Time Series](https://www.kaggle.com/datasets/htagholdings/property-sales?resource=download) data set, where the saledate column is `30/09/2007`, which is not the default panda data type format. We need to support some kind of data type and conversion here. Any idea you have.

americast:
>I will handle 3 first, which does not allow me to doing forecast on tables with customized column names.

@xzdandy I had added some support for customized column names in #969 . It's handled by the `id` and `time` variables. Are they not working for you?

[...]

PR #1087, opened 2023-09-10
Support `SELECT expr;` which does not require `FROM table` 

xzdandy:
Address the `SELECT Forecast(12) FROM AirData;` to `SELECT Forecast(12);` in #1081 

- [x] update parser, binder, optimizer, and executor to allow project without children. 
- [x] update forecasting test cases and documentation.
- [x] add unit test and short integration test for `SELECT expr;`.
- [x] add documentation that we support `SELECT expr;`.


xzdandy:
Discussed offline. We will send a dummy batch instead of checking `batch is None` everywhere.

xzdandy:
@gaurav274 @jiashenC Please review. We should be ready for a forecasting notebook after this PR.

jiashenC (approved):
Thanks, the PR looks good to me.

Just curious about the workflow of the forecasting model for my own knowledge.

* During training (i.e., `CREATE FUNCTION`), the forecasting model reads a table and memorizes all of the data from the table to its internal state (e.g., weights). 
* During inference, the forecasting model outputs the prediction from its internal state.

Is this the reason why the `SeqScan` is no longer needed during inference?

xzdandy:
> Thanks, the PR looks good to me.
> 
> Just curious about the workflow of the forecasting model for my own knowledge.
> 
> * During training (i.e., `CREATE FUNCTION`), the forecasting model reads a table and memorizes all of the data from the table to its internal state (e.g., weights).
> * During inference, the forecasting model outputs the prediction from its internal state.
> 
> Is this the reason why the `SeqScan` is no longer needed during inference?

Yes. It always works on the original dataset or table.

jiashenC:
> > Thanks, the PR looks good to me.
> > Just curious about the workflow of the forecasting model for my own knowledge.
> > 
> > * During training (i.e., `CREATE FUNCTION`), the forecasting model reads a table and memorizes all of the data from the table to its internal state (e.g., weights).
> > * During inference, the forecasting model outputs the prediction from its internal state.
> > 
> > Is this the reason why the `SeqScan` is no longer needed during inference?
> 
> Yes. It always works on the original dataset or table.

I see. Then, if the user wants the same model to work with a different dataset, the user needs to call `CREATE FUNCTION` again. Is that correct?

xzdandy:
> > > Thanks, the PR looks good to me.
> > > Just curious about the workflow of the forecasting model for my own knowledge.
> > > 
> > > * During training (i.e., `CREATE FUNCTION`), the forecasting model reads a table and memorizes all of the data from the table to its internal state (e.g., weights).
> > > * During inference, the forecasting model outputs the prediction from its internal state.
> > > 
> > > Is this the reason why the `SeqScan` is no longer needed during inference?
> > 
> > 
> > Yes. It always works on the original dataset or table.
> 
> I see. Then, if the user wants the same model to work with a different dataset, the user needs to call `CREATE FUNCTION` again. Is that correct?

That is correct. On the other hand, If I understand correctly, for forecasting, the same model can not work for other datesets. It is trained for a specific dataset. @americast who should have more knowledge in this space.

americast:
Thanks @xzdandy for the PR aned @jiashenC for the review. Right, a function would be linked to a trained model linked to a specific dataset. Thus, while calling the function, we would not need to pass any dataset information.

PR #1094, opened 2023-09-11
Fixes date and frequency issues in forecasting

### LINKED ISSUE #1081: Improve the univariate statsforecast function in EvaDB

americast:
Fixes #1081 pt 2.

### DIFF
--- a/docs/source/reference/ai/model-forecasting.rst
+++ b/docs/source/reference/ai/model-forecasting.rst
@@ -48,7 +48,7 @@ EvaDB's default forecast framework is `statsforecast <https://nixtla.github.io/s
    * - PREDICT (required) 
      - The name of the column we wish to forecast.
    * - TIME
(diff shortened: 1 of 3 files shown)


xzdandy (approved):
Thanks! Very clean solution. Could you also update https://evadb.readthedocs.io/en/latest/source/reference/ai/model-forecasting.html#forecast-parameters for the supported time format? Thanks!

PR #1135, opened 2023-09-16
Pinecone vector store

Chitti-Ankith:
Added support for Pinecone in EvaDB. Since pinecone requires an API key, created an evironment variable in evadb yml file. If not set, we check for os environment.


xzdandy:
Good work [Chitti-Ankith](https://github.com/Chitti-Ankith)! Thanks for the contribution. Is it possible to add some test cases?

Chitti-Ankith:
Any suggestions on how to add them since we need an API Key to init

evadb/utils/generic_utils.py
```diff
@@ -489,6 +489,16 @@ def try_to_import_qdrant_client():
         )
 
 
+def try_to_import_pinecone_client():
+    try:
+        import pinecone  # noqa: F401
+    except ImportError:
+        raise ValueError(
+            """Could not import pinecone_client python package.
+                Please install it with 'pip install pinecone_client`."""
```
xzdandy, line 498: nit typo pinecone-client

xzdandy:
> Any suggestions on how to add them since we need an API Key to init

Get it. We can mock the pinecone client, and have a unit test case in this case.

jarulraj:
@xzdandy Is there an example of such mocking that @Chitti-Ankith can use? 

Would this approach also work for testing third-party structured data sources like Clickhouse added by @Preethi1609?

xzdandy:
We can refer the unit testcases under https://github.com/georgia-tech-db/evadb/blob/staging/test/unit_tests/storage. Yes, it also works for Clickhouse. 

Meanwhile, if it is possible, we shall have some kind of integration test to verify it works end to end at staging branch. Does pinecone have any free-tier API keys?

jarulraj:
This wraps up `Qdrant`, `Pinecone`, `FAISS` and `pgvector`. That should be enough for our vector benchmark.

We could consider adding a subset of these systems if they are easy to add with this framework: 
`Weaviate`, `Milvus`, `Vespa`, and `Chroma`.

test/integration_tests/long/test_similarity.py
```diff
@@ -388,3 +392,39 @@ def test_end_to_end_index_scan_should_work_correctly_on_image_dataset_qdrant(sel
...
+
+        # We need to always drop the index as Pinecone's free tier only supports a single current index.
+        drop_index_query = "DROP INDEX IF EXISTS testpineconeindeximagedataset;"
+        execute_query_fetch_all(self.evadb, drop_index_query)
+        create_index_query = """CREATE INDEX testpineconeindeximagedataset
+                                ON testSimilarityImageDataset (DummyFeatureExtractor(data))
+                                USING PINECONE;"""
+        execute_query_fetch_all(self.evadb, create_index_query)
+        # Sleep to ensure the pinecone records get updated as Pinecone is eventually consistent.
+        time.sleep(20)
```
Chitti-Ankith, line 417: @xzdandy @jiashenC @gaurav274 @jarulraj since Pinecone is eventually consistent, we can't immediately query results after creating indexes. Hence i had to put a sleep in between here. Going through the documentation, I couldnt find a proper solution for this.

One suggestion was to look at the total_vector_count in describe_index_stats() fn https://docs.pinecone.io/docs/insert-data and wait till it gets updated but this won't work for updates. Any suggestions on this? Should we just assume that the clients will wait for the results to persiste before querying.

test/integration_tests/long/test_similarity.py
```diff
@@ -388,3 +392,39 @@ def test_end_to_end_index_scan_should_work_correctly_on_image_dataset_qdrant(sel
...
+
+    def test_end_to_end_index_scan_should_work_correctly_on_image_dataset_pinecone(
+        self,
+    ):
+        # Set the env variables.
+        original_pinecone_key = os.environ.get("PINECONE_API_KEY")
+        original_pinecone_env = os.environ.get("PINECONE_ENV")
+
+        os.environ["PINECONE_API_KEY"] = "657e4fae-7208-4555-b0f2-9847dfa5b818"
+        os.environ["PINECONE_ENV"] = "gcp-starter"
```
Chitti-Ankith, line 407: Crated this free tier API key just for EvaDB testing sake.

evadb/utils/generic_utils.py
Chitti-Ankith: Actually pinecone_client also installs the same package

Chitti-Ankith:
Also the tests are failing with the issue Andy mentioned here https://evadb.slack.com/archives/C047S6Z0V40/p1695092906362269?thread_ts=1695071176.493499&cid=C047S6Z0V40. We should probably get it fixed instead of having to revert the change and test everytimg

xzdandy:
> Also the tests are failing with the issue Andy mentioned here https://evadb.slack.com/archives/C047S6Z0V40/p1695092906362269?thread_ts=1695071176.493499&cid=C047S6Z0V40. We should probably get it fixed instead of having to revert the change and test everytimg

You can pull the latest staging. This has been already fixed by @gaurav274

test/integration_tests/long/test_similarity.py
xzdandy: You can use `self.evadb.config.update_value("third_party", "PINECONE_API_KEY", "657e4fae-7208-4555-b0f2-9847dfa5b818")` to update the value. You may also want to move this logic into cls setup and teardown. Otherwise, if the test failed, the environment variable will not be recovered.

test/integration_tests/long/test_similarity.py
xzdandy: I am not able to find a flush function either. For testcases, I think sleep is fine. One option is to print out a warning message when user is creating index using pinecone to indicate that the operation is eventual consistent.

test/integration_tests/long/test_similarity.py
Chitti-Ankith: Used environment vars itself in setup and teardown

xzdandy (approved):

PR #1136, opened 2023-09-16
feat: support order by using the projection columns

gaurav274:


evadb/binder/statement_binder.py
```diff
@@ -199,6 +200,9 @@ def _bind_select_statement(self, node: SelectStatement):
                 node.target_list = extend_star(self._binder_context)
             for expr in node.target_list:
                 self.bind(expr)
+                if isinstance(expr, FunctionExpression):
+                    add_func_expr_outputs_to_binder_context(expr, self._binder_context)
```
xzdandy, line 204: How about other expressions?

gaurav274: Only function expressions generate new columns. 
We don't support `AS` in the projection list. Will extend the support after doing it.

xzdandy (approved):

PR #1146, opened 2023-09-17
`CREATE OR REPLACE FUNCTION`

### LINKED ISSUE #1131: `CREATE OR REPLACE support`
### Search before asking

### Description

When training models, tuning the parameters are common. Adding a `CREATE OR REPLACE` to simplify the query, so users do not need to run a `DROP` query every time. 

### Use case

- CREATE OR REPLACE FUNCTION

xzdandy:
- [x] Support from parser to executor.
- [x] Fix existing unit tests and short integration tests. 
- [x] Add unit tests for `CREATE OR REPALCE FUNCTION` 
- [x] Update documentation
- [x] Add integration test
- [x] Check no long integration test failed due to this PR

Close #1131

### DIFF
--- a/docs/source/overview/model-inference.rst
+++ b/docs/source/overview/model-inference.rst
@@ -3,23 +3,25 @@
 Model Inference
 ===============
 
-In EvaDB, every model is a function. We can compose SQL queries using functions as building units similar to conventional SQL functions. EvaDB's `cascades optimizer <https://faculty.cc.gatech.edu/~jarulraj/courses/8803-s21/slides/22-cascades.pdf>` will optimize the evaluation of user-defined functions for lower latency. Go over :ref:`optimizations` for more details.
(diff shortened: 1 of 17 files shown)


jiashenC (commented):
What will be the bahavior of `CREATE OR REPLACE FUNCTION ... IF NOT EXISTS ...`?

It seems like Postgres (https://www.postgresql.org/docs/current/sql-createfunction.html) only allows `CREATE OR REPLACE` but not `IF NOT EXISTS` for `FUNCTION`.

xzdandy:
> What will be the bahavior of `CREATE OR REPLACE FUNCTION ... IF NOT EXISTS ...`?
> 
> It seems like Postgres (https://www.postgresql.org/docs/current/sql-createfunction.html) only allows `CREATE OR REPLACE` but not `IF NOT EXISTS` for `FUNCTION`.



> What will be the bahavior of `CREATE OR REPLACE FUNCTION ... IF NOT EXISTS ...`?
> 
> It seems like Postgres (https://www.postgresql.org/docs/current/sql-createfunction.html) only allows `CREATE OR REPLACE` but not `IF NOT EXISTS` for `FUNCTION`.

Yes, Postgres only supports ‘OR REPLACE’. I think it does not hurt support both, so did not remove the ‘IF NOT EXISTS’ support. When both are specified, an error message is raised.

jiashenC (approved):

PR #1153, opened 2023-09-17
feat: integrate with pgvector

jiashenC:
I will need some feedback on the design here. @xzdandy @gaurav274 

This PR reflects my initial design for in-database index features like pgvector. My initial idea is to offload as much as possible to the native database. When we create an index and do the index scan, we simply push an emulated query to the underlying database. While this is doable, this will introduce separate implementation paths in different components including the optimizer and the executor. 
 
Other than the current design, another option is to reuse the third-party vector integration interface. There are also a few details that I am not clear about
* For other vector libraries, we keep an index entry in our own catalog. Do we still maintain that even for pgvector? In the original implementation, the index catalog entry is linked to a table catalog entry. Because in this case, the table is inside Postgres, we also need to change the implementation here a little bit.
* ~~(Issue related to performance). Following the current create index implementation, data will be fetched from Postgres first but it is not really needed. The create index will anyway run inside Postgres.~~
* ~~The current vector index scan is implemented based on `_row_id`. When it is scanning data from Postgres, we need to figure out a way to populate `_row_id` for the native database engine.~~

### DIFF
--- a/evadb/binder/statement_binder.py
+++ b/evadb/binder/statement_binder.py
@@ -30,13 +30,7 @@ from evadb.binder.binder_utils import (
     resolve_alias_table_value_expression,
 )
 from evadb.binder.statement_binder_context import StatementBinderContext
-from evadb.catalog.catalog_type import (
-    ColumnType,
-    NdArrayType,
-    TableType,
-    VectorStoreType,
-    VideoColumnName,
-)
+from evadb.catalog.catalog_type import ColumnType, TableType, VideoColumnName
 from evadb.catalog.catalog_utils import get_metadata_properties, is_document_table
 from evadb.configuration.constants import EvaDB_INSTALLATION_DIR
 from evadb.expression.abstract_expression import AbstractExpression, ExpressionType
@@ -52,7 +46,6 @@ from evadb.parser.select_statement import SelectStatement
 from evadb.parser.statement import AbstractStatement
 from evadb.parser.table_ref import TableRef
 from evadb.parser.types import FunctionType
-from evadb.third_party.databases.interface import get_database_handler
 from evadb.third_party.huggingface.binder import assign_hf_function
 from evadb.utils.generic_utils import (
     load_function_class_from_file,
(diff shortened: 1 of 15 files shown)


xzdandy:
If the native database supports vector indexing, I think we should push down it to the native database system. This can be an optimizer rule. If the native database does not support vector indexing, we will do it ourselves.

gaurav274:
Why are we not pushing the `CREATE INDEX` query to `Postgres`?

gaurav274:
Based on offline discussion closing this PR

jiashenC:
I reopen this PR because the index scan pass is also implemented in this branch. 

This PR adds a feature to allow users to create an index on a table using `pgvector` and also do a similarity search using the existing `pgvector` index. 

Create index query.

```sql
CREATE INDEX test_index ON test_data_source.test_vector (embedding) 
    USING PGVECTOR
```
When users attempt to create an index in `pgvector`, it is internally translated to a native `Postgres` query to push down the create index query.

Index scan.

```sql
SELECT idx, embedding FROM test_data_source.test_vector 
    ORDER BY Similarity(DummyFeatureExtractor(Open(...)), embedding)
    LIMIT 1
```
I take some shortcuts in the optimizer that if the data source is from `Postgres`, translate the similarity query to the semantically equivalent similarity query that works for `Postgres` and push down the query to `Postgres`. That is implemented as part of the `VectorIndexScanExecutor`.

evadb/executor/create_index_executor.py
```diff
@@ -33,26 +35,41 @@ def __init__(self, db: EvaDBDatabase, node: CreateIndexPlan):
...
+        table = self.node.table_ref.table
+        db_catalog_entry = self.catalog().get_database_catalog_entry(
+            table.database_name
+        )
+
+        with get_database_handler(
+            db_catalog_entry.engine, **db_catalog_entry.params
+        ) as handler:
+            columns = table.table_obj.columns
+            # As other libraries, we default to HNSW and L2 distance.
```
jarulraj, line 61: What does other libraries mean?

jarulraj (approved):

evadb/binder/statement_binder.py
```diff
@@ -145,7 +147,34 @@ def _bind_create_index_statement(self, node: CreateIndexStatement):
         assert len(node.col_list) == 1, "Index cannot be created on more than 1 column"
 
         # TODO: create index currently only works on TableInfo, but will extend later.
-        assert node.table_ref.is_table_atom(), "Index can only be created on Tableinfo"
+        assert (
+            node.table_ref.is_table_atom()
+        ), "Index can only be created on an existing table"
+
+        # Vector type specific check.
+        catalog = self._catalog()
```
gaurav274, line 155: Reminder to move it outside

evadb/executor/create_index_executor.py
jiashenC: I meant other vector store types that we currently support (e.g., FAISS).

evadb/plan_nodes/abstract_plan.py
```diff
@@ -126,5 +126,5 @@ def find_all(self, plan_type: Any):
         """
 
         for node in self.bfs():
-            if isinstance(node, plan_type):
+            if isinstance(node, plan_type) or self.opr_type == plan_type:
```
xzdandy, line 129: What case we are trying to take care here?

test/third_party_tests/test_native_similarity_index.py
```diff
@@ -0,0 +1,108 @@
...
+        # Check results.
+        df = execute_query_fetch_all(self.evadb, query).frames
+        self.assertEqual(df["test_vector.idx"][0], 0)
+        self.assertNotIn("test_vector.dummy", df.columns)
+
+        # Clean up.
+        query = "USE test_data_source { DROP INDEX test_index }"
+        execute_query_fetch_all(self.evadb, query)
+        query = "USE test_data_source { DROP TABLE test_vector }"
+        execute_query_fetch_all(self.evadb, query)
```
xzdandy, line 108: Nits: Move the clean up out of the test into teardown, so if the test failed in the middle, we will still clean up the index and table.

evadb/optimizer/rules/rules.py
```diff
@@ -561,26 +571,34 @@ def _exists_predicate(opr):
...
-            catalog_manager().get_index_catalog_entry_by_column_and_function_signature(
+        # Only check the index existence when building on EvaDB data.
+        if not is_postgres_data_source:
+            # Get function_signature.
+            function_signature = (
+                None
+                if isinstance(base_func_expr, TupleValueExpression)
+                else base_func_expr.signature()
+            )
+
```
xzdandy, line 583: Does this mean that for pgvector, `CREATE INDEX test_index ON postgres.mytable (FUNC(data)) USING PGVECTOR` will not work?

evadb/executor/create_index_executor.py
```diff
@@ -33,26 +35,40 @@ def __init__(self, db: EvaDBDatabase, node: CreateIndexPlan):
...
+        ) as handler:
+            # As other libraries, we default to HNSW and L2 distance.
+            resp = handler.execute_native_query(
+                f"""CREATE INDEX {self.node.name} ON {table.table_name}
+                    USING hnsw ({self.node.col_list[0].name} vector_l2_ops)"""
+            )
+            if resp.error is not None:
+                raise ExecutorError(
+                    f"Native engine create index encounters error: {resp.error}"
+                )
```
xzdandy, line 68: Not relevant to this PR. Just come to my mind. We need to be consistent when we raise exception and when we return a panda dataframe that contains the message that the query failed.

xzdandy:
Overall design looks good to me. We may consider the gain/loss of moving some of the push down to optimizer. But I don't think that is the priority now. Left some comments for clarification.

evadb/plan_nodes/abstract_plan.py
jiashenC: No special case. I just feel the `plan_type` here can be the actual instance type or the internal `opr_type`, so I just added it to cover both cases. I can also remove it.

evadb/optimizer/rules/rules.py
jiashenC: The plan for this PR is to add a feature that index can be directly created on the raw table. I think the function expression use case is important, but doing it together makes this PR too complicated so will file a separate one later.

evadb/plan_nodes/abstract_plan.py
xzdandy: Could you give an example `isinstance(node, plan_type)` does not work? I am trying to understand what do you mean  by internal `opr_typle`?

evadb/executor/create_index_executor.py
jiashenC: I see. Then do we still raise error when encountering error when executing a query in the native engine? I will change it for the `VectorIndexScanExecutor` as well.

evadb/binder/statement_binder.py
jiashenC: @gaurav274 Added some changes. Can you give it a read? Not sure shall we still stick with the `singledispatchmethod` approach or do subclass inheritance and add some if statement like the `Executor`.

evadb/executor/create_index_executor.py
xzdandy: We can leave as it current it is. We need to redesign the error message system. For now, as long as the error message is clear, it is okay.

PR #1154, opened 2023-09-17
feat: add github actions to sync and release wheel

jiashenC:
Need to figure out a way to test it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,15 +1,26 @@
 name: Release
 
 on:
-  push:
-    branches:
-      - master
+  workflow_run:
+    workflows: [Sync Staging and Master]
+    types: 
+      - completed
+    branches: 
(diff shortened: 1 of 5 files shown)


gaurav274:
Can we skip the Pypi stage and test it on a fork?

jiashenC:
> Can we skip the Pypi stage and test it on a fork?

Verified the `Sync` workflow works that once the version file is updated, it will automatically sync `master` and `staging`.

The `Release` workflow will be triggered when `Sync` completes successfully. A `.pypirc` file is created but the last make wheel step is skipped for verification.

xzdandy:
Could we update the documentation on how does the release work? Thanks!

jiashenC:
> Could we update the documentation on how does the release work? Thanks!

Added docs.

xzdandy (approved):

Issue #1147, opened 2023-09-17
ORDER BY does not work on non projected columns


gaurav274:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar bug report.

### Bug

```
SELECT name FROM meme_images
ORDER BY Similarity(SiftFeatureExtractor(Open("{}")), SiftFeatureExtractor(data))
LIMIT 1;
```

The `data` column is accessible in the `order by` executor. 

Cause: Breaking PR - https://github.com/georgia-tech-db/evadb/pull/1136

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

PR #1155, opened 2023-09-17
feat: add support for enabling ORDER BY on non projected columns

gaurav274:
In response to PR #1147, we've introduced support for the use of ORDER BY on non-projected columns

```
SELECT name FROM meme_images
ORDER BY Similarity(SiftFeatureExtractor(Open("{}")), SiftFeatureExtractor(data))
LIMIT 1;

SELECT FaceDetector(data) FROM MyVideo WHERE id < 5 order by scores;

SELECT AirForecast(12) order by y;

Select name FROM meme_images ORDER BY name;
```

The following query won't work. 

`Select name AS n FROM meme_images ORDER BY n;`

Currently, we execute projection as the last operation. Only when a function expression is present it is executed before projection, allowing its use in ORDER BY. A complete fix will require us to redesign the binder. Right now, we've decided to defer this task because there is no urgency to support it.

Ref: https://github.com/duckdb/duckdb/blob/52a47a6b311a8aa008e7c78d8472eb52a937fa62/src/planner/expression_binder/order_binder.cpp#L68


xzdandy (approved):
Have you tried running long integration tests to verify the fix?

PR #1159, opened 2023-09-18
Load query exception for invalid file format 

affan00733:


jiashenC (commented):
Thanks, can you please add a test case to test this?

Otherwise, it looks good.

.gitignore
```diff
@@ -7,6 +7,8 @@ __pycache__/
 
 # C extensions
 *.so
+# virtutal env
+test_evadb_venv/
```
jiashenC, line 11: Also, maybe remove this

xzdandy:
Thanks for looking into this! To make sure my understanding is correct, this improves the error message but does not enable the case insensitive type comparison.

evadb/executor/load_executor.py
```diff
@@ -30,6 +31,9 @@ def exec(self, *args, **kwargs):
         """
 
         # invoke the appropriate executor
+        if self.node.file_options["file_format"] is None:
```
gaurav274, line 34: Please loop over all the possible enum values so that we don’t need to update this if we add more types later. Thanks!

jiashenC:
> Thanks for looking into this! To make sure my understanding is correct, this improves the error message but does not enable the case insensitive type comparison.

Yes. I feel just providing better error message is good enough. Shall we also make it case insensitive?

affan00733:
> Thanks, can you please add a test case to test this?
> 
> Otherwise, it looks good.

Yes i will add a testcase

.gitignore
affan00733: sure

affan00733:
> Thanks for looking into this! To make sure my understanding is correct, this improves the error message but does not enable the case insensitive type comparison.

Yes it just gives the correct error message

evadb/executor/load_executor.py
affan00733: @gaurav274 , in this if condition we are checking if we are receiving the file format none. It would be the case only if the file format is not their in lark parser.
And after this there is a condition already available where we are checking all the available file format and passing it to the load executor

test/integration_tests/long/test_load_executor.py
```diff
@@ -500,6 +501,15 @@ def test_load_pdfs(self):
         self.assertEqual(len(result.columns), 4)
         self.assertEqual(len(result), 26)
 
+    def test_load_query_incorrect_fileFormat(self):
+        try:
+            execute_query_fetch_all(
+                self.evadb,
+                f"""LOAD document '{EvaDB_ROOT_DIR}/data/documents/*.pdf' INTO pdfs;""",
+            )
+        except ExecutorError as e:
```
xzdandy, line 510: Nits: we can use `with self.assertRaise` to check the error message.
Example: https://github.com/georgia-tech-db/evadb/blob/ea6bfc66f5428bb261f6a6687521a842d66108f9/test/unit_tests/executor/test_create_udf_executor.py#L72

xzdandy (approved):

test/integration_tests/long/test_load_executor.py
affan00733: Sure, changed it as per the suggestion

PR #1162, opened 2023-09-19
Adding support for Sklearn linear regression in EvaDB

jineetd:
Starting this PR for integrating sklearn linear regression for EvaDb.

### DIFF
--- a/evadb/executor/create_function_executor.py
+++ b/evadb/executor/create_function_executor.py
@@ -141,11 +141,14 @@ class CreateFunctionExecutor(AbstractExecutor):
 
         arg_map = {arg.key: arg.value for arg in self.node.metadata}
         model = LinearRegression()
-        model.fit(X=aggregated_batch.frames, y=arg_map["predict"])
+        print(aggregated_batch.frames)
+        Y = aggregated_batch.frames[arg_map["predict"]]
+        aggregated_batch.frames.drop([arg_map["predict"]], axis=1, inplace=True)
+        model.fit(X=aggregated_batch.frames, y=Y)
         model_path = os.path.join(
             self.db.config.get_value("storage", "model_dir"), self.node.name
         )
-        pickle.dump(model, open(model_path, "rb"))
+        pickle.dump(model, open(model_path, "wb"))
         self.node.metadata.append(
             FunctionMetadataCatalogEntry("model_path", model_path)
         )
--- a/evadb/executor/create_function_executor.py
+++ b/evadb/executor/create_function_executor.py
@@ -141,7 +141,6 @@ class CreateFunctionExecutor(AbstractExecutor):
 
         arg_map = {arg.key: arg.value for arg in self.node.metadata}
         model = LinearRegression()
-        print(aggregated_batch.frames)
         Y = aggregated_batch.frames[arg_map["predict"]]
         aggregated_batch.frames.drop([arg_map["predict"]], axis=1, inplace=True)
         model.fit(X=aggregated_batch.frames, y=Y)
--- a/evadb/executor/create_function_executor.py
+++ b/evadb/executor/create_function_executor.py
@@ -40,6 +40,7 @@ from evadb.utils.generic_utils import (
     string_comparison_case_insensitive,
     try_to_import_forecast,
     try_to_import_ludwig,
+    try_to_import_sklearn,
     try_to_import_torch,
     try_to_import_ultralytics,
 )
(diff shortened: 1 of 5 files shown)


jineetd:
Currently the integration test is failing with the following error:
```
TypeError: Singleton array array('rental_price', dtype='<U12') cannot be considered a valid collection.
============================================================================================= short test summary info ==============================================================================================
FAILED test/integration_tests/long/test_model_train.py::ModelTrainTests::test_sklearn_regression - evadb.executor.executor_utils.ExecutorError: Singleton array array('rental_price', dtype='<U12') cannot be considered a valid collection.
```
Seems like something is wrong with return types somewhere. Need to debug further.

evadb/executor/create_function_executor.py
```diff
@@ -117,6 +118,48 @@ def handle_ludwig_function(self):
...
+        aggregated_batch_list = []
+        child = self.children[0]
+        for batch in child.exec():
+            aggregated_batch_list.append(batch)
+        aggregated_batch = Batch.concat(aggregated_batch_list, copy=False)
+        aggregated_batch.drop_column_alias()
+
+        arg_map = {arg.key: arg.value for arg in self.node.metadata}
+        model = LinearRegression()
+        model.fit(X=aggregated_batch.frames, y=arg_map["predict"])
```
xzdandy, line 144: @jineetd `arg_map["predict"]` returns the column name that user wants to predict, so in your test case, the value is `rental_price`. We can use `aggregated_batch.frames[arg_map["predict"]]` instead to get the values. Notice we also need to remove the `rental_price` column from `X`, you can achieve so by `aggregated_batch.frames.drop(arg_map["predict"], axis=1)` .

test/integration_tests/long/test_model_train.py
```diff
@@ -72,6 +72,23 @@ def test_ludwig_automl(self):
...
+            CREATE FUNCTION IF NOT EXISTS PredictHouseRent FROM
+            ( SELECT number_of_rooms, number_of_bathrooms, days_on_market, rental_price FROM HomeRentals )
+            TYPE Sklearn
+            PREDICT 'rental_price'
+            TIME_LIMIT 120;
+        """
+        execute_query_fetch_all(self.evadb, create_predict_function)
+
+        predict_query = """
+            SELECT PredictHouseRent(*) FROM HomeRentals LIMIT 10;
```
xzdandy, line 86: We need to pass in the columns that used is training instead of all of them.

evadb/functions/sklearn.py
```diff
@@ -0,0 +1,39 @@
...
+    def name(self) -> str:
+        return "GenericSklearnModel"
+
+    def setup(self, model_path: str, **kwargs):
+        try_to_import_sklearn()
+
+        self.model = pickle.load(open(model_path, "wb"))
+
+    def forward(self, frames: pd.DataFrame) -> pd.DataFrame:
+        predictions, _ = self.model.predict(frames, return_type=pd.DataFrame)
```
xzdandy, line 34: We need to proper column rename here. Check the forecast implementation for reference: https://github.com/georgia-tech-db/evadb/blob/58489e6a6637b2fc5046423e2a4c6a10b530ca27/evadb/functions/forecast.py#L55

jineetd:
Unit test is passing with these changes.

evadb/functions/sklearn.py
jineetd: Done

test/integration_tests/long/test_model_train.py
jineetd: Done

evadb/functions/sklearn.py
```diff
@@ -0,0 +1,42 @@
...
+    def setup(self, model_path: str, **kwargs):
+        try_to_import_sklearn()
+
+        self.model = pickle.load(open(model_path, "rb"))
+
+    def forward(self, frames: pd.DataFrame) -> pd.DataFrame:
+        predictions = self.model.predict(frames.iloc[:, :-1])
+        predict_df = pd.DataFrame(predictions)
+        # Assign the prediction column name same as that for data provided.
+        predict_df.rename(columns={0: frames.columns[-1]}, inplace=True)
```
xzdandy, line 37: Could we add some documentation here for clarity? For instance, the last column in frames is the predict column, and so we rename the column of the output dataframe to that.

test/integration_tests/long/test_model_train.py
```diff
@@ -72,6 +72,23 @@ def test_ludwig_automl(self):
         self.assertEqual(len(result.columns), 1)
         self.assertEqual(len(result), 10)
 
+    def test_sklearn_regression(self):
+        create_predict_function = """
+            CREATE FUNCTION IF NOT EXISTS PredictHouseRent FROM
+            ( SELECT number_of_rooms, number_of_bathrooms, days_on_market, rental_price FROM HomeRentals )
+            TYPE Sklearn
+            PREDICT 'rental_price'
+            TIME_LIMIT 120;
```
xzdandy, line 81: remove TIME_LIMIT. It is not used for Sklearn.

xzdandy (changes requested):
Overall looks great. Add some comments. We are good to merge then. Thanks for the contribution!

evadb/functions/sklearn.py
jineetd: Done

test/integration_tests/long/test_model_train.py
jineetd: Ohh okay, removed.

xzdandy (approved):

PR #1180, opened 2023-09-21
Adding SET statement for configuration management

hershd23:
Added

- Parser for SET statement
- Class for SET statement
- StatementType.SET to parser.types


hershd23:
Hi @jiashenC @xzdandy @gaurav274 adding you as reviewers. This PR is up for an early review

Have added parser, executor, plan and optimizer for SET queries please give them a look.

I see that the we send config with a category param as well. Wanted to make sure if we wanted to continue with that design, since in the example given duckDB did not have that
https://duckdb.org/docs/sql/configuration.html

Will take up SELECT command for displaying configs and tests for this in separate PRs

jiashenC (commented):
Thanks! The design looks good to me. Just some clarification questions

1. Are we going to universally use `string` type for `config value`? This can be easier if we want to store those values in the catalog.
2. Are we going to persist configuration values and those changes will be part of this PR?

hershd23:
For 1, I am not very sure since there is an example where we are adding tuples. Example - https://github.com/georgia-tech-db/evadb/blob/08b7e4cb3fedba7c189908a2051b281f7c2df809/benchmark/text_summarization/text_summarization_with_evadb.py#L28. We could serialize-deserialize but that would also require us saving the type.

For 2, I do not think we are. This is more about about creating the query to manage configurations. Persisting to disk can be taken as a natural extension of this problem once done.

evadb/executor/set_executor.py
```diff
@@ -0,0 +1,31 @@
...
+class SetExecutor(AbstractExecutor):
+    def __init__(self, db: EvaDBDatabase, node: SetPlan):
+        super().__init__(db, node)
+
+    def exec(self, *args, **kwargs):
+        # Get method implementation from the config.update_value
+        self._config.update_value(
+            category="default", key=SetPlan.config_name, value=SetPlan.config_value
+        )
+        pass
```
xzdandy, line 31: Nits: remove pass. Shall we return a message that indicates the configuration value has been changed?

evadb/parser/evadb.lark
```diff
@@ -79,6 +79,12 @@ drop_table: DROP TABLE if_exists? uid
     
 drop_function: DROP FUNCTION if_exists? uid
     
+// SET statements (configuration)
+set_statement: SET config_name EQUAL_SYMBOL config_value
```
xzdandy, line 83: If we want to be sync with Postgres, they support TO and =. Reference: https://www.postgresql.org/docs/7.0/sql-set.htm

xzdandy:
Great work. Looks good to me overall. Let's add a test cases to verify the config value has been changed.

evadb/executor/set_executor.py
hershd23: Ah yes pass was just a placeholder for the method. Will remove it

hershd23:
Hi @jiashenC, @xzdandy and @gaurav274. Was confused about one thing and wanted to run something by you.

In eva the current way we do configs is by three params

1. Category
2. config_key
3. config_value

This PR captures config_key and config_value but not category. For that do we want to change the method to include category or not?

If we change category then our query looks something like

```
SET EXECUTOR GPU_IDS = [0, 1]

OR

SET GPU_IDS = [0, 1] IN EXECUTOR
```

or something that involved passing the category as well.

OR

Since DuckDB and Postgres like @xzdandy added they just have key and value. No concept of category. That is the way query is currently implemented. This seems to be the common way DBs add configs.

```
SET GPU_IDS = [0, 1]
```
What do you say?

jiashenC:
> Hi @jiashenC, @xzdandy and @gaurav274. Was confused about this and wanted to run this by you.
> 
> In eva the current way we do configs is by three params
> 
>     1. Category
> 
>     2. config_key
> 
>     3. config_value
> 
> 
> This PR captures config_key and config_value but not category. For that do we want to change the method to include category or not?
> 
> If we change category then our query looks something like
> 
> ```
> SET EXECUTOR GPU_IDS = [0, 1]
> 
> OR
> 
> SET GPU_IDS = [0, 1] IN EXECUTOR
> ```
> 
> or something that involved passing the category as well.
> 
> OR
> 
> Since DuckDB and Postgres like @xzdandy added they just have key and value. No concept of category. That is the way query is currently implemented. This seems to be the common way DBs add configs.
> 
> ```
> SET GPU_IDS = [0, 1]
> ```
> 
> What do you say?

I vote for going with Postgress' approach.

jarulraj:
Let's ditch `category`

https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-SQL

evadb/parser/evadb.lark
hershd23: Done

hershd23:
@gaurav274 @xzdandy @jiashenC this is up for review

xzdandy (approved):
Great work.

hershd23:
Done @xzdandy

hershd23:
Since I am not using the plan classes now, I will remove them

Issue #1127, opened 2023-09-15
`parse_config_yml` opens file but does not close it


aryan-rajoria:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar bug report.

### Bug

ResourceWarning: unclosed file <_io.TextIOWrapper name='evadb.yml' mode='r+' encoding='UTF-8'>

![Screenshot 2023-09-15 at 2 16 30 PM](https://github.com/georgia-tech-db/evadb/assets/57455619/bc3eaa46-9d56-481f-a9f0-4db812ba89d3)

### Environment

- EvaDB: Latest
- Docker Linux

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

PR #1191, opened 2023-09-22
fix: unclosed file error

aryan-rajoria:
fixes warnings received when using superagi. fixes issue #1127


jiashenC:
Thanks, can you also share the bug caused by this? It can help us to track why we do this fix.

jiashenC (approved):

PR #1203, opened 2023-09-24
fix: chatgpt exact cache

jiashenC:


evadb/executor/create_function_executor.py
```diff
@@ -289,6 +289,7 @@ def exec(self, *args, **kwargs):
         if self.catalog().get_function_catalog_entry_by_name(self.node.name):
             if self.node.if_not_exists:
                 msg = f"Function {self.node.name} already exists, nothing added."
+                print(msg)
```
xzdandy, line 292: Nits: remove print

evadb/optimizer/optimizer_utils.py
```diff
@@ -283,8 +313,8 @@ def check_expr_validity_for_cache(expr: FunctionExpression):
     return (
         expr.name in CACHEABLE_FUNCTIONS
         and not expr.has_cache()
-        and len(expr.children) <= 1
-        and isinstance(expr.children[0], TupleValueExpression)
+        # and len(expr.children) <= 1
+        and isinstance(expr.children[1], TupleValueExpression)
```
xzdandy, line 317: Why we need the second parameter to the function to be `TupleValueExpression`? What if the function only accepts one arguments?

test/integration_tests/long/test_reuse.py
```diff
@@ -95,12 +109,43 @@ def _reuse_experiment(self, queries):
...
+        exec_times = []
+        batches = []
+        for i, query in enumerate(queries):
+            timer = Timer()
+            if i != 0:
+                with timer, patch.object(
+                    Batch, "apply_function_expression"
+                ) as mock_batch_func:
+                    mock_batch_func.side_effect = Exception("Results are not reused")
+                    batches.append(execute_query_fetch_all(self.evadb, query))
```
xzdandy, line 124: I like this test! Great idea.

evadb/optimizer/optimizer_utils.py
jiashenC: Fixed this.

evadb/optimizer/optimizer_utils.py
```diff
@@ -206,26 +241,21 @@ def optimize_cache_key(context: "OptimizerContext", expr: FunctionExpression):
...
-                    )
-                )
+        return optimize_cache_key_for_tuple_value_expression(context, keys[0])
+
+    # Handle ConstantValueExpressin + TupleValueExpression
+    if (
+        len(keys) == 2
+        and isinstance(keys[0], ConstantValueExpression)
+        and isinstance(keys[1], TupleValueExpression)
+    ):
```
xzdandy, line 254: Instead of hardcoding, can do register a mapping between the `optimize_cache_key_for ...` and `Expressions`. For example: 
```
optimize_cache_key_registry = {
    ConstantValueExpression: 'optimize_cache_key_for_constant_value_expression', 
    TupleValueExpression: 'optimize_cache_key_for_tuple_value_expression',
}
```
And we can use a for loop to iterate over all keys.

evadb/optimizer/optimizer_utils.py
```diff
@@ -280,12 +310,16 @@ def enable_cache_on_expression_tree(
...
-        and isinstance(expr.children[0], TupleValueExpression)
-    )
+    valid = expr.name in CACHEABLE_FUNCTIONS and not expr.has_cache()
+    if len(expr.children) == 1:
+        # Normal function that only takes one parameter.
+        valid &= isinstance(expr.children[0], TupleValueExpression)
+    elif len(expr.children) == 2:
+        # LLM-based function that takes two parameters.
+        valid &= isinstance(expr.children[0], ConstantValueExpression) and isinstance(
+            expr.children[1], TupleValueExpression
```
xzdandy, line 320: Is there a particular reason we need to enforce the number of keys and type here? Do `FUNC(constant, constant)` or `Func(tuple, tuple)` not work with cache now?

xzdandy (approved):
Overall looks good to me. There are conflict between staging to be fixed. 

Leave some comments about making the code more modularity. Please check whether they make sense. If so and we do not want to include in this PR, please create an issue for records.

evadb/optimizer/optimizer_utils.py
gaurav274: Even I wanted it. But I guess we can do it in a later PR.

jiashenC: Fixed this.

evadb/optimizer/optimizer_utils.py
jiashenC: I think we don't have functions that act like this for now. We can get rid of the check now, but probably need to create more test cases to make sure it works. Or we can get rid of it later once we do have concrete functions to use. Thoughts?

xzdandy: Sounds good.

PR #1209, opened 2023-09-24
feat: add testmon in pytest

jiashenC:
Add `testmon` in pytest

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,18 @@ workflows:
                 ignore: 
                   - master
                   - staging
-
+        ################################
+        #### LONG INTEGRATION TESTS: PR
+        ################################
+        ################################
+        - Linux:
+            name: Long Integration Test | v3.10 | Linux
+            mode: LONG INTEGRATION
+            filters:
+              branches:
+                ignore: 
+                  - master
+                  - staging
         ################################
         #### SHORT THIRDPARTY TESTS: PR
         ################################
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -213,14 +213,10 @@ jobs:
           keys:
             - v1-model_cache-{{ checksum "setup.py" }}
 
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -251,6 +267,20 @@ jobs:
...
+    # Collect the testmondata only on the staging branch 
+      - when:
+          condition:
+            and:
+              - equal: [ staging, << pipeline.git.branch >> ]
+              - equal: [ LONG INTEGRATION, << parameters.mode >> ]
+              - equal: [ "3.10", << parameters.v >> ]
+              - equal: [ DISABLED, << parameters.ray >> ] 
+          steps:
+          - save_cache:
```
xzdandy, line 279: Not clear why we don't need a indentation here. But circle ci complains when we have an indentation here.

jiashenC:
> Do we also need to push the `.testmondata` file?

It will be generated from the `staging` branch and used for open PRs.

.circleci/config.yml
```diff
@@ -201,6 +212,11 @@ jobs:
       - restore_cache:
           keys:
             - v1-model_cache-{{ checksum "setup.py" }}
+
+      # Restore testmondata
+      - restore_cache:
```
jiashenC, line 217: Not something critical, shall we only restore the cache for the open PR long integration cached test?

xzdandy: Actually I tried a different way to simplify the circle ci config. Please take a look.

xzdandy:
@jiashenC @gaurav274 This PR is ready. Please review. Thanks.

xzdandy (approved):

.circleci/config.yml
jiashenC: Oh, I meant it is currently restoring the cache for any test. Is that expected? Do we want to rerun everything for `staging` and `master`? Or we should run only affected as well?

PR #1211, opened 2023-09-25
fix: evadb_client fails on launch

gaurav274:


gaurav274:
> Just curious, why does this cause bug?

I think pprint does not support `flush`.

PR #1227, opened 2023-09-27
fix: issues with function input names not matching the column name

jiashenC:


jiashenC:
@xzdandy @gaurav274 @jarulraj  Need some thoughts on this. Currently, I added a check to verify the input column defined for the function is valid. Otherwise, it raises a `BinderError`.

The implementation works fine, but the problem here is that it will create a lot of break changes to our test suite and existing applications. Currently, there exist many functions input column definitions that do not match with the table. Functions like HuggingFace also have auto-populated input column definitions. It will probably take a long time to fix everything.

xzdandy:
> @xzdandy @gaurav274 @jarulraj Need some thoughts on this. Currently, I added a check to verify the input column defined for the function is valid. Otherwise, it raises a `BinderError`.
> 
> The implementation works fine, but the problem here is that it will create a lot of break changes to our test suite and existing applications. Currently, there exist many functions input column definitions that do not match with the table. Functions like HuggingFace also have auto-populated input column definitions. It will probably take a long time to fix everything.

I have a similar concern when looking into string `CONCAT`, the input arguments are variable length.

gaurav274:
Can you remind me of the need to verify the input column names?

jiashenC:
The motivation is from this issue (https://github.com/georgia-tech-db/evadb/issues/1206). The UDF is using some random input column that does not exist in the table. 

What do you mean by the order of inputs?

xzdandy:
Open discussion: Can we enforce the access of the parameters in functions by index (.e.g, 0, 1, 2, 3, 4) instead of the name? In other words, we only support `*args`, not `**kwargs`. Would that be a good idea or not?

jiashenC:
> Open discussion: Can we enforce the access of the parameters in functions by index (.e.g, 0, 1, 2, 3, 4) instead of the name? In other words, we only support `*args`, not `**kwargs`. Would that be a good idea or not?

So you are suggesting that we skip the name checking but just do typing checking?

xzdandy:
> > Open discussion: Can we enforce the access of the parameters in functions by index (.e.g, 0, 1, 2, 3, 4) instead of the name? In other words, we only support `*args`, not `**kwargs`. Would that be a good idea or not?
> 
> So you are suggesting that we skip the name checking but just do typing checking?

Yes.

jiashenC:
> > > Open discussion: Can we enforce the access of the parameters in functions by index (.e.g, 0, 1, 2, 3, 4) instead of the name? In other words, we only support `*args`, not `**kwargs`. Would that be a good idea or not?
> > 
> > 
> > So you are suggesting that we skip the name checking but just do typing checking?
> 
> Yes.

I see. 

1. That will not solve the original issue that created this PR right?
2. The order of columns in the table definition and in the function definition can have a mismatch. Then, do we strictly enforce that order to be the same?

xzdandy:
> 1. That will not solve the original issue that created this PR right?
> 2. The order of columns in the table definition and in the function definition can have a mismatch. Then, do we strictly enforce that order to be the same?

1. My idea is to access the args by index instead of name in the function. This by passes the problem, but I do not know whether there are cases that args have to be accessed by name
2.  The batch concat in function expression reserved the order, so I don't think that will be problem. For example, `SELECT FUNC(column_1, column_2) FROM Mycsv;`. The order of `column_1` and `column_2` is reserved.

gaurav274:
> > 1. That will not solve the original issue that created this PR right?
> > 2. The order of columns in the table definition and in the function definition can have a mismatch. Then, do we strictly enforce that order to be the same?
> 
> 1. My idea is to access the args by index instead of name in the function. This by passes the problem, but I do not know whether there are cases that args have to be accessed by name.
> 2. The batch concat in function expression reserved the order, so I don't think that will be problem. For example, `SELECT FUNC(column_1, column_2) FROM Mycsv;`. The order of `column_1` and `column_2` is reserved.

We always preserve the order of the columns in a function call.

PR #1233, opened 2023-09-28
GitHub Data Source Integration

xzdandy:
- [x] GitHub Data Source Integration
- [x] Batching support for native storage engine. We can not do batching in storage engine, which does not work with limit. Revert the change.
- [x] Full NamedUser table support
- [x] Enable circle ci local PR cache for testmondata
- [x] Native storage engine `read` refactory
- [x] Testcases
- [x] Github data source documentation


xzdandy:
@gaurav274 @jiashenC Please review the design.

The purpose of the integration:
1. The Github should be a data source instead of a user defined function, which improves the usability.
2. One step forward towards a single query for stargazer analysis and optimization. 
3. The first / the template for a non-sql data source integration. 

PS: testmon finally worked.

jiashenC (commented):
The overall design looks good to me. My question is this is more like stargazer specific handler right? If we later want to add other features of Github APIs, what is the plan to add those?

xzdandy:
> The overall design looks good to me. My question is this is more like stargazer specific handler right? If we later want to add other features of Github APIs, what is the plan to add those?

We will add another table mapping in the `supported_map`.

.circleci/config.yml
```diff
@@ -271,16 +272,21 @@ jobs:
...
+            or:
+              - equal: [ LONG INTEGRATION CACHE, << parameters.mode >> ]
+              - and:
+                - equal: [ LONG INTEGRATION, << parameters.mode >> ]
+                - equal: [ staging, << pipeline.git.branch >> ]
+                - equal: [ "3.10", << parameters.v >> ]
+                - equal: [ DISABLED, << parameters.ray >>]
           steps:
           - save_cache:
               key: v1-testmon_cache-{{ .Branch }}-python<< parameters.v >>-ray<< parameters.ray >>-{{ checksum "setup.py" }}-{{ epoch }}
```
jiashenC, line 284: I might have missed this in the previous PR. Where does the epoch parameter come from?

.circleci/config.yml
```diff
@@ -213,10 +213,11 @@ jobs:
           keys:
             - v1-model_cache-{{ checksum "setup.py" }}
 
-      # Always restore testmondata from staging, python3.10, ray disabled.
+      # First try restoring the testmondata from PR, than staging.
       - restore_cache:
           keys:
-            - v1-testmon_cache-staging-python3.10-rayDISABLED-{{ checksum "setup.py" }}
+            - v1-testmon_cache-{{ .Branch }}-python<< parameters.v >>-ray<< parameters.ray >>-{{ checksum "setup.py" }}-
```
jiashenC, line 219: What is the reason for the extra `-` in the end?

.circleci/config.yml
xzdandy: circle ci will skip cache if the key is the same. In our case, we want the testmondata to be updated every time, so we put the epoch to make the cache key different across runs.

jarulraj:
@kaushikravichandran Can you help add a docs PR for the app integration handler steps? thanks!

evadb/storage/native_storage_engine.py
```diff
@@ -158,46 +158,41 @@ def write(self, table: TableCatalogEntry, rows: Batch):
...
+                    index_dict = {
+                        element.lower(): index for index, element in enumerate(cols)
+                    }
+                    try:
+                        ordered_columns = sorted(
+                            table.columns, key=lambda x: index_dict[x.name.lower()]
+                        )
+                    except KeyError as e:
+                        raise Exception(f"Column mismatch with error {e}")
+                    result = [
```
gaurav274, line 189: Given we are moving to a generator, how about handling it here as well?

`result = (_deserialize_sql_row(row, ordered_columns) for row in result)` should keep `result` a generator.

evadb/third_party/databases/github/github_handler.py
```diff
@@ -0,0 +1,147 @@
...
+
+from evadb.third_party.databases.github.table_column_info import STARGAZERS_COLUMNS
+from evadb.third_party.databases.types import (
+    DBHandler,
+    DBHandlerResponse,
+    DBHandlerStatus,
+)
+
+
+class GithubHandler(DBHandler):
```
gaurav274, line 26: @kaushikravichandran @xzdandy Do we want to merge #1033 PR first?

evadb/third_party/databases/types.py
```diff
@@ -136,3 +154,31 @@ def execute_native_query(self, query_string: str) -> DBHandlerResponse:
...
+            uri = self.get_sqlalchmey_uri()
+            # Create a metadata object
+            engine = create_engine(uri)
+            metadata = MetaData()
+
+            Session = sessionmaker(bind=engine)
+            session = Session()
+            # Retrieve the SQLAlchemy table object for the existing table
+            table_to_read = Table(table_name, metadata, autoload_with=engine)
+            result = session.execute(table_to_read.select()).fetchall()
```
gaurav274, line 179: Minor: Do we need to return a generator?

evadb/storage/native_storage_engine.py
xzdandy: I think list comprehension is also a generator.

evadb/third_party/databases/types.py
xzdandy: generator does not work the sqlalmechy session now . Check #924 for the past investigation we have done.

evadb/storage/native_storage_engine.py
gaurav274: Reminder to fix it.

evadb/third_party/databases/types.py
gaurav274: Drop a comment in the code!

evadb/third_party/databases/github/github_handler.py
xzdandy: We'd like to introduce `SELECT`, `WRITE`, `CREATE`, `DROP` for the non-sql data source integration. And use them in the native_storage_engine.py. Some efforts are needed to refactor the pull request. To unblock the efforts for optimization in a short time period, I choose to do an implementation for Github data source first.  This PR includes the `SELECT` part, which can be a reference for the rest interfaces.

gaurav274 (approved):

PR #1234, opened 2023-09-28
Added basic functionalities of REST apis

imnidhi:


apps/rest/run_evadb.py
```diff
@@ -0,0 +1,86 @@
...
+            PARAMETERS = {
+                "user": "eva",
+                "password": "password",
+                "host": "localhost",
+                "port": "5000",
+                "database":""" + db_name + """
+            }; """
+        
+        res = cursor.query(query).df()
+        return {"msg" : "Created Database" , "response" : res.to_json()}
```
xzdandy, line 25: We do not create database ourselves. Users will use `CREATE DATABASE` to create the connection as needed.

apps/rest/run_evadb.py
```diff
@@ -0,0 +1,86 @@
...
+    return {"response" : res.to_json()}     
+
+@app.route('/table_data/<name>',methods = ['GET'])
+def get_data(name):
+    query = "Select * from" + name + ";"
+    res = cursor.query(query).df() #Show 
+    return {"response" : res.to_json()}
+
+@app.route('/query/<query>',methods = ['GET'])
+def query_from_db(query):
```
xzdandy, line 81: I think all we need is the query. The rest is not necessary. However, we need to optimize the query endpoint. Some I can think of now:
1. we may need a queue to serialize concurrent queries. @jiashenC has done some setup similar for slackbot. You can refer to that. 
2. We may also need the response to be streaming, considering we have a very large table as the result. 

Other than query, we can also have a upload endpoint for users to upload their csvs, pdfs, videos.

imnidhi: Hey Andy. I added the upload endpoint. Could you take a look in my latest commit?

xzdandy:
Hi @imnidhi, looks great! Could you fix the linter and add some testcases if possible? Thanks!

apps/rest/run_evadb.py
```diff
@@ -0,0 +1,57 @@
...
+UPLOAD_FOLDER = cwd + "/files"
+ALLOWED_EXTENSIONS = {'txt', 'pdf', 'png', 'jpg', 'jpeg', 'gif', 'mp3'}
+
+app = Flask(__name__)
+app.config['UPLOAD_FOLDER'] = UPLOAD_FOLDER
+api = Api(app)
+app.secret_key = "evadb-rest-apis"
+
+# Get Query response
+@app.route('/query/<query>',methods = ['GET'])
```
xzdandy, line 19: SQL queries usually have space. I am wondering how this will work. Does the user need to escape the url?

imnidhi: No. I tried with spaces and it automatically converts it to spaces so it should be fine

imnidhi: Also I added an input box similar to upload. It's easier to enter text that way

xzdandy (approved):

PR #1240, opened 2023-09-30
Add stable diffusion integration

xzdandy:
Reopen the #1111.


xzdandy:
Quick notice:

1. Fix testcases
2. Please change the notebook name and if notebook also does not work on circle ci, we can skip the it. You can do it in script/test/test.sh. Do let us know why it does not work.
3. Add a documentation to the function under https://evadb.readthedocs.io/en/stable/source/reference/ai/index.html
4. Both test/integration_tests/long/functions/test_dalle.py and test/integration_tests/long/functions/test_selfdiffusion.py look unit tests instead of long integration tests, which is fine. If so, move them to test/unit_tests.

evadb/functions/function_bootstrap_queries.py
```diff
@@ -202,6 +202,18 @@
             MODEL 'yolov8n.pt';
         """
 
+stablediffusion_function_query = """CREATE FUNCTION IF NOT EXISTS StableDiffusion
+        IMPL '{}/functions/stable_diffusion.py';
+        """.format(
+    EvaDB_INSTALLATION_DIR
+)
+
+dalle_function_query = """CREATE FUNCTION IF NOT EXISTS StableDiffusion
```
xzdandy, line 211: These two parameters are not used.

evadb/functions/stable_diffusion.py
```diff
@@ -0,0 +1,88 @@
...
+import pandas as pd
+
+from evadb.catalog.catalog_type import NdArrayType
+from evadb.functions.abstract.abstract_function import AbstractFunction
+from evadb.functions.decorators.decorators import forward
+from evadb.functions.decorators.io_descriptors.data_types import PandasDataframe
+from evadb.utils.generic_utils import try_to_import_replicate
+
+_VALID_STABLE_DIFFUSION_MODEL = [
+    "sdxl:af1a68a271597604546c09c64aabcd7782c114a63539a4a8d14d1eeda5630c33",
```
xzdandy, line 27: Could you add some documentation / comment on this model?

evadb/functions/stable_diffusion.py
```diff
@@ -0,0 +1,88 @@
...
+                    NdArrayType.STR,
+                ],
+                column_shapes=[(None,)],
+            )
+        ],
+        output_signatures=[
+            PandasDataframe(
+                columns=["response"],
+                column_types=[
+                    NdArrayType.STR,
```
xzdandy, line 59: The response is a image. So the type of is not `NdArrayType.STR`.

evadb/functions/stable_diffusion.py
```diff
@@ -0,0 +1,88 @@
...
+            )
+            os.environ["REPLICATE_API_TOKEN"] = replicate_api_key
+
+        # @retry(tries=5, delay=20)
+        def generate_image(text_df: PandasDataframe):
+            results = []
+            queries = text_df[text_df.columns[0]]
+            for query in queries:
+                output = replicate.run(
+                    "stability-ai/" + self.model, input={"prompt": query}
```
xzdandy, line 81: Where is this directory?

test/integration_tests/long/functions/test_dalle.py
```diff
@@ -0,0 +1,60 @@
...
+        for prompt in test_prompts:
+            insert_query = f"""INSERT INTO ImageGen (prompt) VALUES ('{prompt}')"""
+            execute_query_fetch_all(self.evadb, insert_query)
+
+    def tearDown(self) -> None:
+        execute_query_fetch_all(self.evadb, "DROP TABLE IF EXISTS ImageGen;")
+
+    @patch.dict("os.environ", {"OPENAI_KEY": "mocked_openai_key"})
+    @patch("openai.Image.create", return_value={"data": [{"url": "mocked_url"}]})
+    def test_dalle_image_generation(self, mock_openai_create):
```
xzdandy, line 44: This is a good unit tests we can move to test/unittest. 

On the other hand, you can add an `@open_skip_marker`, so the test will only be run when a openai key is provided and dependency is installed.

test/integration_tests/long/functions/test_selfdiffusion.py
```diff
@@ -0,0 +1,61 @@
...
+        for prompt in test_prompts:
+            insert_query = f"""INSERT INTO ImageGen (prompt) VALUES ('{prompt}')"""
+            execute_query_fetch_all(self.evadb, insert_query)
+
+    def tearDown(self) -> None:
+        execute_query_fetch_all(self.evadb, "DROP TABLE IF EXISTS ImageGen;")
+
+    @stable_diffusion_skip_marker
+    @patch("replicate.run", return_value=[{"response": "mocked response"}])
+    def test_stable_diffusion_image_generation(self, mock_replicate_run):
```
xzdandy, line 44: Similar to test_dalle.py, this is a good unit tests.

Does StableDiffusion can run with some test api keys? If so, we can make it an integration test and use @pytest.mark.xfail(reason="") for the rate limit.

test/integration_tests/long/functions/test_selfdiffusion.py
```diff
@@ -0,0 +1,61 @@
...
+        test_prompts = ["pink cat riding a rocket to the moon"]
+
+        for prompt in test_prompts:
+            insert_query = f"""INSERT INTO ImageGen (prompt) VALUES ('{prompt}')"""
+            execute_query_fetch_all(self.evadb, insert_query)
+
+    def tearDown(self) -> None:
+        execute_query_fetch_all(self.evadb, "DROP TABLE IF EXISTS ImageGen;")
+
+    @stable_diffusion_skip_marker
```
xzdandy, line 42: You also need to add dependency into setup.py. Otherwise, this test will be skipped on circle ci.

evadb/functions/stable_diffusion.py
sudoboi: The response is the link to an image.

evadb/functions/stable_diffusion.py
sudoboi: Oh if you're referring to stability-ai/...., that's not a directory. That is the model that I am specifying to the API.

test/integration_tests/long/functions/test_selfdiffusion.py
sudoboi: I did not add the dependency to setup.py because I'm not sure if we want to add the API library that I am using as a core dependency for EvaDB.

test/integration_tests/long/functions/test_selfdiffusion.py
sudoboi: It does run with API keys, but I figured we cannot depend on the API working every time we want to run a test. That's why I'm just checking that the API function is being at least called. I'm not familiar with  pytest.mark.xfail. If you think we can use that for the same purpose, I can check it out.

evadb/functions/function_bootstrap_queries.py
sudoboi: I did not add it to the list of queries to be loaded by default everytime, so technically yeah, I can remove these safely. I just added it in case we need it later.

xzdandy: Let's add them to the default every time. What do you think? Those two functions's setup does not do much. So there won't be a lot of overheads.

evadb/functions/stable_diffusion.py
xzdandy: I see. Should the response be an image? If the response is the link , how do we compose query? Is the idea we will introduce another generic function `DOWNLOAD_IMAGE` to turn the link into images?

evadb/functions/stable_diffusion.py
xzdandy: Ah. Thanks for the clarification. Is this model public accessible?

test/integration_tests/long/functions/test_selfdiffusion.py
xzdandy: That is to use mark flaky test. So we will run the test every time. But it will not cause the circle ci check to fail if the test failed.

evadb/functions/function_bootstrap_queries.py
sudoboi: Makes sense. I'll add it then.

test/integration_tests/long/functions/test_selfdiffusion.py
xzdandy: You do not need to add to core dependency. You can create an extra dependency for this library, and add that to the dev https://github.com/georgia-tech-db/evadb/blob/e8a181c5c34a6e58cc4d3901eae7b20f45b48940/setup.py#L170

So the circle ci will install the dependency and run the test.

evadb/functions/stable_diffusion.py
sudoboi: If you mean public as in anyone can use it with an API, then yes.

evadb/functions/stable_diffusion.py
sudoboi: I think that might be better. Because I think the links are more efficient to store considering some others are planning to work on semantic caching for this task specifically. I'd like to know your thoughts on this too.

xzdandy: Does the link have an expiration time? Or it is always available. If it is permanent, storing the link sounds good!

xzdandy:
Hi @sudoboi, I do have permission to push to your fork since I am not the reviewer. I need to change the script/test/test.sh to skip the tutorials/18-stable-diffusion.ipynb due to the replicate API key. Overall looks great! Thanks!

test/integration_tests/long/functions/test_stablediffusion.py
```diff
@@ -0,0 +1,68 @@
...
+            execute_query_fetch_all(self.evadb, insert_query)
+
+    def tearDown(self) -> None:
+        execute_query_fetch_all(self.evadb, "DROP TABLE IF EXISTS ImageGen;")
+
+    @stable_diffusion_skip_marker
+    @pytest.mark.xfail(
+        reason="API call might be flaky due to rate limits or other issues."
+    )
+    def test_stable_diffusion_image_generation(self):
```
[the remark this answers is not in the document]
xzdandy: I think it is good to have one test enable in continuous test to make sure other system change do not break the stable diffusion.

xzdandy:
Hi @gaurav274 and @jiashenC, please help review the PR. The PR looks good to me. But I reopened the PR, so I can not approve.

jiashenC (approved):

PR #1244, opened 2023-09-30
feat: create index from projection

jiashenC:
The first step to do automatic index updates on insertions. 

Replace the old version of creating an index, which directly reads data from the storage engine. 

It now reads data from the children's plans: SeqScan and Storage.

### DIFF
--- a/evadb/binder/create_index_statement_binder.py
+++ b/evadb/binder/create_index_statement_binder.py
@@ -12,14 +12,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from evadb.binder.binder_utils import BinderError
+from evadb.binder.binder_utils import BinderError, create_row_num_tv_expr, extend_star
 from evadb.binder.statement_binder import StatementBinder
 from evadb.catalog.catalog_type import NdArrayType, VectorStoreType
+from evadb.expression.function_expression import FunctionExpression
+from evadb.expression.tuple_value_expression import TupleValueExpression
 from evadb.parser.create_index_statement import CreateIndexStatement
 from evadb.third_party.databases.interface import get_database_handler
-from evadb.expression.tuple_value_expression import TupleValueExpression
-from evadb.expression.function_expression import FunctionExpression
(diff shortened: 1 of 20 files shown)


evadb/binder/create_index_statement_binder.py
```diff
@@ -17,12 +17,32 @@
...
     binder.bind(node.table_ref)
-    if node.function:
-        binder.bind(node.function)
+
+    func_project_expr = None
+
+    # Extend projection list.
+    project_expr_list = []
+    for project_expr in node.project_expr_list:
+        if isinstance(project_expr, TupleValueExpression) and project_expr.name == "*":
```
xzdandy, line 33: Can we create index on multiple columns now?

xzdandy:
The design looks good. For record, the plan we are creating is `CREATE_INDEX` - `PROJECTION` - `GET`.

evadb/binder/create_index_statement_binder.py
jiashenC: Not really. I use this to project the raw table column if index is built on table directly. I use `*` just for simplicity for now.

xzdandy: If we have a query `CREATE INDEX test_index ON myTable (myCol);`, can the `project_expr_list` be `[TupleValueExpression(myCol)]` instead of `[TupleValueExpression(*)]`?

evadb/catalog/catalog_manager.py
```diff
@@ -414,9 +414,15 @@ def insert_index_catalog_entry(
         vector_store_type: VectorStoreType,
         feat_column: ColumnCatalogEntry,
         function_signature: str,
+        index_def: str,
```
gaurav274, line 417: Please add the docs for func parameters or describe the need for `index_def` in the catalog file.

gaurav274:
> The first step to do automatic index updates on insertions.

I might be missing something, but I could not understand how "automatic index updates" is done in the code.

gaurav274 (approved):

evadb/binder/create_index_statement_binder.py
jiashenC: Fixed.

PR #1245, opened 2023-10-02
Documentation on vector stores + vector benchmark

Chitti-Ankith:
Added documentation for vector stores including usage examples, dependencies and other requirements.

### DIFF
--- a/docs/source/reference/vector_stores/chromadb.rst
+++ b/docs/source/reference/vector_stores/chromadb.rst
@@ -1,0 +1,17 @@
+ChromaDB
+==========
+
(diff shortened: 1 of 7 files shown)


docs/source/reference/vector_stores/chromadb.rst
```diff
@@ -0,0 +1,17 @@
...
+
+Dependency
+----------
+
+* chromadb
+
+Create Index
+-----------------
+
+.. code-block:: text
```
xzdandy, line 15: Does `.. code-block:: sql` does not work?

xzdandy (approved):
Thanks for adding the documentation for vector stores!

PR #1246, opened 2023-10-02
feat: insertion update index

jiashenC:
Break the feature into multiple PRs. 

We can merge this PR after https://github.com/georgia-tech-db/evadb/pull/1244.


evadb/executor/insert_executor.py
```diff
@@ -53,6 +53,25 @@ def exec(self, *args, **kwargs):
...
+            if is_index_on_current_table:
+                create_index_query_list = index.index_def.split(" ")
+                if_not_exists = " ".join(create_index_query_list[2:5]).lower()
+                if if_not_exists != "if not exists":
+                    create_index_query = (
+                        " ".join(create_index_query_list[:2])
+                        + " IF NOT EXISTS "
+                        + " ".join(create_index_query_list[2:])
+                    )
+                from evadb.server.command_handler import execute_query_fetch_all
```
xzdandy, line 71: What is the problem if we directly run the original create_index query instead of doing this parsing and reconstructing the query?

test/integration_tests/long/test_similarity.py
```diff
@@ -402,6 +408,30 @@ def test_end_to_end_index_scan_should_work_correctly_on_image_dataset_faiss(self
...
+        select_query = """SELECT _row_id FROM testIndexAutoUpdate
+                                ORDER BY Similarity(DummyFeatureExtractor(Open("{}")), DummyFeatureExtractor(Open(img_path)))
+                                LIMIT 1;""".format(
+            self.img_path
+        )
+        explain_batch = execute_query_fetch_all(self.evadb, f"EXPLAIN {select_query}")
+        self.assertTrue("VectorIndexScan" in explain_batch.frames[0][0])
+
+        res_batch = execute_query_fetch_all(self.evadb, select_query)
+        self.assertEqual(res_batch.frames["testindexautoupdate._row_id"][0], 5)
```
xzdandy, line 433: Does this testcase ensure the later insert triggers the create_index?

jiashenC: Yes, otherwise the returned closest vector will be different.

evadb/executor/insert_executor.py
jiashenC: I currently only support index updates when `IF NOT EXISTS` exists. Does that make sense?

xzdandy (approved):
Looks good. Can we add to the documentation that insert into a table with index created will auto update the index? If there is limitation now, for example only works with sqlite and faiss, we can also mention there. 

Thanks!

PR #1254, opened 2023-10-04
feat: run additional functional expression during an index scan

jiashenC:
I am currently trying to support the below query
```sql
SELECT _row_id, Similarity(DummyFeatureExtractor(Open("{}")), DummyFeatureExtractor(data)) 
  FROM testSimilarityImageDataset
  ORDER BY Similarity(DummyFeatureExtractor(Open("{}")), DummyFeatureExtractor(data))
  LIMIT 1
```

It will be converted to such a plan with my current change.
```
|__ ProjectPlan
    |__ VectorIndexScanPlan
        |__ ApplyAndMergePlan
            |__ SeqScanPlan
                |__ StoragePlan
```

It will become very inefficient for the function expression because it is applied on all rows. Any possible way to overcome this? @xzdandy @gaurav274

### DIFF
--- a/evadb/optimizer/rules/rules.py
+++ b/evadb/optimizer/rules/rules.py
@@ -48,6 +48,7 @@ from evadb.plan_nodes.nested_loop_join_plan import NestedLoopJoinPlan
 from evadb.plan_nodes.predicate_plan import PredicatePlan
 from evadb.plan_nodes.project_plan import ProjectPlan
 from evadb.plan_nodes.show_info_plan import ShowInfoPlan
+from evadb.utils.logging_manager import logger
 
 if TYPE_CHECKING:
     from evadb.optimizer.optimizer_context import OptimizerContext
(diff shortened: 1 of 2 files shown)


gaurav274:
> I am currently trying to support the below query
> 
> ```sql
> SELECT _row_id, Similarity(DummyFeatureExtractor(Open("{}")), DummyFeatureExtractor(data)) 
>   FROM testSimilarityImageDataset
>   ORDER BY Similarity(DummyFeatureExtractor(Open("{}")), DummyFeatureExtractor(data))
>   LIMIT 1
> ```
> 
> It will be converted to such a plan with my current change.
> 
> ```
> |__ ProjectPlan
>     |__ VectorIndexScanPlan
>         |__ ApplyAndMergePlan
>             |__ SeqScanPlan
>                 |__ StoragePlan
> ```
> 
> It will become very inefficient for the function expression because it is applied on all rows. Any possible way to overcome this? @xzdandy @gaurav274

Can you remind me why there is an `ApplyAndMergePlan` node in the query plan?

xzdandy:
LIMIT needs to be push down below the final projection. Does that solves the issue?

jiashenC:
`AppyAndMerge` is added when I try to select a functional expression. In this case, the `Similarity(DummyFeatureExtractor(Open("{}")), DummyFeatureExtractor(data))` is used in the `ApplyAndMerge` and applied to every row. For real use case, the feature extractor will be very expensive so it will slow down the query significantly.

I also tried some other queries `SELECT FUNC1(...), FUNC2(...) ORDER BY ... LIMIT ...`. I observed all of those functional expressions are done in `ApplyAndMerge` first before `OrderBy` and `Limit`.

jiashenC:
I think we can close this for now because it is no longer relevant. We can do projection through a subquery.

PR #1256, opened 2023-10-05
Remove table names from column names for `df()` call

pchunduri6:
Removing table names from the `dataframe` during `df()` call. The users can then easily load CSV files generated using `EvaDB` with the `to_csv()` call at a later time (for long-running or expensive queries).

Example:

```
select_query = cursor.query(
    f"SELECT * FROM {repo_name}_StargazerList;"
).df()

select_query.to_csv("stargazers_list.csv", index=False)

# Later
cursor.query(
        f"""
   CREATE TABLE IF NOT EXISTS {repo_name}_StargazerList(
   github_username TEXT(1000));
"""
    ).df()

cursor.query("LOAD CSV 'stargazers_list.csv' INTO {repo_name}_StargazerList;""").df()

```

Do we need the table names for any use cases? For example, for duplicate column names from two different functions - `object_detector_1.labels` and `object_detector_2.labels`?


xzdandy (approved):

xzdandy:
No merge now. Making minor changes to PR.

Issue #1243, opened 2023-09-30
Enhance forecasting with more features


americast:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar feature requests.

### Description

It would be nice to have the following additional features for EvaDB:
1. Add GPU support
2. Static exogenous variables
3. Multivariate forecasting
4. Error when there are <=1 data points
5. Explain `unique_id` better
6. catch frequency related errors better

### Use case

1. GPU support is essential in order to perform autoML techniques quickly. Currently, it has been disabled owing to instability with `ray`.
2. Right now, every exogenous variable is considered temporal in nature. It is important to distinguish them from static ones.
3. Right now, we can use `PREDICT` in only one column. It would be great to be able to predict forecasting for multiple columns at once.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

khushitalesra:
If the price column has comma then forecasting doesn't work. Below error is encountered
ValueError: could not convert string to float: '270,000'
EvaDB Session ended with an error: could not convert string to float: '270,000'

xzdandy:
@gaurav274 LLM based data cleaning can help this case. In this case, without LLM it is also possible, but I think LLM is more general and flexible. One optimization is to only run LLM data cleaning on the tuple that failed instead of all tuples, which saves the cost and time. 

An alternative approach is to skip the rows that does not match the type.
@khushitalesra Is the price column type float? But the content is string in this case.

khushitalesra:
yes earlier I had defined the price column to string because of the comma but then I changed it to float as it was throwing error

gaurav274:
> @gaurav274 LLM based data cleaning can help this case. In this case, without LLM it is also possible(e.g., REGEX), but I think LLM is more general and flexible. One optimization is to only run LLM data cleaning on the tuple that failed instead of all tuples, which saves the cost and time. Another optimization is to choose the LLM model, for a simple task like this we may use a lightweight local model.
> 
> An alternative approach is to skip the rows that does not match the type. @khushitalesra Is the price column type float? But the content is string in this case.

@xzdandy This is a nice application of format inconsistencies.

gaurav274:
@americast let us break this into sub-tasks and assign a separate issue to each. Thanks!

americast:
[...]

PR #1258, opened 2023-10-05
Add feedback for forecasting

### LINKED ISSUE #1257: Feedback for the user when using the `FORECASTING` UDF
### Search before asking

### Description

We provide a whole bunch of options to the user when creating the `FORECASTING` UDF, however, it might not be trivial to realize which options would work the best. It would be great if feedback could be provided when a UDF is called and performs subpar.

### Use case

Users using the `FORECASTING` UDF and having complicated datasets.

americast:
Provide feedback when `Forecasting` UDF is called in the following ways:

- [x] Reporting confidence intervals
- [x] Returning a metric for the forecasting performance.
- [x] Providing suggestions in simple special cases, such as during Flat predictions.

Eg:
```sql
SELECT HomeForecast();
```

```
SUGGESTION: Predictions are flat. Consider using LIBRARY 'neuralforecast' for more accrate predictions.
```

Partially fixes #1257 and #1243.


docs/source/reference/ai/model-forecasting.rst
```diff
@@ -28,16 +28,18 @@ Next, we create a function of `TYPE Forecasting`. We must enter the column name
...
 
 .. code-block:: sql
 
-   SELECT Forecast(12);
+   SELECT Forecast();
 
-Here, the horizon is `12`, which represents the forecast 12 steps into the future.
+.. note::
 
+   `Forecasting` function also provides suggestions by default. If you wish to turn it off, send "False" as an optional argument while calling the function. Eg. `SELECT Forecast("False");`
```
xzdandy, line 42: Shall we do boolean instead of string?  `SELECT Forecast(FALSE);`

evadb/functions/forecast.py
```diff
@@ -48,13 +48,33 @@ def setup(
...
+                        forecast_df.loc[forecast_df["unique_id"] == type_here][
+                            self.model_name
+                        ].nunique()
+                        == 1
+                    ):
+                        suggestion_list.append(1)
+
+            for suggestion in set(suggestion_list):
+                print("\nSUGGESTION: " + self.suggestion_dict[suggestion])
+
```
xzdandy, line 77: For cases like " <=1 data point", do we also want to provide the suggestions? If so, the suggestion needs to be at the `CREATE` time.

americast: This issue has been fixed in #1283.

docs/source/reference/ai/model-forecasting.rst
americast: Thanks, changed it.

americast:
Another concern with this approach is that the suggestion is being provided through a `print` statement. Hence, it would appear on the server side, and the user might miss it. Any hints on how we can send the suggestion to the user?

xzdandy:
> Another concern with this approach is that the suggestion is being provided through a `print` statement. Hence, it would appear on the server side, and the user might miss it. Any hints on how we can send the suggestion to the user?

Good point. We can use logger and capture it and send to the client.

xzdandy:
Hi @americast, thanks for adding for suggestions a metric for the forecasting performance. Do you think we can add that information into `evadb/catalog/models/function_catalog.py` or `evadb/catalog/models/function_metadata_catalog.py`? We can decide which catalog table is better. The idea is to be able to query this information afterwards.

americast:
> Hi @americast, thanks for adding for suggestions a metric for the forecasting performance. Do you think we can add that information into `evadb/catalog/models/function_catalog.py` or `evadb/catalog/models/function_metadata_catalog.py`? We can decide which catalog table is better. The idea is to be able to query this information afterwards.
> 
> Beside, we can also output this information as part of the output of `CREATE FUNCTION`, so user can directly view them.

Thanks @xzdandy. We can take care of this in a later PR.

evadb/binder/statement_binder.py
```diff
@@ -194,12 +208,6 @@ def _bind_delete_statement(self, node: DeleteTableStatement):
 
     @bind.register(CreateTableStatement)
     def _bind_create_statement(self, node: CreateTableStatement):
-        # we don't allow certain keywords in the column_names
-        for col in node.column_list:
-            assert (
-                col.name.lower() not in RESTRICTED_COL_NAMES
-            ), f"EvaDB does not allow to create a table with column name {col.name}"
-
```
xzdandy, line 202: I think we should not remove this. This causes testcase failures and we can not allow user to create columns with certain keywords.

test/integration_tests/long/test_model_forecasting.py
```diff
@@ -143,7 +156,13 @@ def test_forecast_with_column_rename(self):
...
         self.assertEqual(
             result.columns,
-            ["homeforecast.type", "homeforecast.saledate", "homeforecast.ma"],
+            [
+                "homeforecast.type",
+                "homeforecast.saledate",
+                "homeforecast.ma",
+                "homeforecast.ma-lo",
+                "homeforecast.ma-hi",
+            ],
```
xzdandy, line 165: To save time, we can keep one testcases for neural forecasting and one for statsforecast. Does the feedback add a lot of execution overhead? We did not timeout for a 30 minutes test before.

americast: Hmm, this a little weird. The feedback system is turned off by default for `neuralforecast`.

evadb/binder/statement_binder.py
americast: Thanks, updated it.

evadb/executor/create_function_executor.py
```diff
@@ -22,7 +22,9 @@
 from pathlib import Path
 from typing import Dict, List
 
+import numpy as np
 import pandas as pd
+from sklearn.metrics import mean_squared_error
```
xzdandy, line 27: This breaks the documentation build and other usecases, the sklearn is not a default dependency in EvaDB. Either moves the import inside the forecast and use try_to_import_sklearn or add sklearn to the default dependency.

test/integration_tests/long/test_model_forecasting.py
xzdandy: There are two statsforecast tests in the test_model_forecasting. The feedback system is running for those two tests. That takes time?

americast: Not really. Crossvalidation for statsforecast is usually quick. I'll take a look.

evadb/executor/create_function_executor.py
americast: Thanks @xzdandy! `mean_squared_error` is essentially a single line of code in `numpy`. I'll change it.

test/integration_tests/long/test_model_forecasting.py
americast: This is the exact time out error. I'll try to figure out the cause:

```
FAILED test/integration_tests/long/test_model_forecasting.py::ModelTrainTests::test_forecast - evadb.executor.executor_utils.ExecutorError: Timed out initializing process group in store based barrier on rank: 0, for key: store_based_barrier...
```

Perhaps it's something to do with pytorch?
https://discuss.pytorch.org/t/timed-out-initializing-process-group-in-store-based-barrier/119803/20

xzdandy: I do not think that pytorch is used in statsforecasting. Is the cross validation working locally for you?

americast:
@xzdandy The issue seems to be coming up due to multithreading while training `neuralforecast` model on the GPU -- the entire test suite seems to be getting spawned into multiple threads and thus it appears to have got hanged. For now I have removed test for `neuralforecast`.

test/integration_tests/long/test_model_forecasting.py
```diff
@@ -103,8 +110,6 @@ def test_forecast(self):
             TYPE Forecasting
             HORIZON 12
             PREDICT 'y'
-            LIBRARY 'neuralforecast'
-            AUTO 'false'
```
xzdandy, line 107: For my education, is neuralforcast not working? There is no GPU in the circleci, that's where I am getting confused.

americast: I see, was not aware of that. Neuralforecast works fine mostly, except if a corresponding function is created in the same python file where other things are done as well, like creating table etc. This is a GPU-specific error, which I am able to locally reproduce. In the CPU however, Neuralforecast will take a lot of time to train.

xzdandy (approved):

Issue #1140, opened 2023-09-17
Configuration Manager Redesign 


xzdandy:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar feature requests.

### Description

We'd like to move away from the yml file, and use a catalog to store all EvaDB's configuration (e.g., ray enabled or not). We'd like also to introduce a new SQL command to adjust those configurations. 

- [x] Introduce `SET` commands to change the configuration as needed. #1180
- [ ] Switch the configuration into a persistent SQL table.
- [ ] Introduce a `GET` -like commands to get the current configuration value. 

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

hershd23:
I have captured the main ideas in this doc https://docs.google.com/document/d/1tnePjSJqj9zuhHc8Cu-220tjh0Sv0uOlSUYWx-L6wjA/edit?usp=sharing

Right now #1180 takes care of setting configuration parameters. Will have the `get` calls for configs and configuration manager persistance in separate PRs.

Wanted to get your opinion on the `get` calls for config params.
Duck DB does something like https://duckdb.org/docs/sql/configuration.html

```SELECT current_setting('access_mode');```

And Postgres does it with https://www.postgresql.org/docs/current/sql-show.html

```Show <name>```

Which approach should we take, any ideas? @xzdandy @gaurav274 @jiashenC @jarulraj

xzdandy:
Thanks Hersh for the design doc. Provided feedback there. I like the `Show` better, which is consistent with the current `SHOW TABLES` and `SHOW FUNCTOINS` now, though we will need some design like `SHOW SETTINGS` to differentiate from TABLE and FUNCTIONS, so the syntax will not be identical to Postgres.  

Wrapping up #1180 for `SET` and having other changes in following PRs sounds good to me.

PR #1259, opened 2023-10-06
[WIP] Configuration Manager Redesign

hershd23:
Partially solves #1140 
Adding EvaDb changes to the catalog
```
	modified:   evadb/catalog/catalog_manager.py
	new file:   evadb/catalog/models/configuration_catelog.py
	modified:   evadb/catalog/models/utils.py
	new file:   evadb/catalog/services/configuration_catalog_service.py
	modified:   evadb/catalog/sql_config.py
```
Adding EvaDB base configs to a python dict
```
	new file:   evadb/eva_config.py
```


hershd23:
Up for an early review, still in WIP

hershd23:
https://github.com/georgia-tech-db/evadb/blob/62080794c16339f52e17df74cd7956380717e95a/evadb/database.py#L35

@gaurav274 @xzdandy there is one dependency issue that I see with config. The problem with initializaing config as a table in catalog is because catalog itself depends on config as you can see in the method above. Any suggestions on what to do in order to remove this?

xzdandy:
> https://github.com/georgia-tech-db/evadb/blob/62080794c16339f52e17df74cd7956380717e95a/evadb/database.py#L35
> 
> @gaurav274 @xzdandy there is one dependency issue that I see with config. The problem with initializaing config as a table in catalog is because catalog itself depends on config as you can see in the method above. Any suggestions on what to do in order to remove this?

At the first glance, I think we need to decouple the catalog and configuration manager. 
1) Move `catalog_database_uri` out of config, since it can not be changed once a evadb instance is created. 
2) After we move the `catalog_database_uri` out of config, we than no longer need to pass config into catalog. So we can initialize all catalog tables (including the new configuration_catalog) in the catalog manager.

hershd23:
@xzdandy I think the dependency doesn't end there, I saw a few more examples of the dependency

gaurav274:
Moved to #1272

PR #1264, opened 2023-10-06
SHOW command for retrieveing configurations

hershd23:


evadb/executor/show_info_executor.py
```diff
@@ -43,5 +45,11 @@ def exec(self, *args, **kwargs):
                 if table.table_type != TableType.SYSTEM_STRUCTURED_DATA:
                     show_entries.append(table.name)
             show_entries = {"name": show_entries}
+        elif self.node.show_type is ShowType.CONFIG:
+            value = self._config.get_value(
```
gaurav274, line 49: We need to handle cases where the key is not in the config. Return an appropriate error message. Also, add a test case of it. Thanks!

evadb/parser/lark_visitor/_show_statements.py
```diff
@@ -23,7 +23,9 @@ class Show:
...
         token = tree.children[1]
 
-        if str.upper(token) == "FUNCTIONS":
+        if isinstance(token, str) and str.upper(token) == "FUNCTIONS":
             return ShowStatement(show_type=ShowType.FUNCTIONS)
-        elif str.upper(token) == "TABLES":
+        elif isinstance(token, str) and str.upper(token) == "TABLES":
             return ShowStatement(show_type=ShowType.TABLES)
+        elif token is not None:
+            return ShowStatement(show_type=ShowType.CONFIG, show_val=self.visit(token))
```
gaurav274, line 31: We can make all the keys case-insensitive and store them in all caps.

hershd23: Done

evadb/executor/show_info_executor.py
hershd23: Have raised an exception for that for now. Let me know if you think that there is a better way to do it

xzdandy (approved):
Looks great! Thank!

xzdandy:
I think we also need to make all existing code use the `default` catalog.

gaurav274:
> I think we also need to make all existing code use the `default` catalog.

Do you mean the `category`?

xzdandy:
> > I think we also need to make all existing code use the `default` catalog.
> 
> Do you mean the `category`?

Yes. Otherwise, the `SET` and `SHOW` do not affect the current codebase.

hershd23:
Yeah, the current cursor utility works and does not affect existing code. This PR does not remove category from the codebase the earlier one does #1259. I'll make the changes as part of that PR

xzdandy:
> Yeah, the current cursor utility works and does not affect existing code. This PR does not remove category from the codebase the earlier one does #1259. I'll make the changes as part of that PR

I see. We need to think about the efforts that we need to move the configuration manager to table. The `SET` and `SHOW` are very useful even without migrating configuration manager to table. So if the migration will take time, I suggest merge and  enable the `SET` and `SHOW` for the current configuration manager

hershd23:
@xzdandy yes, the SHOW and SET implementations are completely independent of the configuration manager implementation. Only the executor code will change. I would say we can merge this and continue work on #1259. I have an idea of how I can do that as well.

PR #1269, opened 2023-10-09
Add Milvus integration for vector create and search

RichardZhangRZ:
Integrated Milvus vector store into EvaDB. Added a `MilvusVectorStore` class and Milvus type for query parsing and execution.
Below are environment values for the use of the Milvus index:

* `MILVUS_URI` is the URI of the Milvus instance (which would be http://localhost:19530 when running locally). **This value is required**
* `MILVUS_USER` is the name of the user for the Milvus instance.
* `MILVUS_PASSWORD` is the password of the user for the Milvus instance.
* `MILVUS_DB_NAME` is the name of the database to be used. This will default to the `default` database if not provided.
* `MILVUS_TOKEN` is the authorization token for the Milvus instance.


xzdandy:
Hi Richard, thanks for the contribution! What is the status of the pull request?

evadb/third_party/vector_stores/milvus.py
```diff
@@ -0,0 +1,149 @@
...
+                "vector": feature_payload.embedding.reshape(-1).tolist(),
+            }
+            for feature_payload in payload
+        ]
+        ids = [feature_payload.id for feature_payload in payload]
+
+        # Milvus Client does not have upsert operation, perform delete + insert to emulate it
+        self._client.delete(collection_name=self._collection_name, pks=ids)
+
+        self._client.insert(collection_name=self._collection_name, data=milvus_data)
```
Chitti-Ankith, line 126: What happens if we directly call insert on existing ids?

setup.py
```diff
@@ -112,6 +112,8 @@ def read(path, encoding="utf-8"):
 
 chromadb_libs = ["chromadb"]
 
+milvus_libs = ["pymilvus>=2.3.0"]
```
Chitti-Ankith, line 115: Are there any breaking changes in the previous versions?

evadb/third_party/vector_stores/milvus.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def persist(self):
+        self._client.flush(self._collection_name)
```
Chitti-Ankith, line 129: @xzdandy is the persist function needed? When would it be called, I don't think its present for all the vector stores rn

evadb/third_party/vector_stores/milvus.py
Chitti-Ankith: Can this not be used ? https://milvus.io/docs/upsert_entities.md

RichardZhangRZ: There is no `upsert` for the Milvus client object itself. Just inserting on existing ids will not update it.

setup.py
RichardZhangRZ: Only Milvis >= 2.3.0 has native support for cosine similarity, which the `MilvusVectorStore` uses for searching. I chose cosine similarity since the other vector index integration uses it. If we were to switch to IP or L2, then earlier versions can be used.

evadb/third_party/vector_stores/milvus.py
xzdandy: `persist` is called in create_index_exectuor. https://github.com/georgia-tech-db/evadb/blob/a64d24bb7c2864c42e60ff9fa9890c4ac137bd0b/evadb/executor/create_index_executor.py#L163

@jiashenC who has more knowledge in this space.

jiashenC: We can probably just `pass` for this method. It is only needed for `FAISS` to persist it on disk. I think other vector stores will take care of persistence by themselves.

evadb/third_party/vector_stores/milvus.py
jiashenC: Had the same question. Can we get collection from the client and upsert using collection APIs? Maybe I missed something here.

RichardZhangRZ: There seems to be no methods in the client to get the collection objects, just the collection names through `list_collections`. Maybe something like this could work:
```python
client_connection_alias = self._client._using
collection = Collection(self._collection_name, using=client_connection_alias)
collection.upsert(milvus_data)
```
My biggest concern with this though is the use of the private `_using` variable in the client.

test/integration_tests/long/test_similarity.py
```diff
@@ -523,3 +543,33 @@ def test_end_to_end_index_scan_should_work_correctly_on_image_dataset_pinecone(
 
             drop_index_query = "DROP INDEX testpineconeindeximagedataset;"
             execute_query_fetch_all(self.evadb, drop_index_query)
+
+    @pytest.mark.skip(reason="Requires running local Milvus instance")
```
xzdandy, line 547: Is it possible to set up a local Milvus instance for testing in circle ci?

evadb/third_party/vector_stores/milvus.py
RichardZhangRZ: Milvus does not seem to immediately persist inserted entities to disk. I think that `flush` needs to be called to ensure that entities inserted into a collection can be immediately indexed and searched. [Reference](https://milvus.io/api-reference/pymilvus/v2.3.x/Collection/flush().md)

xzdandy:
Hi @RichardZhangRZ , is this PR ready? Thanks!

test/integration_tests/long/test_similarity.py
RichardZhangRZ: Currently trying this out

RichardZhangRZ: Seems like Milvus Lite (a library to spin up a local instance in Python code) only has an outdated version of Milvus and therefore does not support cosine similarity. Therefore, this doesn't seem possible to do :(

RichardZhangRZ:
> Hi @RichardZhangRZ , is this PR ready? Thanks!

There are still ongoing conversations, but other than that, the PR should be ready

.circleci/config.yml
```diff
@@ -486,7 +486,7 @@ jobs:
               source test_evadb/bin/activate
               pip install --upgrade pip
               pip debug --verbose
-              pip install ".[dev,ludwig,qdrant,forecasting,pinecone,chromadb]"
+              pip install ".[dev,ludwig,qdrant,forecasting,pinecone,chromadb,milvus]"
```
xzdandy, line 489: If the test is not running on circle ci, shall we skip the install of milvus?

xzdandy (approved):
Thanks Richard! Very nice PR!

xzdandy:
We need to address some merging conflicts.

xzdandy:
The merge does not look correct. I will give it a try.

.circleci/config.yml
```diff
@@ -230,16 +230,16 @@ jobs:
...
+                pip install ".[dev,pinecone,chromadb,milvus]" # ray < 2.5.0 does not work with python 3.11 ray-project/ray#33864
               fi
               python -c "import evadb;cur=evadb.connect().cursor();cur.query('SET ray=True;')"
             else
               if [ $PY_VERSION != "3.11" ]; then
-                pip install ".[dev,ludwig,qdrant,pinecone,chromadb]"
+                pip install ".[dev,ludwig,qdrant,pinecone,chromadb,milvus]"
               else
-                pip install ".[dev,pinecone,chromadb]" # ray < 2.5.0 does not work with python 3.11 ray-project/ray#33864
+                pip install ".[dev,pinecone,chromadb,milvus]" # ray < 2.5.0 does not work with python 3.11 ray-project/ray#33864
```
xzdandy, line 242: Remove `milvus` since we are running any test.

docs/source/reference/vector_databases/milvus.rst
```diff
@@ -0,0 +1,31 @@
...
+To use Milvus you must have a URI to a running Milvus instance. Here are the `instructions to spin up a local instance <https://milvus.io/docs/install_standalone-docker.md>`_.
+If you are running it locally, the Milvus instance should be running on ``http://localhost:19530``. Please be sure that the Milvus version is >= 2.3.0. Below are values that the Milvus integration uses:
+
+* `MILVUS_URI` is the URI of the Milvus instance (which would be ``http://localhost:19530`` when running locally). **This value is required**
+* `MILVUS_USER` is the name of the user for the Milvus instance.
+* `MILVUS_PASSWORD` is the password of the user for the Milvus instance.
+* `MILVUS_DB_NAME` is the name of the database to be used. This will default to the `default` database if not provided.
+* `MILVUS_TOKEN` is the authorization token for the Milvus instance. 
+
+The above values can either be set in the evadb.yml config file, or in the os environment fields "MILVUS_URI", "MILVUS_USER", "MILVUS_PASSWORD", "MILVUS_DB_NAME", and "MILVUS_TOKEN"
```
xzdandy, line 24: Use `SET` command is the way to do it now. Example: `SET MILVUS_URI = 'http://localhost:19530;'`. We no longer have evadb.yml file. OS environment is not recommended, since it does not work in a client-server setup.

PR #1272, opened 2023-10-10
Configuration Manager Redesign

gaurav274:
1. Removed `config.yml` file. Users can directly use `SET` command. 
2. Moved `OPENAI_KEY` to `OPENAI_API_KEY`


evadb/catalog/models/configuration_catalog.py
```diff
@@ -0,0 +1,45 @@
...
+    `_key:` the key for the config.
+    `_value:` the value for the config
+    """
+
+    __tablename__ = "configuration_catalog"
+
+    _key = Column("name", String(100), unique=True)
+    _value = Column(
+        "engine", TextPickleType()
+    )  # TODO :- Will this work or would we need to pickle ??
```
hershd23, line 34: I had left this comment actually to see if a singleton value like `_value` would get pickled after `json.dumps` that we do in the `TextPickleType()` method. Do check that once if that works.

Also DB column names are `name` and `engine` instead of `key` and `value`

evadb/configuration/bootstrap_environment.py
```diff
@@ -124,17 +104,15 @@ def create_directories_and_get_default_config_values(
...
+    config_obj["evadb_installation_dir"] = str(default_install_dir.resolve())
+    config_obj["datasets_dir"] = str(dataset_location.resolve())
+    config_obj["catalog_database_uri"] = get_default_db_uri(evadb_dir)
+    config_obj["index_dir"] = str(index_dir.resolve())
+    config_obj["cache_dir"] = str(cache_dir.resolve())
+    config_obj["s3_download_dir"] = str(s3_dir.resolve())
+    config_obj["tmp_dir"] = str(tmp_dir.resolve())
+    config_obj["function_dir"] = str(function_dir.resolve())
+    config_obj["model_dir"] = str(model_dir.resolve())
     if category and key:
```
hershd23, line 116: Category is removed now

hershd23 (commented):
Just done a scan and left a couple comments. One was unresolved in my commits itself

evadb/catalog/models/configuration_catalog.py
gaurav274: Why are you calling them `name` and `engine`?

hershd23: No no, I mean I forgot to update them in code. Just pointing it out so that you change that. :sweat_smile:

xzdandy:
I fixed the merge conflict. Please check if I do it in the right way.

evadb/binder/statement_binder.py
```diff
@@ -331,9 +337,17 @@ def _bind_func_expr(self, node: FunctionExpression):
                 )
                 # certain functions take additional inputs like yolo needs the model_name
                 # these arguments are passed by the user as part of metadata
-                node.function = lambda: function_class(
-                    **get_metadata_properties(function_obj)
-                )
+                # we also handle the special case of ChatGPT where we need to send the
+                # OpenAPI key as part of the parameter
+                # ToDO: this should be better handled
```
xzdandy, line 342: Do you think have a CREATE call like the following is the right approach? 
```
CREATE FUNCTION myllm
TYPE llm
MODEL GPT4
OPENAI key;
```
If so, we can mention in the TODO comment.

evadb/binder/statement_binder.py
```diff
@@ -331,9 +337,17 @@ def _bind_func_expr(self, node: FunctionExpression):
...
+                properties = get_metadata_properties(function_obj)
+                if node.name.upper() == str("CHATGPT"):
```
xzdandy, line 344: Shall we use `string_comparison_case_insensitive` in https://github.com/georgia-tech-db/evadb/blob/staging/evadb/utils/generic_utils.py#L614 for case insensitive string comparison?

evadb/configuration/bootstrap_environment.py
```diff
@@ -80,22 +59,22 @@ def bootstrap_environment(evadb_dir: Path, evadb_installation_dir: Path):
 
     assert evadb_dir.exists(), f"{evadb_dir} does not exist"
     assert evadb_installation_dir.exists(), f"{evadb_installation_dir} does not exist"
-    config_obj = {}
-    with default_config_path.open("r") as yml_file:
-        config_obj = yaml.load(yml_file, Loader=yaml.FullLoader)
     config_obj = merge_dict_of_dicts(config_default_dict, config_obj)
-    mode = config_obj["core"]["mode"]
+    mode = config_obj["mode"]
```
xzdandy, line 63: This is not useful. We can only change the mode using `SET` after already bootstrapping the environment.

evadb/executor/create_index_executor.py
```diff
@@ -75,7 +75,7 @@ def _create_native_index(self):
 
     # On-disk saving path for EvaDB index.
     def _get_evadb_index_save_path(self) -> Path:
-        index_dir = Path(self.config.get_value("storage", "index_dir"))
+        index_dir = Path(self.db.catalog().get_configuration_catalog_value("index_dir"))
```
xzdandy, line 78: Can we use `self.catalog` instead?

evadb/executor/executor_utils.py
```diff
@@ -178,7 +178,13 @@ def handle_vector_store_params(
     elif vector_store_type == VectorStoreType.CHROMADB:
         return {"index_path": str(Path(index_path).parent)}
     elif vector_store_type == VectorStoreType.PINECONE:
-        return {}
+        # add the required API_KEYS
+        return {
+            "PINECONE_API_KEY": catalog().get_configuration_catalog_value(
```
xzdandy, line 183: Is the reason we need to pass a function instead of a catalog instance because of ray?

evadb/parser/evadb.lark
```diff
@@ -84,7 +84,7 @@ set_statement: SET config_name (EQUAL_SYMBOL | TO) config_value
 
 config_name: uid
 
-config_value: (string_literal | decimal_literal | boolean_literal | real_literal)
+config_value: constant
```
xzdandy, line 87: Can the config value be a list?

evadb/third_party/vector_stores/pinecone.py
```diff
@@ -30,34 +29,30 @@
...
-        )
+        self._api_key = kwargs.get("PINECONE_API_KEY")
 
         if not self._api_key:
             self._api_key = os.environ.get("PINECONE_API_KEY")
 
         assert (
             self._api_key
-        ), "Please set your Pinecone API key in evadb.yml file (third_party, pinecone_api_key) or environment variable (PINECONE_KEY). It can be found at Pinecone Dashboard > API Keys > Value"
+        ), "Please set your `PINECONE_API_KEY` using set command or environment variable (PINECONE_KEY). It can be found at Pinecone Dashboard > API Keys > Value"
```
xzdandy, line 46: Considering the client-server case, I don't think environment variable is a good idea.

evadb/catalog/catalog_manager.py
```diff
@@ -608,14 +619,28 @@ def create_and_insert_multimedia_metadata_table_catalog_entry(
...
+    def upsert_configuration_catalog_entry(self, key: str, value: any):
+        """Upserts configuration catalog entry"
+
+        Args:
+            key: key name
+            value: value name
+        """
+        self._config_catalog_service.upsert_entry(key, value)
+
+    def get_configuration_catalog_value(self, key: str) -> Any:
```
xzdandy, line 633: Is having a default value better than return None?

evadb/third_party/vector_stores/pinecone.py
gaurav274: Yup, I agree. We can fix it in a later PR.

evadb/parser/evadb.lark
gaurav274: `constant` also includes specifying a list

evadb/binder/statement_binder.py
gaurav274: This should already work.  Will handle it.

gaurav274: Done!

xzdandy (approved):

PR #1283, opened 2023-10-13
Fix current issues with forecasting

### LINKED ISSUE #1243: Enhance forecasting with more features

### LINKED ISSUE #1279: Better frequency inference and message for timeseries forecasting
### Search before asking

### Description

1. Better error message when frequency can not be inferred.  Instead of the following long error message:
```
10-12-2023 06:21:49 ERROR [plan_executor:plan_executor.py:execute_plan:0179] Can not infer the frequency for HomeSaleForecast. Please explicitly set it.
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/evadb/executor/plan_executor.py", line 175, in execute_plan
    yield from output
  File "/usr/local/lib/python3.10/dist-packages/evadb/executor/create_function_executor.py", line 526, in exec
    ) = self.handle_forecasting_function()
  File "/usr/local/lib/python3.10/dist-packages/evadb/executor/create_function_executor.py", line 254, in handle_forecasting_function
    raise RuntimeError(
RuntimeError: Can not infer the frequency for HomeSaleForecast. Please explicitly set it.
ERROR:evadb.utils.logging_manager:Can not infer the frequency for HomeSaleForecast. Please explicitly set it.
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/evadb/executor/plan_executor.py", line 175, in execute_plan
    yield from output
  File "/usr/local/lib/python3.10/dist-packages/evadb/executor/create_function_executor.py", line 526, in exec
    ) = self.handle_forecasting_function()
  File "/usr/local/lib/python3.10/dist-packages/evadb/executor/create_function_executor.py", line 254, in handle_forecasting_function
    raise RuntimeError(
RuntimeError: Can not infer the frequency for HomeSaleForecast. Please explicitly set it.
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
/usr/local/lib/python3.10/dist-packages/evadb/executor/plan_executor.py in execute_plan(self, do_not_raise_exceptions, do_not_print_exceptions)
    174             if output is not None:
--> 175                 yield from output
    176         except Exception as e:

6 frames
RuntimeError: Can not infer the frequency for HomeSaleForecast. Please explicitly set it.

During handling of the above exception, another exception occurred:

ExecutorError                             Traceback (most recent call last)
/usr/local/lib/python3.10/dist-packages/evadb/executor/plan_executor.py in execute_plan(self, do_not_raise_exceptions, do_not_print_exceptions)
    178                 if do_not_print_exceptions is False:
    179                     logger.exception(str(e))
--> 180                 raise ExecutorError(e)

ExecutorError: Can not infer the frequency for HomeSaleForecast. Please explicitly set it.
```
we shall return a simple response with `Can not infer the frequency for HomeSaleForecast. Please explicitly set it.`

2. Frequency has a non-negligible affect on the forecasting result. For example in the [Colab](https://colab.research.google.com/drive/10DBxSd80n-VfIoLsJxoZNo_yhrZNVfUn#scrollTo=hqffplIZjln6), when frequency is 'W', we have a flat prediction and frequency is 'M', we no longer have flat prediction. Given the auto frequency infer is not working, how should user choose the frequency? 

### Use case

Check [Colab Notebook](https://colab.research.google.com/drive/10DBxSd80n-VfIoLsJxoZNo_yhrZNVfUn#scrollTo=hqffplIZjln6) for the data and reproducing the error. 

americast:
This PR aims to solve the following issues:

- [x] Throwing error when non-numeric characters are in the data (partially fixes #1243)
- [x] Math domain error with `statsforecast`.
- [x] Fix GPU support for `neuralforecast`.
- ~Neuralforecast support for directly using batched data.~
- ~Auto frequency determination ( #1279).~

Will create separate PRs for the last two points.


xzdandy:
Hi @americast
1. We can have separate PRs instead of one PR for everything. 
2. For non-numeric characters are in the data, it is not only a problem for forecasting, for example, classification/predictions may also suffer this data cleaning problem. I think the alternative approach is to provide a REGEX function for generic uses instead of forecast only.  

Let me know what you think. You don't need to worry about the REGEX function, and can look into forecast-specific issue more.

americast:
Sure. Will create separate PRs for pt 2 and 4.

evadb/functions/forecast.py
```diff
@@ -38,16 +38,22 @@ def setup(
...
-        loaded_model = pickle.load(f)
-        f.close()
+        self.library = library
+        if "neuralforecast" in self.library:
+            from neuralforecast import NeuralForecast
+
+            loaded_model = NeuralForecast.load(path=model_path)
+            self.model_name = model_name[4:] if "Auto" in model_name else model_name
+        else:
+            f = open(model_path, "rb")
```
gaurav274, line 48: I would prefer using `contextManager`

evadb/executor/create_function_executor.py
```diff
@@ -163,6 +164,16 @@ def handle_sklearn_function(self):
             self.node.metadata,
         )
 
+    def convert_to_numeric(self, x):
+        x = re.sub("[^0-9.]", "", str(x))
```
xzdandy, line 168: This converts arbitrary string into numbers, which can hide real errors from user.

Also, notice the meaning `123,456` is affect by the localization.  Check https://stackoverflow.com/questions/6633523/how-can-i-convert-a-string-with-dot-and-comma-into-a-float-in-python for details. 

My suggestion is:

a) Do nothing. Just throw the error message, forecast expects a numeric column but gets a x type column. To be more advanced, we can suggest some data cleaning method. 
b) Only handling the case that locale supports. For the rest, fallback to throwing error message.

evadb/executor/create_function_executor.py
```diff
@@ -422,7 +457,10 @@ def handle_forecasting_function(self):
             FunctionMetadataCatalogEntry("library", library),
         ]
 
-        os.environ.pop("CUDA_VISIBLE_DEVICES", None)
+        if save_old_cuda_env is not None:
+            os.environ["CUDA_VISIBLE_DEVICES"] = save_old_cuda_env
+        else:
+            os.environ.pop("CUDA_VISIBLE_DEVICES", None)
```
xzdandy, line 463: Please check https://stackoverflow.com/questions/2059482/temporarily-modify-the-current-processs-environment for using context manager to modify the environment variable.

evadb/executor/create_function_executor.py
```diff
@@ -394,13 +416,26 @@ def handle_forecasting_function(self):
...
                 model.fit(df=data, val_size=horizon)
+                model.save(model_path, overwrite=True)
             else:
+                # The following lines of code helps eliminate the math error encountered in statsforecast when only one datapoint is available in a time series
+                for col in data["unique_id"].unique():
+                    if len(data[data["unique_id"] == col]) == 1:
+                        data = data._append(
+                            [data[data["unique_id"] == col]], ignore_index=True
+                        )
+
```
xzdandy, line 434: We need to be explicit. If some data points are discarded, the user needs to be aware.  Can we print out some warning information in this case?

evadb/executor/create_function_executor.py
americast: Thanks for pointing this out! The issue here is any a times there are dangling strings that might needed to be removed automatically. Like units or currency symbols. I'll look into fixing this via method b.

evadb/executor/create_function_executor.py
americast: We are not discarding any data point here. The math domain error occurs when there is only a single data point; this code helps replicate that value instead of throwing the error. Technically, it is not possible to forecast with just one data point.

americast:
@xzdandy Resolved all the concerns

evadb/executor/create_function_executor.py
```diff
@@ -245,253 +282,280 @@ def handle_ultralytics_function(self):
...
-        except Exception as e:
-            err_msg = f"{str(e)}. HORIZON must be integral."
-            logger.error(err_msg)
-            raise FunctionIODefinitionError(err_msg)
+        cuda_devices_here = "0"
+        if "CUDA_VISIBLE_DEVICES" in os.environ:
+            cuda_devices_here = os.environ["CUDA_VISIBLE_DEVICES"].split(",")[0]
+
+        with set_env(CUDA_VISIBLE_DEVICES=cuda_devices_here):
+            aggregated_batch_list = []
```
xzdandy, line 290: If I understand correctly, only the neueralforcast fit needs to change the `CUDA_VISISBLE_DEVICES`.  Can we move the `set_env` only around the neuralforcast's `model.fit`? Thanks!

xzdandy:
I am fixing the Doc.

xzdandy (approved):
I will update the tutorial notebook to showcase these two features.

americast:
> I will update the tutorial notebook to showcase these two features.

Thanks a lot @xzdandy! Let me know once you are done with that add I will add some more comparisons to it.

PR #1289, opened 2023-10-15
SnowFlake Integration for EvaDB

kslohith:
Please suggest if this feature needs more test cases


test/third_party_tests/test_native_executor.py
```diff
@@ -228,6 +228,26 @@ def test_should_run_query_in_clickhouse(self):
...
+    def test_should_run_query_in_snowflake(self):
+        # Create database.
+        params = {
+            "user": "eva",
+            "password": "password",
+            "account": "account_number",
+            "database": "EVADB",
+            "schema": "SAMPLE_DATA",
+            "warehouse": "warehouse",
+        }
```
xzdandy, line 240: Is this a valid connection?

kslohith: No, it is not, do we need to test it against a real snowflake account ?

xzdandy: Yes. Otherwise the test will not run. Does snowflake has any free account we can test with?

xzdandy: I see. Then we can skip this test. Use `@pytest.mark.skip(reason="...")`

xzdandy:
Please also update the documentation page: https://evadb.readthedocs.io/en/stable/source/reference/databases/index.html Thanks!

xzdandy (approved):

PR #1290, opened 2023-10-16
fix: text_summarization uses drop udf

aryan-rajoria:
text_summarization uses drop udf instead of drop function.


xzdandy (approved):

xzdandy:
Thanks for catching it!

PR #1291, opened 2023-10-16
REST apis for EvaQL

ethanyang2000:


jarulraj:
@rohithmulumudy @imnidhi Do we need these additional APIs?

jarulraj:
This is probably not needed right now.

Issue #1282, opened 2023-10-13
Table with column name starting with `_` fails to be created, followed by failure to drop or create the table again


americast:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar bug report.

### Bug

I get an error if I run a `CREATE TABLE` query that has a column name starting with `_`:

```sql
CREATE TABLE hello (
        _row_id INTEGER,
        price TEXT,
        beds TEXT,
        bath TEXT,
        area TEXT,
        address TEXT);
```

This is the error I get:
```bash
ExecutorError: (sqlite3.IntegrityError) UNIQUE constraint failed: column_catalog.name, column_catalog.table_id
[SQL: INSERT INTO column_catalog (name, type, is_nullable, array_type, array_dimensions, table_id) VALUES (?, ?, ?, ?, ?, ?)]
[parameters: ('_row_id', 'INTEGER', 1, None, '()', 1)]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
```
However, strangely, after this, I am unable to either drop nor create the table `hello` again.

```sql
CREATE TABLE hello (
        rowid INTEGER,
        price TEXT,
        beds TEXT,
        bath TEXT,
        area TEXT,
        address TEXT);
```
leads to `ExecutorError: Table: hello already exists`. And,

```sql
DROP TABLE IF EXISTS hello;
```

leads to `ExecutorError: Failed to drop the table hello with Exception No table found with name hello`.

The only way to recover from here seems to be to delete the `evadb_data` directory.

### Environment

Latest staging on Ubuntu 20.04

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

gaurav274:
@americast `_row_id ` is a reserved keyword, and users cannot create a table with this column name.

gaurav274:
I can reproduce the other error. Will fix it.

PR #1293, opened 2023-10-17
fix: make the table/function catalog insert operation atomic

gaurav274:
Fixes: #1282

### DIFF
--- a/evadb/catalog/services/table_catalog_service.py
+++ b/evadb/catalog/services/table_catalog_service.py
@@ -57,8 +57,20 @@ class TableCatalogService(BaseService):
             # populate the table_id for all the columns
             for column in column_list:
                 column.table_id = table_catalog_obj._row_id
-            column_list = self._column_service.insert_entries(column_list)
+            column_list = self._column_service.create_entries(column_list)
 
+            # atomic operation for adding table and its corresponding columns.
+            try:
+                self.session.add_all(column_list)
+                self.session.commit()
+            except Exception as e:
+                self.session.rollback()
+                self.session.delete(table_catalog_obj)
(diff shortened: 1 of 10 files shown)


evadb/catalog/services/table_catalog_service.py
```diff
@@ -57,8 +57,20 @@ def insert_entry(
...
+            # atomic operation for adding table and its corresponding columns.
+            try:
+                self.session.add_all(column_list)
+                self.session.commit()
+            except Exception as e:
+                self.session.rollback()
+                self.session.delete(table_catalog_obj)
+                self.session.commit()
+                logger.exception(
+                    f"Failed to insert entry into table catalog with exception {str(e)}"
```
americast, line 71: Is this error supposed to be shown when a table with `_row_id` is column is created?
I still get the same error as earlier:
```
ExecutorError: (sqlite3.IntegrityError) UNIQUE constraint failed: column_catalog.name, column_catalog.table_id                                        
[SQL: INSERT INTO column_catalog (name, type, is_nullable, array_type, array_dimensions, table_id) VALUES (?, ?, ?, ?, ?, ?) RETURNING _row_id]       
[parameters: ('_row_id', 'INTEGER', 1, None, '()', 5)]                                                                                                
(Background on this error at: https://sqlalche.me/e/20/gkpj) 
```

americast (approved):
Thanks! `DROP TABLE` works fine now after creating table fails.

evadb/catalog/services/table_catalog_service.py
gaurav274: I see. I'll raise an error if the user specifies such a column name.

PR #1294, opened 2023-10-17
fix: improve testcase

gaurav274:
1. Test default values of `chunk_size` and `chunk_overlap`


xzdandy (approved):
I think for all default values. It is better to define in the constant.py so we have a centralized place for them.

gaurav274:
> I think for all default values. It is better to define in the constant.py so we have a centralized place for them.

Looks like I made a mistake. The default values are something else. Will fix it!

PR #1298, opened 2023-10-17
Convert nested join in Vector Queries to Pandas Merge. 

Chitti-Ankith:
Profiling on Vector Scan showed that we are spending a lot of time in the post-processing logic doing a Nested Join. This is an initial commit to change that into a Join using Pandas. Change showed ~50% improvement in Similarity Queries.

### DIFF
--- a/evadb/executor/vector_index_scan_executor.py
+++ b/evadb/executor/vector_index_scan_executor.py
@@ -139,7 +139,6 @@ class VectorIndexScanExecutor(AbstractExecutor):
                 row_dict = row.to_dict()
                 res_data_list.append(row_dict)
 
-
         result_df = pd.DataFrame(res_data_list)
 
         final_df = pd.merge(
@@ -147,7 +146,7 @@ class VectorIndexScanExecutor(AbstractExecutor):
             result_df,
             left_on="row_num_np",
             right_on=row_num_col_name,
-            how="inner"
+            how="inner",
         )
 
         if "row_num_np" in final_df:
--- a/evadb/executor/vector_index_scan_executor.py
+++ b/evadb/executor/vector_index_scan_executor.py
@@ -124,22 +124,31 @@ class VectorIndexScanExecutor(AbstractExecutor):
             )
 
         final_df = pd.DataFrame()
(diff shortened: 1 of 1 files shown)


jiashenC:
For 20% speedup, how many rows does the table contain?

evadb/executor/vector_index_scan_executor.py
```diff
@@ -124,20 +123,24 @@ def _evadb_vector_index_scan(self, *args, **kwargs):
...
 
-            # Nested join.
-            for _, row in batch.frames.iterrows():
-                for idx, row_num in enumerate(row_num_np):
-                    if row_num == row[row_num_col_name]:
-                        res_row = dict()
-                        for col_name in column_list:
-                            res_row[col_name] = row[col_name]
-                        res_row_list[idx] = res_row
+            result_df = pd.merge(
```
jiashenC, line 134: Instead of doing O(n) of merging, will we get better performance if get all batches from the child and do merging only once?

Chitti-Ankith: > Instead of doing O(n) of merging, will we get better performance if get all batches from the child and do merging only once?

Thanks for the suggestion, I have also made changes to not add child frames into the result df before merging to avoid unnecessary processing. The speedup is 2X now.

evadb/executor/vector_index_scan_executor.py
```diff
@@ -124,20 +123,36 @@ def _evadb_vector_index_scan(self, *args, **kwargs):
...
+        result_df = result_df.reindex(row_num_np)
+        row_num_df.set_index(pd.Index(row_num_np), inplace=True)
+
+        final_df = pd.merge(
+            row_num_df,
+            result_df,
+            left_index=True,
+            right_index=True,
+            how="left",
+            # sort=False
```
jiashenC, line 153: Just remove this?

jiashenC (approved):

evadb/executor/vector_index_scan_executor.py
Chitti-Ankith: Done

PR #1305, opened 2023-10-19
Starting the changes for XGBoost classification integration.

jineetd:
We shall add XGBoost classification support in EVADB.


xzdandy (approved):
Thanks! Looks great.

xzdandy:
Hi, could you merge the staging? I think it is good opportunity for us also to fix the congratulation error (caused by the refactor in #1272 ) for xgboost . Thanks!

xzdandy:
The model training test take too much time on circle ci. We skip the old test and only keep the ones that have been recently changed.

PR #1306, opened 2023-10-19
Cost-based Batch Size 

xzdandy:
To start with, we follow a simple cost-based rule:
1. Use the maximum batch size from storage engine when there is no limit
2. Use the minimum (1) batch size from storage engine when there is limit.


gaurav274:
Couldn't understand how the limit condition is enforced in the optimizer.

PR #1308, opened 2023-10-20
Job scheduler changes

dungnmaster:
This PR adds support for creating and dropping jobs in evadb based on this: https://github.com/georgia-tech-db/evadb/issues/1248


xzdandy:
Hi @dungnmaster, thanks for looking into Job scheduler. What is the status of the PR? Could you summarize the design which helps us understanding the implementation. Thanks!

dungnmaster:
> Hi @dungnmaster, thanks for looking into Job scheduler. What is the status of the PR? Could you summarize the design which helps us understanding the implementation. Thanks!

- In the current implementation, the job scheduler process can be started using the start_jobs() method of the EvaDbConnection object and stopped using the stop_jobs() method.
- The start_jobs() method, starts a new Python process that keeps polling for valid executable tasks in the catalog and also executes them in the same process. This will continue until the job scheduler is stopped or the calling python parent process terminates.

evadb/utils/job_scheduler.py
```diff
@@ -0,0 +1,120 @@
...
+
+                    # update the next trigger time for this job
+                    self._update_next_schedule_run(next_executable_job)
+
+                    # update the previosly inserted job history record with endtime
+                    self._evadb.catalog().update_job_history_end_time(
+                        next_executable_job.row_id,
+                        execution_time,
+                        datetime.datetime.now(),
+                    )
```
xzdandy, line 99: We can update the result of job in the catalog. so it is queryable. We can do this in later PRs. Maybe add a TODO.

xzdandy (approved):
Great work! Very good quality. Could you also update the documentation https://evadb.readthedocs.io/en/stable/source/reference/evaql.html for the `CREATE JOB` query? Thanks!

PR #1310, opened 2023-10-22
Add documentation note for connecting to existing SQL DB

Chitti-Ankith:
I have seen students getting confused with the documentation wherein they try to always create a connection to some database (Eg : Postgres) before trying to insert or query data. Adding a note in the documentation to help clear this point up as well as describe that Eva has native storage support for SQLite.

### DIFF
--- a/docs/source/overview/connect-to-data-sources.rst
+++ b/docs/source/overview/connect-to-data-sources.rst
@@ -6,10 +6,6 @@ EvaDB supports a wide range of data sources including SQL database systems, obje
 Connect to an Existing SQL Database System
 ------------------------------------------
 
-.. note::
-
-   For prototyping, users can opt to establish a connection to :ref:`SQLite<sqlite>`.
-
 1. Use the :ref:`CREATE DATABASE<create-database>` statement to connect to an **existing** SQL database.
 
 .. code-block::
@@ -27,6 +23,10 @@ Connect to an Existing SQL Database System
 .. note::
 
    Go over the :ref:`CREATE DATABASE<create-database>` statement for more details. The :ref:`Databases<databases>` page lists all the database systems that EvaDB currently supports.
+
+.. note::
+
+   For prototyping, users can opt to establish a connection to :ref:`SQLite<sqlite>`.
 
 2. Preview the data using ``SELECT``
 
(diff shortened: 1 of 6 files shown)


docs/source/overview/connect-to-data-sources.rst
```diff
@@ -6,6 +6,12 @@ EvaDB supports a wide range of data sources including SQL database systems, obje
 Connect to an Existing SQL Database System
 ------------------------------------------
 
+.. note::
+
+   Connecting to an existing SQL database is unnecessary for inserting or querying data in Eva. This step is only required if
+   users intend to work with data already present in some SQL DB. Eva has native storage support for SQLite, which can be utilized without
+   establishing any connection. Refer to :ref:`EvaQL<evaql>` page for more details.
```
xzdandy, line 13: Hi, I think the idea is that we want the user to use the third party data sources instead of the native storage support for SQLite.  Notice the third party data source also has SQLite support, which does not require an existing SQL database. We have also updated the documentation page of each third party data source to indicate whether it requires an already running database instance. This [page](https://evadb.readthedocs.io/en/stable/source/reference/databases/postgres.html) is an example for  PostgreSQL.

jarulraj:
We could maybe clarify that for prototyping, `sqlite` is enough.

docs/source/overview/connect-to-data-sources.rst
xzdandy: Thanks @Chitti-Ankith for the feedback! We can add a note in [Connect to an Existing SQL Database System](https://evadb.readthedocs.io/en/stable/source/overview/connect-to-data-sources.html) that says `For local prototyping, we can establish a connection to [sqlite](https://evadb.readthedocs.io/en/stable/source/reference/databases/sqlite.html)`. I think the key is that we want the user to establish a connection (e.g., to SQLite) instead of using the native storage support.

Chitti-Ankith: Done

jarulraj:
@Chitti-Ankith For reference, I added the shared information about prototyping to the `shared` rest file.

Chitti-Ankith:
> @Chitti-Ankith For future reference, I added the shared information about prototyping to the `shared` rest file.

Got it, thanks @jarulraj !

PR #1313, opened 2023-10-23
feat: Llm operator

gaurav274:


evadb/executor/llm_executor.py
```diff
@@ -0,0 +1,36 @@
...
+class LLMExecutor(AbstractExecutor):
+    def __init__(self, db: EvaDBDatabase, node: LLMPlan):
+        super().__init__(db, node)
+        self.llm_expr = node.llm_expr
+        self.alias = node.alias
+
+    def exec(self, *args, **kwargs) -> Iterator[Batch]:
+        child_executor = self.children[0]
+        for batch in child_executor.exec(**kwargs):
+            llm_result = self.llm_expr.evaluate(batch)
```
xzdandy, line 32: Is the batch optimization done in the LLMExecutor and will be added in future PRs?

evadb/expression/expression_utils.py
```diff
@@ -296,3 +298,20 @@ def _has_simple_expressions(expr):
...
+        return True
+    return False
+
+
+def extract_llm_expressions_from_project(exprs: List[AbstractExpression]):
+    remaining_exprs = []
+    llm_exprs = []
+    for expr in exprs:
+        if is_llm_expression(expr):
+            llm_exprs.append(expr.copy())
```
xzdandy, line 314: Add a note here, the chained function call will not work here. For example `STRTODATAFRAME(LLM('EXTRACT SOME COLUMN', data))`

evadb/optimizer/statement_to_opr_converter.py
```diff
@@ -228,9 +230,35 @@ def _visit_union(self, target, all):
...
+        if llm_nodes:
+            # add existing plan as a child of llm
+            plan_root = LogicalLLM(llm_exprs[0])
+            plan_root.append_child(self._plan)
+            for expr in llm_exprs[1:]:
+                new_root = LogicalLLM(expr)
+                new_root.append_child(plan_root)
+                plan_root = new_root
+            self._plan = plan_root
+
```
xzdandy, line 257: IMO, the generic way is to do it in optimizer with apply and merge. What will the plan looks like, if we have `SELECT id, LLM(...) FROM some_table;`

evadb/executor/llm_executor.py
gaurav274: Yes

evadb/expression/expression_utils.py
gaurav274: Yes, I'll add that in next PR

evadb/optimizer/statement_to_opr_converter.py
gaurav274: It will be `Project(id, llm.response) -> LLMExec() -> Get`

evadb/functions/llms/openai.py
```diff
@@ -0,0 +1,113 @@
...
+
+        validate_kwargs(kwargs, allowed_keys=_DEFAULT_PARAMS.keys())
+        self.model_params = {**_DEFAULT_PARAMS, **kwargs}
+
+        self.model_name = self.model_params["model"]
+
+    def generate(self, prompts: List[str]) -> List[str]:
+        import openai
+
+        @retry(tries=6, delay=20)
```
pchunduri6, line 68: It might be a good time to also add logging to the retry logic - 
https://tenacity.readthedocs.io/en/latest/#before-and-after-retry-and-logging

This will log the retry attempts in our logger so the user knows when rate-limiting errors occur. I found this helpful when waiting for a long time. The downside is that we must add the tenacity library to the requirements.

evadb/functions/llms/openai.py
```diff
@@ -0,0 +1,113 @@
...
+            answer = response.choices[0].message.content
+            results.append(answer)
+
+        return results
+
+    def get_cost(self, prompt: str, response: str):
+        try_to_import_tiktoken()
+        import tiktoken
+
+        encoding = tiktoken.encoding_for_model(self.model_name)
```
pchunduri6, line 100: If we already have the response, we can directly compute the cost using the `response["usage"]` parameter?
Tiktoken would be good to estimate the cost before executing the query (helpful for query opt). Maybe we can have two functions, estimate_cost and get_cost. Estimating cost is not simple, though, because we do not know the completion tokens apriori. We would then need a heuristic for the estimated completion tokens.

evadb/constants.py
```diff
@@ -21,3 +21,4 @@
 IFRAMES = "IFRAMES"
 AUDIORATE = "AUDIORATE"
 DEFAULT_FUNCTION_EXPRESSION_COST = 100
+LLM_FUNCTIONS = ["chatgpt", "completion"]
```
pchunduri6, line 24: Are we not adding the LLM operator to the parser? So, the allowed LLM names are restricted here?

evadb/functions/llms/openai.py
gaurav274: Yes, a valid concern. I was also thinking about it.

PR #1327, opened 2023-10-31
Add the validation score and training time for create_function in XGBoost

jineetd:
Let us show the validation score and training time for the XGBoost AutoML model trained. This shall give us fair enough idea on how the model trained on the training data set.

### DIFF
--- a/evadb/executor/create_function_executor.py
+++ b/evadb/executor/create_function_executor.py
@@ -687,8 +687,8 @@ class CreateFunctionExecutor(AbstractExecutor):
                 pd.DataFrame(
                     [
                         msg,
-                        "Validation Score: " + best_score,
-                        "Training time: " + train_time,
+                        "Validation Score: " + str(best_score),
+                        "Training time: " + str(train_time),
                     ]
                 )
             )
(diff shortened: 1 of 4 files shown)


evadb/executor/create_function_executor.py
```diff
@@ -674,7 +682,10 @@ def exec(self, *args, **kwargs):
             msg = f"Function {self.node.name} overwritten."
         else:
             msg = f"Function {self.node.name} added to the database."
-        yield Batch(pd.DataFrame([msg]))
+        if best_score and train_time:
+            yield Batch(pd.DataFrame([msg, best_score, train_time]))
```
xzdandy, line 686: Could we add some text? Other wise, it will be just two numbers without any explanation.

xzdandy:
Could you help fix `evadb/binder/function_expression_binder.py:115:21: F841 local variable 'OpenAI_key' is assigned to but never used`? It is not relevant to this PR. Change the 'OpenAI_key' to `openai_key` in evadb/binder/function_expression_binder.py. Thanks!

evadb/executor/create_function_executor.py
jineetd: Added some annotations

xzdandy (approved):

PR #1334, opened 2023-11-02
logging an error message for invalid files while loading

rohithmulumudy:
Issue - [721](https://github.com/georgia-tech-db/evadb/issues/721)

Currently, we abort the entire process when the load executor encounters a corrupted file.

### DIFF
--- a/test/integration_tests/long/test_load_executor.py
+++ b/test/integration_tests/long/test_load_executor.py
@@ -206,17 +206,16 @@ class LoadExecutorTests(unittest.TestCase):
                 self.evadb, "SELECT name FROM MyVideos", do_not_print_exceptions=True
             )
 
-    def test_should_rollback_if_video_load_fails(self):
+    def test_should_rollback_or_skip_if_video_load_fails(self):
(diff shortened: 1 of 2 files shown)


test/integration_tests/long/test_load_executor.py
```diff
@@ -231,24 +230,32 @@ def test_should_rollback_if_video_load_fails(self):
                         do_not_print_exceptions=True,
                     )
 
+            # Load one correct file and one empty file
+            # one file should get added
+            with tempfile.TemporaryDirectory() as tmp_dir:
+                shutil.copy2(str(valid_videos[0]), tmp_dir)
+                shutil.copy2(str(empty_file.name), tmp_dir)
```
xzdandy, line 237: Could I get a clarification why we skip the empty file?

rohithmulumudy: Currently, we are aborting the entire process when the load executor encounters a corrupted file. Post my changes, all the valid files will get loaded, and we are gonna log the corrupted files and ignore them

xzdandy (approved):

PR #1336, opened 2023-11-02
Add test_eva_db to gitignore

LordDarkula:
The environment created in the setup instructions in the documentation calls the environment `test_eva_db`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,6 +101,7 @@ env.bak/
 venv.bak/
 env38/
 env_eva/
+test_eva_db/
 
 # Spyder project settings
 .spyderproject


hershd23:
Is this related to the fact that you have created an env and you want to gitignore that env?

If so can you use one of the env names already present in the docker file? Examples being `env`, `venv` and `env_eva`

LordDarkula:
> Is this related to the fact that you have created an env and you want to gitignore that env?
> 
> If so can you use one of the env names already present in the docker file? Examples being `env`, `venv` and `env_eva`

The environment created in the setup instructions in the documentation calls the environment test_eva_db

hershd23:
Hmm fair enough, in that case it makes sense

@gaurav274 @kaushikravichandran for approval

hershd23 (approved):

PR #1337, opened 2023-11-02
CREATE INDEX IF NOT EXISTS is broken.

Chitti-Ankith:
This PR fixes an issue in CREATE INDEX IF NOT EXISTS command wherein if 'IF NOT EXISTS' is passed, we had an unreferenced variable issue. Added Unit Tests to check the correctness of both the cases. 

Also reverted the index changes while merging dataframes after vector scan, as it's failing for some cases where indexes can be undefined.


xzdandy (approved):
Thanks!

Issue #1235, opened 2023-09-29
Introduce `EXTRACT_COLUMNS` to extract structured tables from unstructured text


xzdandy:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar feature requests.

### Description

`EXTRACT_COLUMNS`  will be similar to `EXTRACT_OBJECT` for videos, which is not a standard user defined functions. In optimizer, it will be translated to a valid EvaDB query plan tree with multiple functions and operators.

**Example Usage**

```
EXTRACT_COLUMNS(
    "gpt-3.5-turbo", 
    "faiss",
    [
        ["name", TEXT, "name of the user profile", "logicx"], 
        ["country", TEXT, "country the user comes from", "United States"]
    ], 
     input_relationship
)
```

- The first argument specifies the llm model to use
- The second argument specifies the vector database to use. If the second column is `""`, then RAG will not be used. In the first release of `EXTRACT_COLUMNS`, we will not support RAG.
- The third augments specifies the column we want to extract, for every column, we specify 
    - the name of column
    - type of column
    - natural language to describe how to extract that column
    - an example value. 
- The fourth is the input_relationship
- The output returns a batched panda dataframe that contains those extracted columns. This is a one-to-one mapping for the input_relationship.

If we want to provide more fined grained controls, for example tuning hyper paramters, we can also introduce a `CREATE FUNCTION`, which allows us to have a key-value based configuration. 

@gaurav274 @jiashenC Please provide feedback. Thanks. 

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

pchunduri6:
>     [
>         ["name", "name of the user profile", "logicx"], 
>         ["country", "country the user comes from", "United States"]
>     ], 

1. How would this translate to the LLM prompt in the background -- e.g., one prompt for each column, single prompt by combining all columns
2. The LLM extraction is brittle, so careful prompt engineering is required. Is it safe to use this structure without providing the option to engineer the prompt?
3. With RAG queries, there is information loss, so accurate extraction will get trickier. Tracking the output accuracy could be challenging.

xzdandy:
Hi @pchunduri6, very good feedback.

1) Backend will be translated to what we have similar in the stargazers. Optimization like batching will be applied accordingly. There are also new opportunities like merging, for example, some columns are extracted in predicate while some are extracted in projection. 
```
[...]

Issue #11, opened 2018-12-12
UDF - Speed Detector


jaehobang:
We can have a object tracker to determine speed

PR #1338, opened 2023-11-03
[WIP] Unstructured data to structured data conversion

hershd23:
Added custom function for extracting columns from unstructured data
	new file:   ../evadb/functions/extract_columns.py

### DIFF
--- a/evadb/functions/extract_columns.py
+++ b/evadb/functions/extract_columns.py
@@ -1,0 +1,166 @@
+# coding=utf-8
+# Copyright 2018-2023 EvaDB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from io import BytesIO
+
+import numpy as np
+import pandas as pd
+import json
+from retry import retry
+
+from evadb.catalog.catalog_type import NdArrayType
+from evadb.functions.abstract.abstract_function import AbstractFunction
+from evadb.functions.decorators.decorators import forward
+from evadb.functions.decorators.io_descriptors.data_types import PandasDataframe
+from evadb.utils.generic_utils import try_to_import_openai
+from evadb.utils.logging_manager import logger
+
+
+class ExtractColumnsFunction(AbstractFunction):
+    @property
+    def name(self) -> str:
+        return "EXTRACT_COLUMNS"
+
+    def setup(
+        self, 
+        model="gpt-3.5-turbo",
+        temperature: float = 0,
+        openai_api_key=""
+    ) -> None:
+        self.model = model
+        self.temperature = temperature
+        self.openai_api_key = openai_api_key
+
+    @forward(
+        input_signatures=[
+            PandasDataframe(
+                columns=["input_rows"],
+                column_types=[
+                    NdArrayType.STR,
+                ],
+                column_shapes=[(1,)],
+            )
+        ],
+        output_signatures=[
+            PandasDataframe(
+                columns=["response"],
+                column_types=[
+                    NdArrayType.STR,
+                ],
+                column_shapes=[(1,)],
+            )
+        ],
+    )
+    def forward(self, unstructured_df):
+        """
+        NOTE (QUESTION) : Can we structure the inputs and outputs better
+        The circumvent issues surrounding the input being only one pandas dataframe and output columns being predefined 
+        Will add all column types as a JSON and parse in the forward function
+        Provide only the file name from which the input will be read
+        Output in JSON which can be serialized and stored in the results column of the DF
+        """
+        
+        try_to_import_openai()
+        import openai
+        
+        @retry(tries=6, delay=20)
+        def completion_with_backoff(**kwargs):
+            return openai.ChatCompletion.create(**kwargs)
+
+        openai.api_key = self.openai_api_key
+        # If not found, try OS Environment Variable
+        if len(openai.api_key) == 0:
+            openai.api_key = os.environ.get("OPENAI_API_KEY", "")
+        assert (
+            len(openai.api_key) != 0
+        ), "Please set your OpenAI API key using SET OPENAI_API_KEY = 'sk-'  or environment variable (OPENAI_API_KEY)"
+
+        def generate_structured_data(unstructured_df: PandasDataframe):
+            results = []
+            #column_types = json.loads(unstructured_df[unstructured_df.columns[0]])
+            input_rows = unstructured_df[unstructured_df.columns[0]]
+            
+            column_types_dict = {
+                "columns":
+                [
+                    {
+                        "name": "Issue Category",
+                        "description": "The category of the issue",
+                        "type": "One of (hardware, software)"
+                    },
+                    {
+                        "name": "Raw Issue String",
+                        "description": "The raw issue string containing the exact input given by the user",
+                        "type": "string"
+                    },
+                    {
+                        "name": "Issue Component",
+                        "description": "The component that is causing the issue",
+                        "type": "string"
+                    },
+                ]
+            }
+            
+            column_types = json.dumps(column_types_dict)
+            
+            base_prompt = """
+            You are given a user query. Your task is to extract the following fields from the query and return the result in json format.\n
+            """
+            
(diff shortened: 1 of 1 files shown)


hershd23:
@xzdandy I created a python notebook as well but it gets gitignored while the rest of the tutorial notebooks don't any idea?

hershd23:
Solves #1235

evadb/functions/extract_columns.py
```diff
@@ -0,0 +1,166 @@
...
+        if len(openai.api_key) == 0:
+            openai.api_key = os.environ.get("OPENAI_API_KEY", "")
+        assert (
+            len(openai.api_key) != 0
+        ), "Please set your OpenAI API key using SET OPENAI_API_KEY = 'sk-'  or environment variable (OPENAI_API_KEY)"
+
+        def generate_structured_data(unstructured_df: PandasDataframe):
+            results = []
+            #column_types = json.loads(unstructured_df[unstructured_df.columns[0]])
+            input_rows = unstructured_df[unstructured_df.columns[0]]
```
xzdandy, line 94: Let 's start with one column a time for simplicty. 
```
SELECT issue_category, raw_issue_string, issue_compent FROM unstructured_data
JOIN LATERAL  EXTRACT_COLUMN("The category of the issue", "One of (hardware, software)", input rows) AS T(issue_category)
JOIN LATERAL  EXTRACT_COLUMN("The raw issue string containing the exact input given by the user", "string", input rows) AS T(raw_issue_string)
JOIN LATERAL  EXTRACT_COLUMN("The component that is causing the issue", "string", input rows) AS T(issue_compent);
```
- To keep everything simple and have a basic version. Let's do not worry about the type for now. Always returning string. 
- We will support direct `AS` in the projection. So the query can be rewritten without `JOIN LATERAL`. 
- We can later support `EXTRACT_COLUMNS()` if there is a need.
- I think we need a more well-defined language to specify the column constraints. Like `[ 'Enum(hardware, software)', Type(String), Example(hardware), Max(100), Min(0)]`. But we do not need to worry about it now.

evadb/functions/extract_columns.py
```diff
@@ -0,0 +1,166 @@
...
+        Will add all column types as a JSON and parse in the forward function
+        Provide only the file name from which the input will be read
+        Output in JSON which can be serialized and stored in the results column of the DF
+        """
+        
+        try_to_import_openai()
+        import openai
+        
+        @retry(tries=6, delay=20)
+        def completion_with_backoff(**kwargs):
```
xzdandy, line 80: I think it is better to base our implementation on the https://github.com/georgia-tech-db/evadb/blob/staging/evadb/functions/chatgpt.py instead of duplicating it. So if we make optimizations to the chatgpt call, we do not need to duplicate it in the `extract_column`.

hershd23: Not sure how we do that?

hershd23: Do we import the chatgpt function and call it from here?

evadb/functions/extract_columns.py
```diff
@@ -0,0 +1,166 @@
...
+
+        openai.api_key = self.openai_api_key
+        # If not found, try OS Environment Variable
+        if len(openai.api_key) == 0:
+            openai.api_key = os.environ.get("OPENAI_API_KEY", "")
+        assert (
+            len(openai.api_key) != 0
+        ), "Please set your OpenAI API key using SET OPENAI_API_KEY = 'sk-'  or environment variable (OPENAI_API_KEY)"
+
+        def generate_structured_data(unstructured_df: PandasDataframe):
```
hershd23: @pchunduri6 this is what I am talking about

hershd23:
@xzdandy @pchunduri6 moved to a "one-column-at-a-time" implementation as you recommended.

The notebook has the implementation

hershd23:
For one column at a time I think this PR is ready for review @xzdandy @pchunduri6.

For the other changes discussed with either of you, I think it makes sense to take that up in a separate PR else this will bloat. Let me know what you think

evadb/functions/extract_column.py
```diff
@@ -0,0 +1,101 @@
...
+        """.format(
+            field_name, description, data_type
+        )
+
+        print(prompt)
+        print(content)
+
+        output_df = pd.DataFrame({"response": []})
+
+        for row in input_rows:
```
xzdandy, line 92: We shall not use the first field_name, description, data_type  for all the rest rows. 
```python
for row in unstructured_df.itertuples():
    field_name = row[0]
    description = row[1]
    data_type = row[2]
    input_rows = row[3]

    ...
```

tutorials/20-structured-data.ipynb
```diff
@@ -0,0 +1,208 @@
+{
```
xzdandy, line 1: <pre class="ql-syntax" spellcheck="false">Line #11.    print(table)
</pre><p>Remove</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/georgia-tech-db/evadb/pull/1338/discussion/#comment-1408957380'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='4'/>

xzdandy, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    print(table)
</pre><p>Use $table instead of print(table) for better output.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/georgia-tech-db/evadb/pull/1338/discussion/#comment-1408957390'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='6'/>

xzdandy, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    for _, row in table.iterrows():
</pre><p>Similarity directly use table for the output. Please also remove print in the function to avoid confusions. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/georgia-tech-db/evadb/pull/1338/discussion/#comment-1408957397'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='7'/>

xzdandy:
Can we also add a long integration test for the function under https://github.com/georgia-tech-db/evadb/tree/staging/test/integration_tests/long/functions? We can skip the test in circle ci due to openai key, but I think it is good to have one. 

It can either be end-to-end (i.e., SQL queries) or directly test the function class.

hershd23:
Yes @xzdandy on it

evadb/functions/extract_columns.py
hershd23: Didn't get this

hershd23:
Also this is failing the linter check for a Colab Notebook. Can you point me towards information on how to add that

tutorials/20-structured-data.ipynb
xzdandy, line 1: We need to skip the notebook test at https://github.com/georgia-tech-db/evadb/blob/c2457b2a0a6153682d398ac07ea6f18656f8d7ba/script/test/test.sh#L97 due to open ai key

xzdandy (changes requested):

xzdandy:
> Also this is failing the linter check for a Colab Notebook. Can you point me towards information on how to add that

Remove the last empty cell.

hershd23:
```
12-01-2023 17:31:12 [check_notebook_format:295] ERROR: ERROR: Notebook /Users/hershdhillon23/projects/evadb/script/formatting/../../tutorials/20-structured-data.ipynb does not contain correct Colab link -- update the link.
```

Do not have a collar link right now

xzdandy:
> ```
> 12-01-2023 17:31:12 [check_notebook_format:295] ERROR: ERROR: Notebook /Users/hershdhillon23/projects/evadb/script/formatting/../../tutorials/20-structured-data.ipynb does not contain correct Colab link -- update the link.
> ```
> 
> Do not have a collar link right now

The current notebook actually does not work on the colab. I was trying to make it work yesterday and I think it needs several modifications. One fix can help is that can you add the `EXTRACT_COLUMN` to bootstrap functions in https://github.com/georgia-tech-db/evadb/blob/staging/evadb/functions/function_bootstrap_queries.py

pchunduri6:
Should we perform this operation using ChatGPT directly or use something like pandasAI to write a function using LLM and then extract the column we need? Writing a function is much cheaper token cost-wise, but less robust.
@hershd23 @xzdandy Any thoughts?

xzdandy:
> Should we perform this operation using ChatGPT directly or use something like pandasAI to write a function using LLM and then extract the column we need? Writing a function is much cheaper token cost-wise, but less robust.
> @hershd23 @xzdandy Any thoughts?

Hi @pchunduri6, I think it depends on the task. If the extract column is based on patterns, I think we can generate regex for saving the cost and improve efficiency. On the other hand, if the task is semantic based, we need to rely on the LLM to extract the information.

PR #1339, opened 2023-11-03
Support semicolon and escaped strings in lark

pchunduri6:
Support semi-colons in string literals for queries of the form:
```
"""SELECT ChatGPT("Here's a; question", "This is the context") FROM TAIPAI;"""
```

Also support string escape to run ChatGPT queries more easily:
```
"""SELECT ChatGPT('Here\\'s a question', 'This is the context') FROM TAIPAI;"""
```


jarulraj:
The PR looks good to me @pchunduri6

pchunduri6:
> The PR looks good to me @pchunduri6. Anything pending?

It should be good to merge @jarulraj. Can you merge it?

PR #1340, opened 2023-11-04
added fitz and unstructuredio pdf loader py evadb script

Stru17:


xzdandy:
Hi, @Stru17 how should the PDFReader be used by the EvaDB?

jarulraj:
The idea is to replace add an additional PDF Reader backed by Unstructured IO -- not replace the default PDF reader.

jarulraj:
@Stru17 The output and structure of this `UnstructuredIOPDFReader` class should match that of the original `PDFReader` class. Currently, it is more of a Python script.

seansru:
I have provided an updated script, please take a look and provide some feedback on if I should further change/update it. I have also message the professor on Slack and would appreciate if I can get a response as soon as possible!

Issue #1067, opened 2023-09-07
Load PDFs give error "cannot open broken document"


aryan-rajoria:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar bug report.

### Bug

```
09-07-2023 13:15:53 ERROR [plan_executor:plan_executor.py:execute_plan:0182] cannot open broken document
Traceback (most recent call last):
  File "/home/great/projects/pyenvs/slacks/lib/python3.11/site-packages/evadb/executor/plan_executor.py", line 178, in execute_plan
    yield from output
  File "/home/great/projects/pyenvs/slacks/lib/python3.11/site-packages/evadb/executor/project_executor.py", line 33, in exec
    for batch in child_executor.exec(**kwargs):
  File "/home/great/projects/pyenvs/slacks/lib/python3.11/site-packages/evadb/executor/seq_scan_executor.py", line 40, in exec
    for batch in child_executor.exec(**kwargs):
  File "/home/great/projects/pyenvs/slacks/lib/python3.11/site-packages/evadb/storage/pdf_storage_engine.py", line 36, in read
    for batch in reader.read():
  File "/home/great/projects/pyenvs/slacks/lib/python3.11/site-packages/evadb/readers/abstract_reader.py", line 54, in read
    for data in self._read():
  File "/home/great/projects/pyenvs/slacks/lib/python3.11/site-packages/evadb/readers/pdf_reader.py", line 35, in _read
    doc = fitz.open(self.file_url)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/great/projects/pyenvs/slacks/lib/python3.11/site-packages/fitz/fitz.py", line 4041, in __init__
    _fitz.Document_swiginit(self, _fitz.new_Document(filename, stream, filetype, rect, width, height, fontsize))
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
fitz.fitz.FileDataError: cannot open broken document
```
```
FileDataError                             Traceback (most recent call last)
[...]

PR #1343, opened 2023-11-07
Verified that issue #1067 is resolved and added documentation for load pdf functionality.

kslohith:
Issue #1067 about not being able to load pdf files, was verified to be working with evadb documentation pdf and a new page for loading pdf is added to the documentation.
Screenshot 2023-11-07 at 1 33 01 AM

### DIFF
--- a/docs/_toc.yml
+++ b/docs/_toc.yml
@@ -45,6 +45,7 @@ parts:
           - file: source/reference/evaql/load_csv
           - file: source/reference/evaql/load_image
           - file: source/reference/evaql/load_video
+          - file: source/reference/evaql/load_pdf
           - file: source/reference/evaql/select
           - file: source/reference/evaql/explain
           - file: source/reference/evaql/show_functions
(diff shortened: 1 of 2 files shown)


jarulraj:
@kslohith Can you also please fix the broken link -- https://app.circleci.com/pipelines/github/georgia-tech-db/evadb/6047/workflows/bf1f29a3-36c9-40a0-97bb-fc084621b354/jobs/43847?invite=true#step-104-375531_124

Right now, if we load a corrupted pdf, what is the behavior?

jarulraj:
I have added a partial unit test here (not fully written) -- https://github.com/georgia-tech-db/evadb/tree/kslohith-staging

kslohith:
@jarulraj I tried to load a corrupted pdf file, it was successfully loaded into evadb, giving this response.
`                         0
0  Number of loaded PDF: 1`
But when i try to query the table, the response is empty, i think nothing was inserted.
We should modify this behaviour and give a error message when the upload itself was a corrupted pdf instead of giving a success message. 
Let me know if i should take this up and fix it.

Issue #1286, opened 2023-10-13
`SHOW ALL_CONFIGS` Feature Addition


hershd23:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar feature requests.

### Description

Once #1272 is merged we would like a way, using the `SHOW` command itself to display all present configs in the config table.

Something like `SHOW ALL_CONFIGS` or something similar would be required since the current parser implementation is like

`show_statement: SHOW (FUNCTIONS | TABLES | uid)`

### Use case

Displaying all configurations is an important for EvaDB users

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

hershd23:
Picking this issue myself

PR #1345, opened 2023-11-10
`SHOW CONFIGS` now gets all configs

hershd23:
Added method to get all configs
	modified:   evadb/catalog/catalog_manager.py
Added case to get all configs
	modified:   evadb/executor/show_info_executor.py
Test for get all configs
	modified:   test/integration_tests/short/test_show_info_executor.py


hershd23:
@gaurav274 and @xzdandy what should I compare it against in the all configs test? A little confused about that

hershd23:
Solved #1286

evadb/catalog/catalog_manager.py
```diff
@@ -646,3 +646,6 @@ def get_configuration_catalog_value(self, key: str, default: Any = None) -> Any:
         if table_entry:
             return table_entry.value
         return default
+
+    def get_all_configuration_catalog_entries(self):
```
xzdandy, line 650: Can we have a type hint for the return value?

evadb/executor/show_info_executor.py
```diff
@@ -51,15 +51,22 @@ def exec(self, *args, **kwargs):
...
                 show_entries.append(db.display_format())
         elif self.node.show_type is ShowType.CONFIG:
-            value = self.catalog().get_configuration_catalog_value(
-                key=self.node.show_val.upper(),
-            )
             show_entries = {}
-            if value is not None:
-                show_entries = {self.node.show_val: [value]}
+            # CONFIG is a special work, which is used to display all the configurations
+            if self.node.show_val.upper() == ShowType.CONFIG.name:
```
xzdandy, line 56: Does this mean that `CONFIG` is a keyword we can not set?

test/integration_tests/short/test_show_info_executor.py
```diff
@@ -128,6 +127,12 @@ def test_show_config_execution(self):
         with self.assertRaises(Exception):
             execute_query_fetch_all(self.evadb, "SHOW BADCONFIG")
 
+    def test_show_all_configs(self):
+        show_all_config_value = execute_query_fetch_all(self.evadb, "SHOW CONFIG")
+
+        # QUES : What do I test this against?
+        print(show_all_config_value)
```
xzdandy, line 134: Is it possible to get all configs from a clean evadb instance and compare them? So this testcases also serve for default available config options.

evadb/executor/show_info_executor.py
hershd23: Yes, still confused between using `CONFIG` vs `CONFIGS` since we use show with plurals like `FUNCTIONS` or `DATABASES`

test/integration_tests/short/test_show_info_executor.py
hershd23: I can use the python dict that we created to check for those configs, but the values can be changed and we do not want it to fail if any change has been made. Maybe we only check for keys

evadb/catalog/catalog_manager.py
hershd23: Sure, will get on this

evadb/executor/show_info_executor.py
xzdandy: We can have some list similar to `from evadb.catalog.sql_config import RESTRICTED_COL_NAMES` to prevent the access to the system key words.

test/integration_tests/short/test_show_info_executor.py
xzdandy: My opinion is that the test should fail when we add a new default configurations.  Or do you mean the other cases that will lead to the values to be changed?

evadb/executor/show_info_executor.py
hershd23: Should I add it in this list or create a new one?

xzdandy: Let's have a new one.

test/integration_tests/short/test_show_info_executor.py
hershd23: The problem is that values like the path are specific to each user and system like `EVADB_INSTALLATION_DIR`

xzdandy (approved):

PR #1359, opened 2023-11-15
Pytesseract Function

IshSiva:


test/integration_tests/long/functions/test_pytesseract.py
```diff
@@ -0,0 +1,45 @@
...
+        execute_query_fetch_all(self.evadb, f"DROP FUNCTION IF EXISTS {function_name};")
+
+        create_function_query = f"""CREATE FUNCTION IF NOT EXISTS{function_name}
+            IMPL 'evadb/functions/pytesseract_function.py';
+        """
+        execute_query_fetch_all(self.evadb, create_function_query)
+
+        ocr_query = f"SELECT {function_name}(data) FROM MyImage;"
+        output_batch = execute_query_fetch_all(self.evadb, ocr_query)
+        self.assertEqual(1, len(output_batch))
```
xzdandy, line 45: Is the result deterministic? If so, can we check the result of pytesseract?

test/integration_tests/long/functions/test_pytesseract.py
```diff
@@ -0,0 +1,45 @@
...
+        self.evadb.catalog().reset()
+
+        load_image_query = """LOAD IMAGE 'data/ocr/Example.jpg' INTO MyImage;"""
+
+        execute_query_fetch_all(self.evadb, load_image_query)
+
+    def tearDown(self) -> None:
+        execute_query_fetch_all(self.evadb, "DROP TABLE IF EXISTS MyImage;")
+
+    @unittest.skip("Needs Pytesseract")
```
xzdandy, line 33: - Please use the proper skipper maker. You can find examples in https://github.com/georgia-tech-db/evadb/blob/staging/test/markers.py
- And add pytesseract into the dev [setup.py](https://github.com/georgia-tech-db/evadb/blob/staging/setup.py), so the testcases will not be skipped on the circleci.

test/integration_tests/long/functions/test_pytesseract.py
IshSiva: we can check the results but pytesseract is not too good. So we may not get the required text in the output.

test/integration_tests/long/functions/test_pytesseract.py
IshSiva: updated this.

PR #1368, opened 2023-11-17
Migrate ChatGPT function to openai v1.0

pchunduri6:
Migrate ChatGPT function to openai v1.0.

The test is skipped in circleCI because we must supply the `OPENAI_API_KEY`. The test passes on local machine.

- [x] Upgrade ChatGPT function.
- [x] Upgrade Dall-e function.
- [x] Update unit test cases.
- [x] Verify that notebooks work correctly.


pchunduri6:
@Chitti-Ankith can you check if these changes fix the OpenAI version issues you were facing?

xzdandy (approved):
Thanks! Do we need to update the notebook?

pchunduri6:
> Thanks! Do we need to update the notebook?

I manually checked the notebooks. The notebooks only use the function definitions. Since we only changed the function implementation, I did not see the need for any updates.

PR #1376, opened 2023-11-20
Fix COUNT(*) ERROR for PDF SELECT Queries

### LINKED ISSUE #1170: `COUNT(*)` does not work as expected
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar bug report.

### Bug

COUNT(*) throws an `invalid column id` error.
image

### Environment

_No response_

- [x] Yes I'd like to help by submitting a PR!

jkim3663:
https://github.com/georgia-tech-db/evadb/issues/1170

COUNT(*) did not work with GROUP BY, while COUNT(column_name), or COUNT(SEGMENT(sth)) worked with group by. 
The issue was due to wrong column name assingment while creating a tuple value expression. 

I added a test case and a sample pdf that is same as the one from the issue.


xzdandy (approved):
Great job. Thanks for fixing the `COUNT(*)`!

PR #1380, opened 2023-11-21
Fix string helper functions not initialized

pchunduri6:
String helper functions CONCAT, UPPER, LOWER are not initialized due to a minor bug in the code.


Chitti-Ankith (approved):
Thanks for fixing the issue @pchunduri6 !

Issue #1217, opened 2023-09-26
Make it clear that `forward` in user defined functions only take one pandas dataframe as input. 


xzdandy:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar feature requests.

### Description

- [ ] Throw an error `forward() only takes one pandas dataframe as input.`, when a user-defined function call takes more than one pandas dataframe.
- [ ] Update the documentation to highlight that `forward() only takes one pandas dataframe as input`

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

PR #1382, opened 2023-11-22
Fix: Error handling for forward() to only take one pandas dataframe

duanduanzhou:
This pull request addresses Issue #1217 , focusing on enhancing error handling and input constraints within the `forward() `method. The key change involves enforcing a length constraint on `io_signature` to ensure `forward()` only accepts a single pandas DataFrame as input.

### DIFF
--- a/evadb/functions/test_bad_fuzzyjoin_udf.py
+++ b/evadb/functions/test_bad_fuzzyjoin_udf.py
@@ -1,0 +1,81 @@
+# coding=utf-8
+# Copyright 2018-2023 EvaDB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


test/integration_tests/long/test_function_executor.py
```diff
@@ -322,6 +322,22 @@ def test_should_raise_if_function_file_is_modified(self):
             # with self.assertRaises(AssertionError):
             execute_query_fetch_all(self.evadb, select_query)
 
+    def test_should_raise_error_for_multiple_dataframes(self):
```
xzdandy, line 325: Let's move this to the short integration tests.

evadb/functions/test_bad_fuzzyjoin_udf.py
```diff
@@ -0,0 +1,81 @@
...
+import pandas as pd
+from thefuzz import fuzz
+
+from evadb.catalog.catalog_type import NdArrayType
+from evadb.functions.abstract.abstract_function import AbstractFunction
+from evadb.functions.decorators.decorators import forward, setup
+from evadb.functions.decorators.io_descriptors.data_types import PandasDataframe
+
+
+class FuzzDistance(AbstractFunction):
```
xzdandy, line 41: We do not need to create a separate file for this function under `evadb/functions`, since this is not a valid function. 

We can directly define this function in the integration test file.

xzdandy:
This PR fixes #1217

PR #1387, opened 2023-11-23
Query Cost Estimation

kslohith:
This is the initial draft for cost estimation for queries. Ideally this should be in the optimizer module but for the time being i have added the changes to the executor module, For example when you use the explain command now the command along with returning the query plan will also return the estimated number of rows which are accessed. 

HLL Diagram
Screenshot 2023-11-22 at 7 21 08 PM


kslohith:
Screenshot 2023-11-22 at 7 52 19 PM

PR #1388, opened 2023-11-24
feat: supporting spaces in column names for csv files

fikremen:
takes `reverse quote id`, removes back ticks, and converts it to `simple id`.


americast (approved):
Thanks @fikremen!

xzdandy (approved):
Thanks. It looks good. Could we add a note in the documentation on using reverse quote for space in column names?

PR #1392, opened 2023-11-25
Notifying users about incorrect forward headers

### LINKED ISSUE #1329: Notify Users about Invalid Forward Headers
### Search before asking

### Description

As many issues have pointed out earlier (#1315, #1217, #1206, #1150), the forward header causes a lot of issues for users. 
In this particular request, I am asking for a mechanism to detect faulty forward headers so that users can fix the problems quickly and not have to dig into all the wrong areas of code. 

I think some ideas for requirements should be:
1. Notify users if the signature is missing. Right now, the error says "Cannot infer io signature from the decorator..."
2. Notify users if the signature is malformed (e.g. missing column names, types, or shapes)

### Use case

_No response_

hershd23:
To the existing error message adding
1. What signature failed (input or output)
2. Providing a signature template to the user
3. Added link to the forward decorator in the EvaDB documentation

Pointing users to the following documentation
https://evadb.readthedocs.io/en/stable/source/reference/ai/custom-ai-function.html#part-1-writing-a-custom-function

### DIFF
--- a/evadb/functions/decorators/utils.py
+++ b/evadb/functions/decorators/utils.py
@@ -50,7 +50,7 @@ def missing_io_signature_helper() -> str:
     More information on the how to create the forward decorator can be found here:
     https://evadb.readthedocs.io/en/stable/source/reference/ai/custom-ai-function.html#part-1-writing-a-custom-function
     """
-    print(signature_template)
+
     return signature_template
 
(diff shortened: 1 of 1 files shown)


xzdandy (approved):
Thanks for improving the error message!

Issue #1331, opened 2023-11-01
Accessing EvaDB tables using Postgres query engine


pchunduri6:
### Search before asking

- [X] I have searched the EvaDB [issues](https://github.com/georgia-tech-db/eva/issues) and found no similar feature requests.

### Description

Currently, it is not clear whether we can access EvaDB tables using a different query engine like Postgres. This will be useful when we want to leverage Postgres features like window functions and string_agg on the output of EvaDB tables. One concrete workflow that benefits from this is RAG-based question-answering:
1. Create vector index of a document in EvaDB.
2. Use similarity to retrieve top-5 similar chunks to the user question.
3. Concatenate all the retrieved chunks in `Postgres` using group by and string_agg.
4. Run an LLM function on the concatenated output.

If this is not possible, we would need to add support for `STRING_AGG` and `GROUP BY 'chunks'` natively in EvaDB.

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

ryanmle2001:
Claimed for EvaDB Project 2 @jarulraj

PR #1393, opened 2023-11-25
Native support for STRING_AGG

ryanmle2001:
Related to issue #1331 

Process:
- Added STRING_AGG to evadb.lark using LARK grammar per [EvaDB new command](https://tinyurl.com/evadb-new-command) documentation 
- Define new ExpressionType. AGGREGATION_STRING_AGG enum 
- Add AGGREGATION_STRING_AGG to aggregation_expression evaluate() 
- Define STRING_AGG aggregation in Batch class with error handlers for 1) invalid column input and 2) invalid delimiter 
- Unit tests


mathanmahe:
https://github.com/georgia-tech-db/evadb/pull/1427

script/test/test.sh
xzdandy, line 1: Could you revert the test.sh change?

test/unit_tests/expression/test_aggregation.py
```diff
@@ -116,6 +116,58 @@ def test_aggregation_max(self):
...
+                    1: ["Goodbye", "Everyone", "EvaDB", "Out"],
+                }
+            )
+        )
+        batch = aggr_expr.evaluate(tuples, delimiter=" ")
+        self.assertEqual("Hello World EvaDB Here", batch.frames.iloc[0][0])
+        self.assertNotEqual(str(aggr_expr), None)
+
+    def test_aggregation_string_agg_incorrect_column(self):
+        columnName = TupleValueExpression(name=0)
```
xzdandy, line 138: Could you elaborate on the situation which this test is for? Thanks.

test/unit_tests/expression/test_aggregation.py
```diff
@@ -116,6 +116,58 @@ def test_aggregation_max(self):
...
+                    0: ["Hello", "World", "EvaDB" "Here"],
+                    1: ["Goodbye", "Everyone", "EvaDB" "Out"],
+                }
+            )
+        )
+        with pytest.raises(AssertionError):
+            aggr_expr.evaluate(tuples, delimiter=" ")
+
+    def test_aggregation_string_agg_incorrect_delimiter(self):
+        columnName = TupleValueExpression(name=0)
```
xzdandy, line 155: Could you elaborate on the situation which this test is for? Thanks.

xzdandy (changes requested):
Thanks for adding the unit tests they look good. I think we also need an integration test where we have the `STRING_AGG` in a SQL query to verify that it works from end to end.

test/unit_tests/expression/test_aggregation.py
ryanmle2001: This unit test checks if the column is valid - that is, if a column being used in STRING_AGG is inside the table or not

test/unit_tests/expression/test_aggregation.py
ryanmle2001: Test if the delimiter argument is a string. Strings can only be concatenated with string delimiters

ryanmle2001:
> Thanks for adding the unit tests they look good. I think we also need an integration test where we have the `STRING_AGG` in a SQL query to verify that it works from end to end.

Can you send some code pointers on writing an integration test here?

xzdandy:
> > Thanks for adding the unit tests they look good. I think we also need an integration test where we have the `STRING_AGG` in a SQL query to verify that it works from end to end.
> 
> Can you send some code pointers on writing an integration test here?

@ryanmle2001 Here https://github.com/georgia-tech-db/evadb/blob/bbfa483bdf3797c02c1ec4efadfd0f3162c1a68c/test/integration_tests/long/test_huggingface_functions.py#L412 is an example, replace the SEGMENT with the STRING_AGG

script/test/test.sh
xzdandy: I think the access the permission is still changed. 100644 → 100755

PR #1397, opened 2023-11-25
Add E2B Sandboxed Code Execution Integration

YukiHinana:
This change adds a new built-in function for executing python code generated by OpenAI Chat Completion model and the supporting documentation for said function.


evadb/functions/e2b_exec.py
```diff
@@ -0,0 +1,61 @@
...
+        ],
+    )
+    def forward(self, frames: InputType) -> InputType:
+        self.sandbox = Sandbox()
+        output = []
+        stderrs = []
+        for code in frames['codejson']:
+            code_obj = json.loads(code)
+            code_text = json.loads(code_obj['function_call']['arguments'])['code']
+            self.sandbox.filesystem.write('/home/user/test.py', code_text)
```
YukiHinana, line 64: This is hard coded, but since it is what is executed in the VM/sandbox. this is probably fine for now.

PR #1417, opened 2023-11-25
Add column matching for function pipelining

Y1feiwang:
matches custom function's expected name with the actual input name through first matching name, then shape of the data in the column, and at last type. All change are made in evadb/models/storage/batch.py.


xzdandy:
Hi, I think this PR does not merge correctly with staging. It reverts mange changes on the staging branch. We need to fix the merge errors. Thanks!

PR #1418, opened 2023-11-26
Added SHOW JOBS statement to list all jobs

kompy99:
## SHOW JOBS

Added support to list all jobs using the `SHOW` statement.

### Usage

`SHOW jobs`

## Minor Bug fix
Fixed minor bug in CREATE JOB

**Bug description**: Error thrown when `START` parameter is empty in `CREATE JOB` statement due to minor mistake in use of datetime.now()


americast (changes requested):
Thanks! It would be great if you could please fix formatting issues using the inbuilt linter by issuing the following command: `$ bash script/test/test.sh -m LINTER`.

PR #1420, opened 2023-11-26
Bug Fix for SELECT with uppercase columns (Approach 1)

anmolagarwalcp810:
Problem Statement
Select query for tables with tables with upper case columns wasn't working: [Issue 1079](https://github.com/georgia-tech-db/evadb/issues/1079)

image

# Solution
Since the design principle of EvaDB is that the columns are to be treated as lower case, the solution consisted of the following steps:
1. Modify CREATE TABLE command to convert columns passed as input to lower case. This means that table itself has lower cased columns from the start.
2. With above change, SELECT works successfully.
3. However, INSERT command still didn't convert columns to lower case, and was giving same error as column not found. Hence, also modified backend of INSERT command to convert columns to lower case.

# Output after Bug Fix
image


anmolagarwalcp810:
Just a doubt: currently even though I have converted columns to always remain lower case, TupleValueExpression objects still still store column name in original case, would that be an issue?
For example in the value of name member inside one TupleValueExpression object remains "featCol" instead of "featcol" in test case `tests/unit_tests/parser/test_parser.py::ParserTests.test_create_index_statement()`.

anmolagarwalcp810:
Also, all test cases passed locally, except these three:
`
FAILED test/integration_tests/long/test_create_table_executor.py::CreateTableTest::test_should_create_table_from_select_lateral_join - AssertionError: False is not true
`
`
FAILED test/integration_tests/long/test_single_document_similarity.py::SingleDocumentSimilarityTests::test_single_pdf_should_work - evadb.executor.executor_utils.ExecutorError: 'mypdf._row_number'
`
`
FAILED test/unit_tests/test_eva_cmd_client.py::CMDClientTest::test_evadb_client - RuntimeError: There is no current event loop in thread 'MainThread'.
`
`
======================================================================================== 3 failed, 538 passed, 30 skipped, 13 deselected, 2 xfailed, 1 xpassed, 24 warnings in 5738.51s (1:35:38) ========================================================================================
FULL TEST CODE: --|1|-- FAILURE
`
However, when I ran each of them individually in visual code pytests, each of them passed without any issues. Not sure, why they weren't working when running them in bundle.

PR #1421, opened 2023-11-26
Enhancing INSERT Command to Support Inserting Multiple Values

anmolagarwalcp810:
# Problem Statement
INSERT command wasn't inserting multiple values into table. It was only inserting the first tuple and ignoring everything else.
image

# Solution
Modified backend of INSERT command by finding out two things:
1. Place where other tuples are dropped except the first one. That happened inside `parser/lark_visitor/_insert_statements.py`.
2. Place where INSERT command is actually executed, and does it support inserting multiple tuples and it does, `executor/insert_executor.py`.

After find those two things, I simply captured all the tuples coming from the tree created by Lark, and passed them to the executor through the planner. Tried to make sure that there are no issues inside planner because of that.

I also modified hash functions of InsertTableStatement, InsertPlan and LogicalInsert classes because value_list member became 2-dimensional after my change, and couldn't be hashed. So I converted each element of value_list to tuple, and then hash functions were working.

# Output after enhancing INSERT command
image


anmolagarwalcp810:
Also, all test cases passed locally, except these two:
`FAILED test/integration_tests/long/test_create_table_executor.py::CreateTableTest::test_should_create_table_from_select_lateral_join - AssertionError: False is not true
`
`FAILED test/unit_tests/test_eva_cmd_client.py::CMDClientTest::test_evadb_client - RuntimeError: There is no current event loop in thread 'MainThread'.
`
However, when I ran each of them individually in visual code pytests, each of them passed without any issues. Not sure, why they weren't working when running them in bundle.

xzdandy (approved):
Thanks for adding inserting multiple value support.

PR #1423, opened 2023-11-26
Adding cockroach-db intergration support for EvaDB

VrtanoskiAndrej:
Implementing the CockroachDB Handler, adding unit tests and adding documentation support for CockroachDB.

### DIFF
--- a/docs/source/reference/databases/cockroachdb.rst
+++ b/docs/source/reference/databases/cockroachdb.rst
@@ -1,0 +1,36 @@
+CockroachDB
+==========
+
+The connection to CockroackDB is based on the `psycopg2 <https://pypi.org/project/psycopg2/>`_ library.
+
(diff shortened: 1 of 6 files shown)


docs/source/reference/databases/cockroachdb.rst
```diff
@@ -0,0 +1,36 @@
+CockroachDB
+==========
```
americast, line 2: The docs build is failing in this line. Maybe we need a few more `=`?

americast (changes requested):
You may use the in-built linter to format the code with `bash script/test/test.sh -m LINTER`.

americast (changes requested):
Perhaps you would have to add cockroachdb installation procedure in `setup.py` for the unit test to pass.

PR #1427, opened 2023-11-26
Postgres integration 

mathanmahe:
Related to issue #1331 

Postgres implementation is complete, string_agg is incomplete


americast (changes requested):
Thanks for your PR @mathanmahe! It would be great if you could enlist `psycopg2` as a dependency in `setup.py`.

PR #1454, opened 2024-05-17
Fixes broken unit test - test_evadb_client

Tanish0019:
## Problem
Unit test suite is broken. ([Last run](https://app.circleci.com/pipelines/github/georgia-tech-db/evadb/6360/workflows/4b306256-29cf-421c-b173-02877f63b431/jobs/45957))
Broken test - `CMDClientTest::test_evadb_client`
Error: `RuntimeError: There is no current event loop in thread 'MainThread'`
The test runs fine when run individually but fails when whole suite is run together because other asyncio tests update the asyncio loop settings and policies.

## Solution
Update test class to extend from `IsolatedAsyncioTestCase` which provides better isolation when running multiple asyncio test units.

### DIFF
--- a/test/unit_tests/test_eva_cmd_client.py
+++ b/test/unit_tests/test_eva_cmd_client.py
@@ -23,7 +23,7 @@ from evadb.evadb_config import BASE_EVADB_CONFIG
 
 
 # @pytest.mark.skip
-class CMDClientTest(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


Tanish0019:
Hi @xzdandy, I was just playing around with the project and caught this quick fix so thought about contributing back. Currently the build is failing in the Install Evadb package step unrelated to my change. Do you know if this is a known issue or we can trigger a retry maybe? Thanks!

